PR #1, opened 2015-07-01
WIP: Run buildkite agent on instances rather than docker

lox:
This simplifies the whole setup and eliminates a layer of docker-in-docker.


toolmantim:
is this still the direction you're shooting for? Agents directly on the machines?

lox:
I'm kind of torn on this. The downside is that it's very difficult to support multiple parallel docker versions. I think probably no.

toolmantim:
Yeah, perhaps you'd create a new cluster to test a new Docker version? And I guess a project could depend on a new Docker feature… but I guess the downside of running on bare metal is they're less disposable / more cruft-accreting?

PR #3, opened 2015-07-07
Mappings.AWSArch2AMI.us-east-1.HVM64EBS updated.

pda:
Previous AMI resulted in CREATE_FAILED: AMI cannot be described

New AMI is:
us-east-1 vivid 15.04 DEVEL amd64 hvm:ebs-io1 20150616.1  ami-53986a38  hvm


pda:
Turns out this is not meant to be a public AMI.

PR #4, opened 2015-07-07
EC2 LaunchConfiguration AssociatePublicIpAddress: true

pda:
Ensures each EC2 instance is allocated a public IP address.  VPC may default to
not allocating public IP address.


lox:
I thought this was an attribute of the subnet?

PR #9, opened 2015-10-11
Remove "LTS" from the AMI release tag

dgoodlad:
According to [Ubuntu](https://wiki.ubuntu.com/LTS), 15.04 isn't an LTS release; 15.10 will be.

### DIFF
--- a/packer/buildkite-ubuntu-15.04.json
+++ b/packer/buildkite-ubuntu-15.04.json
@@ -10,7 +10,7 @@
       "ami_description": "A docker host image based on Ubuntu 15.04",
       "ami_groups": ["all"],
       "tags": {
-        "Release": "15.04 LTS",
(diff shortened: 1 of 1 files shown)


pda:
LTS is scheduled every two years (every fourth release), the most recent being 14.04 and the next being 16.04… unless there's some big changes I'm not seeing.

So yeah… :+1: for this PR removing it, but :-1: for 15.10 being the next :wink:

daveslutzkin:
:+1: for @pda's reasoning

dgoodlad:
>  :-1: for 15.10 being the next

:joy: Fight the good fight, @pda!

lox:
Ugh, my bad. For some reason I thought that all .04's were LTS. From that graph, 16.04 is the next LTS, right?

PR #11, opened 2015-11-10
Update s3 links due to permanent aws redirect

krutisfood:
Original link results in 301 permanent redirect to location of new link.


pda:
:+1:

PR #21, opened 2016-02-18
Single buildkite-agent on host for easier autoscaling

lox:
This fundamentally changes how buildkite-aws-stack works by retreating from the idea of running many bk agents on a single docker host to a single buildkite agent running directly on the host. A customized AMI is built to speed up autoscaling operations.


lox:
Ok, this is happening.

toolmantim:
I endorse this lack of magic :tada:

lox:
Somehow in the process of simplifying I've added more code.

toolmantim:
Don't forget all the code inside the moving parts we're removing :)

PR #22, opened 2016-02-22
Add generic support for encrypted secrets stored on s3

lox:
This change will require that secrets like the github key be encrypted at rest and decrypted per job with a project specific secret key.

See specifically the changes to the README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ aws cloudformation create-stack \
 | BuildkiteOrgSlug             | Your Buildkite Organization slug (e.g 99designs)                     |                 |
 | BuildkiteAgentToken          | Your Buildkite Agent Token                                           |                 |
 | BuildkiteQueue               | The Buildkite queue to give the agents                               | elastic         |
-| ProvisionBucket              | An S3 bucket and prefix that contains your credentials/keys          |                 |
+| SecretsBucket                | An S3 bucket (and optional prefix) that contains secrets             |                 |
 | InstanceType                 | The EC2 instance size to launch                                      | t2.nano         |
 | MinSize                      | The minimum number of instances to launch                            | 1               |
 | MaxSize                      | The maximum number of instances to launch                            | 1               |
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -38,19 +38,28 @@ Check out [`buildkite-elastic.yml`](templates/buildkite-elastic.yml) for more de
...
+You provide a key via an environment var in your Buildkite config called `BUILDKITE_SECRETS_KEY` which will be used to decrypt all the files found in the secrets bucket.
 
-For Docker Hub credentials, you can use `DOCKER_HUB_USER`, `DOCKER_HUB_PASSWORD` and `DOCKER_HUB_EMAIL`.
+Two files are specifically looked for, `id_rsa_github`, for checking out your git code and optionally `env`, which contains environment variables to expose to the job command.
+
+### Uploading your Secrets
 
 ```bash
-aws s3 cp --acl private my_id_rsa_key "s3://my-provision-bucket/id_rsa_github"
+PASSPHRASE=$(date +%s%N | sha256sum | base64 | head -c 32 ; echo) # generates a 32 character randomish password, use gdate for MacOS
```
Vektah: This is pretty terrible advice for key generation. If someone knows roughly when you create it they have a very small space to search - one per second they are off by

What about:
```
< /dev/urandom tr -dc '[:alnum:]' | head -c32
```

lox: Fair point, needs more entropy. It does include nanoseconds, but agree it's bad advice. That example doesn't work under OSX.

Vektah: no `/dev/random`?

lox:
```
 ~  › < /dev/urandom tr -dc '[:alnum:]' | head -c32
tr: Illegal byte sequence
Zh%
```

Vektah: `head -c 32  /dev/random | base64 | head -c 32` is another way to go. uses very little apart from /dev/random and will block until the system has enough entropy.

templates/buildkite-elastic.yml
```diff
@@ -185,9 +185,9 @@ Resources:
         config:
           commands:
             01-download-github-key:
-              test: test -n "$(ProvisionBucket)"
+              test: test -n "$(SecretsBucket)"
               command: |
-                aws s3 cp s3://$(ProvisionBucket)/id_rsa_buildkite /root/.ssh/id_rsa
+                aws s3 cp s3://$(SecretsBucket)/id_rsa_buildkite /root/.ssh/id_rsa
```
pda: Is the plan to updated or removed this? The `aws s3 cp …` command without `--sse-c` will fail for a client-encrypted object;

```
$ aws s3 cp s3://pda-test/payload ./
A client error (400) occurred when calling the HeadObject operation: Bad Request

$ aws s3 cp --sse-c --sse-c-key "$key" s3://pda-test/payload ./
download: s3://pda-test/payload to ./payload
```

And if it did successfully download, the secret will be unencrypted at rest.

Is the plan to download the secrets from S3 for each job? That seems to be the only way to use S3 client-side encryption here. Or should the encryption/decryption be done completely external to S3? Or does “encrypted at rest” refer to “at rest on S3” and not on the agent's local disk?

lox:
See most recent change @pda.

packer/conf/hooks/environment
```diff
@@ -3,30 +3,25 @@ set -eu
...
-  echo "~~~ Authenticating with :docker: as ${DOCKER_HUB_USER}"
-  docker login --username="${DOCKER_HUB_USER}" --password="${DOCKER_HUB_PASSWORD}" --email="${DOCKER_HUB_EMAIL}"
-elif [[ -n "${DOCKER_HUB_USER:-}" && -n "${DOCKER_HUB_PASSWORD:-}" ]] ; then
-  echo "Missing DOCKER_HUB_EMAIL."
+if [[ -n "$BUILDKITE_SECRETS_BUCKET" ]] && [[ -z "$BUILDKITE_SECRETS_KEY" ]] ; then
+  echo "A SecretsBucket was provided, but no encryption key via BUILDKITE_SECRETS_KEY."
   exit 1
 fi
 
+SSH_KEY_URL="${SSH_KEY_URL:-s3://${BUILDKITE_SECRETS_BUCKET}/id_rsa_github}"
```
pda: Is it worth keeping this user-specifiable given it's coupled to `--sse-c-key $BUILDKITE_SECRETS_KEY` etc?
If anything perhaps just the `id_rsa_github` component could be `${SSH_KEY_NAME:-id_rsa_github}`.

lox: Fair point.

lox:
Merging unless further feels @pda

packer/conf/hooks/environment
```diff
@@ -3,30 +3,26 @@ set -eu
...
+eval "$(ssh-agent -s)"
 
 # startup an ssh-agent with the ssh-key
 if [[ -n "${SSH_KEY_URL:-}" ]] ; then
-  echo "~~~ Downloading :github: ssh key"
-  TMPKEY=$(mktemp /tmp/buildkite.XXXXXXXXXX)
-  trap "rm $TMPKEY" exit
-  env -i aws s3 cp ${SSH_KEY_URL:-} $TMPKEY
-  ssh-add $TMPKEY
+  echo "~~~ Downloading github ssh key from $SSH_KEY_URL"
```
pda: The download has already been attempted here, the message implies it's about to happen.
You could wrap the download in a function which outputs the key or exits non-zero on failure, rather than stashing it in a temporary variable up above.

PR #23, opened 2016-02-24
Support encrypted env file in secrets

lox:
Currently the README mentions that you can have an encrypted `env` file in your `SecretsBucket` which will be sourced and executed as part of the Buildkite environment hook.

### DIFF
--- a/packer/conf/hooks/environment
+++ b/packer/conf/hooks/environment
@@ -2,11 +2,11 @@
 set -eu
 
 s3_exists() {
-  env -i aws s3 ls "${BUILDKITE_SECRETS_BUCKET}/$1" "$2" &>/dev/null
+  env -i aws s3 ls "s3://${BUILDKITE_SECRETS_BUCKET}/$1" "$2" &>/dev/null
 }
 
 s3_download() {
-  env -i aws s3 cp --quiet --sse-c AES256 --sse-c-key "${BUILDKITE_SECRETS_KEY}" "${BUILDKITE_SECRETS_BUCKET}/$1" /dev/stdout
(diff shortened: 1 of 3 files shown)


toolmantim:
What's this env thing?

lox:
Updated the description @toolmantim. Because you might have different projects running on the one stack, you need a way to deliver ENV's.

packer/conf/hooks/environment
```diff
@@ -19,11 +14,21 @@ if [[ -n "${BUILDKITE_SECRETS_BUCKET:-}" ]] && [[ -z "${BUILDKITE_SECRETS_KEY:-}
...
-# startup an ssh-agent with the ssh-key
 eval "$(ssh-agent -s)"
-ssh-add <(download_github_key "$SSH_KEY_NAME")
+
+if [[ -n "${BUILDKITE_SECRETS_BUCKET:-}" ]] ; then
+  [[ -d ~/secrets ]] || mkdir -p ~/secrets
+
+  echo "Downloading ssh key from $BUILDKITE_SECRETS_BUCKET"
+  if s3_download "${SSH_KEY_NAME:-id_rsa_github}" ~/secrets/id_rsa_github ; then
+    ssh-add ~/secrets/id_rsa_github
```
Vektah: What was the reason to move away from <()?

It seems a little nicer then having to manually clean up these files.

toolmantim:
Oh, the `env` technique is a good one

packer/conf/hooks/environment
lox: Yeah, agree @Vektah

PR #24, opened 2016-02-24
Custom autoscaling based on Buildkite/ScheduledJobsCount

lox:
This is the beginning of autoscaling based on the metrics provided by https://github.com/buildkite/buildkite-cloudwatch-metrics-publisher. 

You can specify an `AutoscaleStrategy` of `scheduledjobs` to get this new behaviour.

A future PR will likely bring in the Lambda functions directly into this stack so we have a one-stop-shop.


toolmantim:
:clap:

PR #25, opened 2016-02-29
Make upstart workaround conditional

sj26:
This will be done for us in the new buildkite-agent 2.1.5 stable package.

/cc @lox


lox:
+1

PR #26, opened 2016-03-01
Allow empty secrets key

lox:
For easier backwards compatibility with stacks that didn't have encrypted keys in a provision bucket, this allows `BUILDKITE_SECRETS_KEY` to be empty.

It also provides some extra configuration points via ENV:

 - `BUILDKITE_SECRETS_PREFIX` - This is a key prefix for the secrets bucket, by default it's the name of the pipeline
 - `SSH_KEY_NAME` - The name of the key file (under the above prefix), defaults to `id_rsa_github`

This change also sets the default url for env and the ssh key file to be under the above `BUILDKITE_SECRETS_PREFIX`, which is a breaking change. You can set an empty secrets prefix to emulate previous behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,11 +45,13 @@ Set your [Agent Query Rules](https://buildkite.com/docs/agent/agent-meta-data) t
 
 ## Secrets
 
-Your stack has access to the `SecretsBucket` parameter you passed in. This should be used in combination with strong encryption to ensure that your CI secrets (such as Github credentials) are reasonably secure. See the [Security](#security) section for more details.
-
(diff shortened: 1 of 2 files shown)


toolmantim:
So does this mean you could set up a thing where all teams know that if they create a bucket "x" for a pipeline, then that stuff will magically appear in their build? Because that's an ace default. I wonder if we can do that for a filesystem based env default for the agent… like a standard `envs` directory and `ENV_SECRET` support, but it's just based on file paths. This is based on S3

lox:
It's the one bucket, but it looks in `mybucket/myproject`, where `myproject` is the BUILDKITE_SECRETS_PREFIX slug. We could do something like support specific bucket names too, for instance anything that ends with `-buildkite`, or perhaps anything with a custom bucket tag of some sort.

PR #27, opened 2016-03-02
Add back support for DOCKER_HUB_AUTH

lox:
This also unsets the docker auth vars so that they aren't exposed to the command


toolmantim:
:+1: I'm curious what's this used for?

lox:
@toolmantim basically older docker versions used a different format for configuration. This was supported in earlier versions of buildkite-aws-stack, so this just adds back that support to make it more backwards compatible.

PR #28, opened 2016-03-02
Use docker 1.10 user namespacing

Vektah:
Docker 1.10 introduces user namespacing. This allows us to lock down the docker daemon to a non root user account.

This means doing 
```
docker run -it -v /:/hostroot busybox touch /hostroot/things-owned-by-root
```
Will no longer be able to modify the host as root.

Apart from being much more secure, it means that files created inside a docker based build script will be owned by buildkite-agent, instead of root. This happens often when projects volume mount the project root into the container.

This is a brand new feature, and there are likely to be bugs. At the very least wait till 1.10.3 for https://github.com/docker/docker/pull/20446 to be fixed.

Things that do not work with userns:
 - **Using --privileged mode flag on docker run**
 - sharing PID or NET namespaces with the host (--pid=host or --net=host)
 - sharing a network namespace with an existing container (--net=container:*other*)
 - sharing an IPC namespace with an existing container (--ipc=container:*other*)
 - A --readonly container filesystem (this is a Linux kernel restriction against remounting with modified flags of a currently mounted filesystem when inside a user namespace)
 - external (volume or graph) drivers which are unaware/incapable of using daemon user mappings

### DIFF
--- a/packer/conf/docker.conf
+++ b/packer/conf/docker.conf
@@ -2,4 +2,4 @@
 # running containers.  The default value of 1048576 mirrors the value
 # used by the systemd service unit.
 DAEMON_MAXFILES=1048576
-OPTIONS="--default-ulimit nofile=1024:4096 -s overlay"
+OPTIONS="--default-ulimit nofile=1024:4096 -s overlay --userns-remap=buildkite-agent"
--- a/packer/conf/subgid
+++ b/packer/conf/subgid
@@ -1,0 +1,1 @@
+buildkite-agent:496:65536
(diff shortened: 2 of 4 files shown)


lox:
This also solves #30

jeromegn:
I also don't think docker 1.10 is available yet for the Amazon Linux AMI (which this AMI depends on) :(

lox:
Indeed, although the worse blocker is an open bug in 1.10. This PR will hang around until 1.10.1 is out.

Vektah:
I've updated this to use docker 1.10.3, should be good to go but CI dosent seem to trigger on 3rd PRs.

PR #32, opened 2016-03-17
Use docker 1.10 user namespacing

Vektah:
Docker 1.10 introduces user namespacing. This allows us to lock down the docker daemon to a non root user account.

This means doing
```
docker run -it -v /:/hostroot busybox touch /hostroot/things-owned-by-root
```
Will no longer be able to modify the host as root.

Apart from being much more secure, it means that files created inside a docker based build script will be owned by buildkite-agent, instead of root. This happens often when projects volume mount the project root into the container.

This is a brand new feature, and there are likely to be bugs. ~~At the very least wait till 1.10.3 for docker/docker#20446 to be fixed.~~

Things that do not work with userns:

Using --privileged mode flag on docker run
 - sharing PID or NET namespaces with the host (--pid=host or --net=host)
 - sharing a network namespace with an existing container (--net=container:other)
 - sharing an IPC namespace with an existing container (--ipc=container:other)
 - A --readonly container filesystem (this is a Linux kernel restriction against remounting with modified flags of a currently mounted filesystem when inside a user namespace)
 - external (volume or graph) drivers which are unaware/incapable of using daemon user mappings

### DIFF
--- a/packer/buildkite-ami.json
+++ b/packer/buildkite-ami.json
@@ -20,11 +20,11 @@
     },
     {
       "type": "shell",
-      "script": "scripts/install-docker.sh"
+      "script": "scripts/install-buildkite.sh"
     },
     {
       "type": "shell",
-      "script": "scripts/install-buildkite.sh"
+      "script": "scripts/install-docker.sh"
     }
   ]
 }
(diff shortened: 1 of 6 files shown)


Vektah:
This is https://github.com/buildkite/buildkite-aws-stack/pull/28 pushed into this repo so that CI will run.

lox:
@Vektah mind rebasing and updating for docker 1.11.1?

toolmantim:
I think we should just close this for now, until we're ready to give it a stab again

Issue #30, opened 2016-03-02
Docker daemon runs as root, writing to volume mounts leaves files unreadable for buildkite-agent


lox:
If builds run docker commands and mount in volumes, files are written as root, due to Docker's root-only-ness. 

This means that subsequent builds fail with permission errors because they can't `git clean` the files.

lox:
More details here https://github.com/docker/docker/issues/3124

jeromegn:
I just stumbled on this very issue. My workaround is to build an image ADDing the files instead of mounting a volume on a generic image.

lox:
A terrible hack that we have as a workaround is to use the docker socket to chown the files in a `post-command` hook:

```
#!/bin/bash
set -eu

docker run -it -v $(pwd):/project busybox chown -R 498.496 /project
```

PR #33, opened 2016-03-18
Chown the builds before checkout

Vektah:
This PR adds a 
```
chown -R  buildkite-agent.buildkite-agent /var/lib/buildkite-agent/builds
```

before doing a checkout. 

This works around permission issues caused volume mounts leaving files owned by root in the checkout. 

I think the long term aim should be https://github.com/buildkite/buildkite-aws-stack/pull/32, as it fixes the underlying security issue and not just the symptom, but it needs more thorough testing.


lox:
This fixes #30

PR #36, opened 2016-03-22
bump docker-compose version

whatupdave:
Let's you use version 2 of the docker-compose.yml file https://docs.docker.com/compose/compose-file/#version-2

### DIFF
--- a/packer/scripts/install-docker.sh
+++ b/packer/scripts/install-docker.sh
@@ -9,7 +9,7 @@ sudo service docker start
 sudo docker info
 
 # installs docker-compose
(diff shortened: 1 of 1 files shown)


lox:
That would need Docker 1.10, I believe we are still on 1.9 aren't we?

whatupdave:
Oh good point, I was using this locally with https://github.com/buildkite/buildkite-aws-stack/pull/32

Vektah:
https://github.com/buildkite/buildkite-aws-stack/pull/37 updates to 1.10.3, so this should be good to go.

PR #37, opened 2016-03-31
Bump to docker 1.10

Vektah:
This splits the docker 1.10.3 version bump out of https://github.com/buildkite/buildkite-aws-stack/pull/32

### DIFF
--- a/packer/scripts/install-docker.sh
+++ b/packer/scripts/install-docker.sh
@@ -6,7 +6,10 @@ sudo usermod -a -G docker ec2-user
 sudo cp /tmp/conf/docker.conf /etc/sysconfig/docker
 
 # Overwrite the yum packaged docker with the latest
+# Releases can be found at https://github.com/docker/docker/releases
+# shasums can be found at $URL.sha256
 sudo wget https://get.docker.com/builds/Linux/x86_64/docker-1.10.3 -O /usr/bin/docker
+echo 'a8315f0ff661e6a24a8b83743b6a4be87765bd000fab628de1a8c398b84966cc  /usr/bin/docker' | sha256sum -c
 sudo chmod +x /usr/bin/docker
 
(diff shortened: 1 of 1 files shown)


mbj:
Please validate the downloaded binary against a checksum such as sha256.

Vektah:
@mbj Hows that look?

mbj:
@Vektah you verify the binary AFTER you put it into PATH, so even when verifaction fails the potential dangerous binary can be executed by others quite easily. (Okay execute permissions are missing, still).

Also I suggest to add `--strict`.

mbj:
@Vektah Also does a failing `sha256sum --check` exit the script?

Vektah:
```
roci :: ~ » echo 'a8315f0ff661e6a24a8b83743b6a4be87765bd000fab628de1a8c398b84966cc  /usr/bin/docker' | sha256sum -c
/usr/bin/docker: FAILED
sha256sum: WARNING: 1 computed checksum did NOT match
roci :: ~ » echo $?
1
```

The packer image isnt going to be built on saved if it failed so the binary should never be used but I guess if someone copy pastes it somewhere else its better to have the extra step...

pda:
However `echo foo | shasum` isn't a "simple command" so the non-zero exit may not terminate the script despite `bash -e`.

Vektah:
@pda from http://www.gnu.org/software/bash/manual/html_node/Pipelines.html
> Each command in a pipeline is executed in its own subshell (see Command Execution Environment). The exit status of a pipeline is the exit status of the last command in the pipeline, unless the pipefail option is enabled ...

In this case I'm not expecting echo to fail so pipefail isnt needed. 

Pipefail is super important for things like
```
cur foo.com/installer | sh (which is super evil regardless)
```

mbj:
I think pipefail should be enabled by default, as its the better default.

PR #43, opened 2016-04-20
Pass subnets into metrics stack, fixes #42

### LINKED ISSUE #42: VpcId isn't propagated to the MetricsStack
When AWS Stack goes to create the MetricsStack instance, it uses the Default VPC, instead of the VpcId defined in the properties (or the VPC is creates for you). If the AWS Account doesn't have a Default VPC, the cloud formation script exits with a cryptic error.

pasted_image_at_2016_04_18_03_30_pm

Thanks @ryanbigg for helping us find this problem!

lox:
Previously the created VPC or specified existing subnets weren't being passed down to the metrics stack, leading to failures like #42.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ build-ami:
 upload: build/aws-stack.json
 	aws s3 sync --acl public-read build s3://buildkite-aws-stack/
 
-create-stack: build/aws-stack.json
+create-stack: templates/mappings.yml build/aws-stack.json
 	aws cloudformation create-stack \
 	--output text \
 	--stack-name buildkite-$(shell date +%Y-%m-%d-%H-%M) \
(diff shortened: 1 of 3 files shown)


keithpitt:
Ah nice! If you're looking for someone to test @ryanbigg may be able to for us!

radar:
I'll give this a go now :)

radar:
I uploaded the template to AWS using their upload functionality, entered all the details and now I get this error when I try to create the stack:

image

lox:
Apologies, this is still WIP, but I'm close.

lox:
@radar you'll need to use `make create-stack` and have a config.json if you are doing it from the console.

keithpitt:
Ah sorry, that was my fault @radar - I got too excited about the fix :)

PR #52, opened 2016-05-04
Added option for ArtifactsBucket to CFN

tobyjoe:
This PR adds an optional `ArtifactsBucket` parameter (and associated policies) to the AWS stack.

This is nice for uploading artifacts outside of buildkite agents...

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ aws cloudformation create-stack \
 | BuildkiteApiAccessToken      | A Buildkite API token for metrics                                    |                 |
 | BuildkiteQueue               | The Buildkite queue to give the agents                               | elastic         |
 | SecretsBucket                | An S3 bucket (and optional prefix) that contains secrets             |                 |
+| ArtifactsBucket              | An S3 bucket (and optional prefix) that contains build artifacts     |                 |
 | InstanceType                 | The EC2 instance size to launch                                      | t2.nano         |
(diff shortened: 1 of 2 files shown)


lox:
Many thanks! Nicely done.

toolmantim:
Nice!! I'd like to improve the label. So this sets up the correct policy, but doesn't create the bucket or set the environment variable to be used by the agent right? (we should be clear with the secrets bucket too)

lox:
Good point @toolmantim, is there an agent setting for setting up a custom bucket? This should absolutely do that. It doesn't create the bucket, but we should document that.

toolmantim:
Yep https://buildkite.com/docs/agent/build-artifacts#using-your-own-s3-bucket

```
export BUILDKITE_S3_DEFAULT_REGION=eu-central-1 # default is us-east-1
export BUILDKITE_S3_ACL=private # default is public-read
export BUILDKITE_ARTIFACT_UPLOAD_DESTINATION=s3://name-of-your-s3-bucket/$BUILDKITE_JOB_ID
```

And then if you wanted to expose the artifacts as web-clickable straight from Buildkite, and have all their linked/embedded artifacts working (like a coverage html report type thing), you could spin up https://hub.docker.com/r/daveoxley/s3-proxy/ via EB or something, and serve them up with basic auth. If you export `BUILDKITE_S3_ACCESS_URL` we'll use that host to link to instead of `s3.amazon.etc`

PR #53, opened 2016-05-04
Tweaks to makefile for easier setup

tobyjoe:
This PR adds a setup step to the default make target (in case bundle isn't installed, for example), and makes it possible to provide variables to switch AWS profiles and stack JSON publication bucket endpoints.

It also adds a target to update your stack.

I did remove the date stamping of stacks because it made updates difficult. We could cache the value locally or something... or make it a parameter?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all clean build build-ami upload create-stack
+.PHONY: all clean build build-ami upload create-stack update-stack config.json
 
 BUILDKITE_STACK_BUCKET ?= buildkite-aws-stack
 
@@ -27,9 +27,10 @@ build-ami:
 upload: build/aws-stack.json
 	aws s3 sync --acl public-read build s3://${BUILDKITE_STACK_BUCKET}/
 
-create-stack: templates/mappings.yml build/aws-stack.json
+config.json:
 	test -s config.json || { echo "Please create a config.json file"; exit 1; }
 
+create-stack: config.json templates/mappings.yml build/aws-stack.json
 	aws cloudformation create-stack \
 	--output text \
 	--stack-name buildkite \
@@ -43,9 +44,7 @@ validate: build/aws-stack.json
 	--output table \
 	--template-body "file://${PWD}/build/aws-stack.json"
 
-update-stack: templates/mappings.yml build/aws-stack.json
-	test -s config.json || { echo "Please create a config.json file"; exit 1; }
-
+update-stack: config.json templates/mappings.yml build/aws-stack.json
 	aws cloudformation update-stack \
 	--output text \
 	--stack-name buildkite \
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,5 @@
-
 .PHONY: all clean build build-ami upload create-stack
-AWS_PROFILE ?= default
+
 BUILDKITE_STACK_BUCKET ?= buildkite-aws-stack
 
 all: setup build
@@ -17,7 +16,6 @@ build/aws-stack.json: $(wildcard templates/*.yml)
 
 setup:
 	bundle check || ((which bundle || gem install bundler --no-ri --no-rdoc) && bundle install --path vendor/bundle)
-	cp config.json.example config.json
 
 clean:
 	-rm -f build/*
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -14,30 +16,40 @@ build/aws-stack.json: $(wildcard templates/*.yml)
 	bundle exec cfoo $^ > $@
 
 setup:
-	which bundle || gem install bundler --no-ri --no-rdoc
-	bundle install --path vendor/bundle
+	bundle check || ((which bundle || gem install bundler --no-ri --no-rdoc) && bundle install --path vendor/bundle)
+	cp config.json.example config.json
```
lox: Not sure this is wise, perhaps it could just fail if config.json doesn't exist with a helpful message?

lox:
Thanks, reviewing.

Might I suggest https://github.com/99designs/aws-vault for profile management? :)

Makefile
```diff
@@ -14,30 +16,40 @@ build/aws-stack.json: $(wildcard templates/*.yml)
...
 	aws cloudformation create-stack \
 	--output text \
-	--stack-name buildkite-$(shell date +%Y-%m-%d-%H-%M) \
+	--stack-name buildkite \
 	--disable-rollback \
 	--template-body "file://${PWD}/build/aws-stack.json" \
 	--capabilities CAPABILITY_IAM \
-	--parameters '$(shell cat config.json)'
+	--parameters '$(shell cat config.json)' \
+	--profile "$(AWS_PROFILE)"
```
lox: AFAIK, the aws-cli will read the AWS_PROFILE env anyway, so this shouldn't be needed.

Makefile
tobyjoe: Yeah, I think you're right. A nice "run this command and edit your conf!" message is a better idea. I'll patch that.

Makefile
tobyjoe: You're totally spot-on with that "AFAIK" - I'll tear those bits out! (does explain why my tests were FLAWLESS, though! heh)

tobyjoe:
Updated the branch in my fork with feedback. 😍

lox:
Looking good, I think with the new update changelist stuff in cloudformation updating the stack makes more sense.

README.md
```diff
@@ -21,6 +21,17 @@ aws cloudformation create-stack \
   --parameters <(cat config.json)
 ```
 
+
+Alternately, if you prefer to use this repo, clone it and run the following command to set up things locally and create a remote stack.
+```bash
+make && make create-stack
```
lox: This is redundant, just `make create-stack` will do it. By itsself `make` just uses the first target, which is build, which ends up being a dependency of `create-stack` anyway.

README.md
```diff
@@ -21,6 +21,17 @@ aws cloudformation create-stack \
...
+
+# If you need to specify a profile:
```
lox: I'd be tempted to just replace this with something like:

"You can use any of the `AWS...` environment variables that the aws-cli supports."

Makefile
```diff
@@ -14,30 +15,41 @@ build/aws-stack.json: $(wildcard templates/*.yml)
...
+
 	aws cloudformation create-stack \
 	--output text \
-	--stack-name buildkite-$(shell date +%Y-%m-%d-%H-%M) \
+	--stack-name buildkite \
 	--disable-rollback \
 	--template-body "file://${PWD}/build/aws-stack.json" \
 	--capabilities CAPABILITY_IAM \
-	--parameters '$(shell cat config.json)'
+	--parameters '$(shell cat config.json)' \
```
lox: Trailing back slash? On the update too.

Makefile
```diff
@@ -14,30 +15,41 @@ build/aws-stack.json: $(wildcard templates/*.yml)
...
 
 build-ami:
 	cd packer/; packer build buildkite-ami.json
 
 upload: build/aws-stack.json
-	aws s3 sync --acl public-read build s3://buildkite-aws-stack/
+	aws s3 sync --acl public-read build s3://${BUILDKITE_STACK_BUCKET}/
 
 create-stack: templates/mappings.yml build/aws-stack.json
+	test -s config.json || { echo "Please create a config.json file"; exit 1; }
```
lox: Could this just be its own target?

lox:
Thank you @tobyjoe

PR #54, opened 2016-05-05
Support increments for scaling up and down

lox:
As part of a discussion with @arthens and @sj26, we discovered that the stack scale up fires quicker than instances can actually launch. This slows down individual scale up events with a longer cooldown, but introduces a scale up and down increment so that the stack can still quickly scale up and scale down slowly.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -112,12 +112,12 @@ Parameters:
     Default: 0
 
   ScaleOutAdjustment:
-    Description: The number of agents to add on each scale out event
+    Description: The number of agents to add on each scale out event (ScheduledJobsCount > 0 for 1 minute)
     Type: Number
     Default: 5
 
   ScaleInAdjustment:
-    Description: The number of agents to remove on each scale in event
+    Description: The number of agents to remove on each scale in event (ScheduledJobsCount == 0 for 5 minutes)
     Type: Number
     Default: -2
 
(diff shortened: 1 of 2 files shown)


templates/buildkite-elastic.yml
```diff
@@ -109,6 +111,16 @@ Parameters:
     Type: Number
     Default: 0
 
+  ScaleOutAdjustment:
+    Description: The number of agents to add on each scale out event
+    Type: Number
+    Default: 5
+
+  ScaleInAdjustment:
+    Description: The number of agents to remove on each scale in event
```
toolmantim: The number of agents to add on each scale out event ("ScheduledJobsCount > 0 for 1 minute")

The number of agents to remove on each scale in event ("ScheduledJobsCount == 0 for 5 minutes")

toolmantim: Sorry, I was also meant to say, what do you think about describing the conditions here… just so people know what they're tweaking?

toolmantim:
@lox I just updated the labels. Let's merge this! (we need faster scale out for our Rails app)

toolmantim:
(we have 50 parallel jobs per build)

lox:
Technically, I haven't tested it results in faster scale outs :)

toolmantim:
Yeah… so we'd need 10 scale ups if it was 5 at a time every 5 minutes. That's 50 minutes of cooldowns. I think we'd add 25 a time perhaps? So it's max 10 minutes wait. Then scale down might be 10 at a time…

PR #55, opened 2016-05-05
Add lifecycled for hooking shutdown events gracefully

lox:
This is take 2 on #38. Since then I've built https://github.com/lox/lifecycled, so this simply installs that and saves a lot of complexity.


toolmantim:
let's do this! 👍

toolmantim:
whoops… accidental close…

lox:
WIP no more. Just doing some final testing.

PR #56, opened 2016-05-06
Fix inverted scaling policy

bgentry:
The scale up policy was configured to remove instances, and the scale down policy was configured to add them.

Introduced in 005807d6d348488aff5796af2b28adb45e23c3d6

### DIFF
--- a/templates/autoscale.yml
+++ b/templates/autoscale.yml
@@ -6,7 +6,7 @@ Resources:
       AdjustmentType: ChangeInCapacity
       AutoScalingGroupName: $(AgentAutoScaleGroup)
       Cooldown: 300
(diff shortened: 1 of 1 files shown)


bgentry:
btw, I'm thinking it would make sense to rename these from `ScaleInAdjustment` / `ScaleOutAdjustment` to `ScaleUpAdjustment` / `ScaleDownAdjustment` just to make the variable names align more closely with the context they're being used in. Might prevent similar mistakes in the future.

lox:
Face palm.

lox:
Agreed with the name changes @bgentry, I was trying to match the AWS event names.

PR #60, opened 2016-05-08
Use the latest agent tag in the packer hash

toolmantim:
The current packer build doesn't detect a new buildkite agent being available. This adds the latest agent release tag to the packer hash, so when there's a new tag we'll get a new AMI built and an updated stack JSON.


lox:
Thanks @toolmantim 👌

PR #66, opened 2016-05-11
Publish mappings.yml and aws-stack.json for branches

lox:
This changes how `templates/mappings.yml` is pulled down, previously in development it was easy to accidentally use the master AMIs and end up with a mixed result. This now allows for a easier workflow when on a branch. 

Now to create a stack locally, you need to either run `make download-mappings` first or `make build-ami`.


toolmantim:
Nice!

PR #68, opened 2016-05-12
[WIP] Publish unstable and experimental agent stacks

toolmantim:


toolmantim:
Let's just embed the 3 binaries in the AMI…

PR #69, opened 2016-05-16
Use docker user namespacing

toolmantim:
Rebases #32 against master and removes the old permission fix hook.

See #32 for details on the what, why and caveats

### DIFF
--- a/packer/buildkite-ami.json
+++ b/packer/buildkite-ami.json
@@ -24,11 +24,11 @@
     },
     {
       "type": "shell",
-      "script": "scripts/install-docker.sh"
+      "script": "scripts/install-buildkite.sh"
     },
     {
       "type": "shell",
-      "script": "scripts/install-buildkite.sh"
+      "script": "scripts/install-docker.sh"
     },
     {
       "type": "shell",
--- a/packer/conf/docker.conf
+++ b/packer/conf/docker.conf
@@ -2,4 +2,4 @@
 # running containers.  The default value of 1048576 mirrors the value
 # used by the systemd service unit.
 DAEMON_MAXFILES=1048576
-OPTIONS="--default-ulimit nofile=1024:4096 -s overlay"
+OPTIONS="--default-ulimit nofile=1024:4096 -s overlay --userns-remap=buildkite-agent"
--- a/packer/conf/subgid
+++ b/packer/conf/subgid
@@ -1,0 +1,1 @@
+buildkite-agent:496:65536
--- a/packer/conf/subuid
+++ b/packer/conf/subuid
@@ -1,0 +1,1 @@
+buildkite-agent:498:65536
--- a/packer/scripts/install-buildkite.sh
+++ b/packer/scripts/install-buildkite.sh
@@ -9,8 +9,7 @@ gpgcheck=0
 priority=1
 EOF
 
-sudo yum -y -q install buildkite-agent
-sudo usermod -a -G docker buildkite-agent
+sudo yum -y install buildkite-agent
 
 # Allow buildkite to fix checkout permissions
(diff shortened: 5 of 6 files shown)


toolmantim:
@lox @Vektah I'm going to give this a spin in prod today…

Vektah:
I think there are issues with uid > 1 at the moment. It's currently fine
for the root user inside the container but any other accounts don't get
squashed.

lox:
Yup, came here to say the same as @Vektah, think it would be sensible to keep the permissions resetting stuff.

toolmantim:
What if we reinstate the permissions hook, add a boolean to an “Experimental Features” section of the CF parameters to turn this on?

Is there a fix for the user problem you mention? :|

lox:
Sounds good to me @toolmantim. I'll leave it up to @Vektah to explain the issue, I roughly understand it, but whether it's a bug or an unexpected side-effect of the uid mapping I'm not sure of.

Vektah:
This stackoverflow has a decent description of the issue
http://stackoverflow.com/questions/35291520/docker-and-userns-remap-how-to-manage-volume-permissions-to-share-data-betwee

Basically I thought that userns could 'squash' the user IDs in the
container to a single one on the host, but it only lets you apply and
offset and limit. The way it's currently set its offsetting so root =
buildkite-agent and limit =1 but as soon as anything runs as a non root
user in a container it dies.

This means it's good as a security feature but not a great fit for this use.

Vektah:
I think a complete solution would involve some filesystem acls to force
uid/gid + userns to give security from the underlying container.
On 17 May 2016 6:11 PM, "Adam Scarr" <adam@vektah.net> wrote:

This stackoverflow has a decent description of the issue
http://stackoverflow.com/questions/35291520/docker-and-userns-remap-how-to-manage-volume-permissions-to-share-data-betwee

Basically I thought that userns could 'squash' the user IDs in the
container to a single one on the host, but it only lets you apply and
offset and limit. The way it's currently set its offsetting so root =
buildkite-agent and limit =1 but as soon as anything runs as a non root
user in a container it dies.

This means it's good as a security feature but not a great fit for this use.

toolmantim:
Hmm… sounds like it's good to add for security purposes then, but then still have the fix permissions script. And we can make it the default? (no need for a parameter)

lox:
I'd say we can eventually, I'm just worried about what undiscovered issues there are.

toolmantim:
Ok… so how about we keep the fix permissions and put what's here behind a "Experimental Features" parameter group for testing?

Vektah:
I think to do this for security purposes you would want to bump the uid and gid start ranges up, otherwise they will collide with other users. For example if you had user id = 2 in the container any files it writes will be created as uid = 500 on the host (buildkite-agent = 498 + 2)

Or use `--userns-remap="default"` and let docker create a throw away user, which would be the same thing.

I don't think its usable without some filesystem level squashing though.

lox:
Yeah, I think `--userns-remap=default` is probably the best way forward, there are too many other question marks for me with mapping into an existing range. Then we just flatten permissions after commands run via the current sudo approach.

lox:
Although then we need a way for the container to read the stuff owned by buildkite-agent. Blargh.

Docs here btw https://docs.docker.com/engine/reference/commandline/daemon/#daemon-user-namespace-options

lox:
Perhaps we could just generate a `/etc/subuid/etc/subuid with the root user mapped to `buildkite-agent` and then the a whole lot of entries for subsequent users with high, throwaway UID's?

lox:
The other major gotcha with this is that `--privileged` wont be supported.

Vektah:
Could use http://bindfs.org --force-user

toolmantim:
I think we should just close this for now, until we're ready to give it a stab again

PR #70, opened 2016-05-17
Allow you to select an agent release

toolmantim:
This adds the ability to switch to unstable and experimental agent builds, updates to Docker Compose 1.7.1, cleans up the agent meta-data by removing the EC2 tags, adds a `aws-buildkite-stack` meta-data flag so we can see who's using this, and reorgs much of the packer install files and folder structure.


packer/scripts/install-docker.sh
```diff
@@ -13,7 +13,7 @@ sudo cp /tmp/conf/docker.conf /etc/sysconfig/docker
 # Overwrite the yum packaged docker with the latest
 # Releases can be found at https://github.com/docker/docker/releases
 # shasums can be found at $URL.sha256
-wget https://get.docker.com/builds/Linux/x86_64/docker-$DOCKER_VERSION.tgz -O /tmp/docker.tgz
+wget -nv https://get.docker.com/builds/Linux/x86_64/docker-$DOCKER_VERSION.tgz -O /tmp/docker.tgz
```
lox: For the sake of consistency, could we use `sudo curl -Lf -o ` here?

packer/scripts/install-lifecycled.sh
```diff
@@ -13,7 +13,7 @@ sudo curl -Lf -o /etc/init/lifecycled.conf \
 cat << EOF | sudo tee /usr/bin/buildkite-lifecycle-handler
 #!/bin/sh -eu
 echo "stopping buildkite-agent gracefully"
-service buildkite-agent stop
+pgrep buildkite-agent | xargs kill
```
lox: What's the reason behind this change @toolmantim?

templates/buildkite-elastic.yml
```diff
@@ -50,6 +51,11 @@ Parameters:
     Description: Your Buildkite organization's slug
     Type: String
 
+  BuildkiteAgentRelease:
+    Description: The buildkite-agent release stream ("stable", "unstable" or "beta")
```
lox: unstable or experimental or edge?? :)

templates/buildkite-elastic.yml
```diff
@@ -314,18 +320,28 @@ Resources:
...
+                INSTANCE_ID=\$(curl -sf http://169.254.169.254/latest/meta-data/instance-id)
+                DOCKER_VERSION=\$(docker --version | cut -f3 -d' ' | sed 's/,//')
+                BUILDKITE_AGENT_BINARY=/usr/bin/buildkite-agent-$(BuildkiteAgentRelease)
+                if [[ "$(BuildkiteAgentRelease)" == "stable" ]]; then
+                  BOOTSTRAP_SCRIPT=/etc/buildkite-agent/2-1-stable-bootstrap.sh
+                else
+                  BOOTSTRAP_SCRIPT="\$BUILDKITE_AGENT_BINARY bootstrap"
+                fi
+
+                cat << EOF > /etc/buildkite-agent
```
lox: Sooooo much better. Why didn't I just do this in the first place?? :)

packer/scripts/install-buildkite.sh
```diff
@@ -1,17 +1,35 @@
...
+sudo chmod +x /usr/bin/buildkite-agent-experimental
+
+sudo useradd buildkite-agent
 sudo usermod -a -G docker buildkite-agent
 
+sudo mkdir -p /etc/buildkite-agent/hooks
+sudo chown -R buildkite-agent: /etc/buildkite-agent/hooks
+
+# This can be removed when stable refers to 3.0
+sudo wget -nv https://raw.githubusercontent.com/buildkite/agent/2-1-stable/templates/bootstrap.sh -O /etc/buildkite-agent/2-1-stable-bootstrap.sh
```
lox: curl for consistency (unless there is some reason wget is better here)

templates/buildkite-elastic.yml
```diff
@@ -314,18 +320,28 @@ Resources:
...
+                name="$(AWS::StackName)-\$INSTANCE_ID"
+                token="$(BuildkiteAgentToken)"
+                meta-data=\$(printf 'queue=%s,docker=%s,stack=%s,aws-buildkite-stack' "$(BuildkiteQueue)" "\$DOCKER_VERSION" "$(AWS::StackName)")
+                meta-data-ec2=true
+                meta-data-ec2-tags=true
```
lox: Perhaps this might have been a bad idea, we're already providing most of the useful tags. I feel like if we took this approach we should also be tagging the instances. Outside of the scope of this PR, but thoughts?

lox:
Think this is great, although I wonder if it might be better to symlink the binary into place rather than switch it?

templates/buildkite-elastic.yml
toolmantim: ohlulz

templates/buildkite-elastic.yml
toolmantim: Yeah, I reckon we could ditch them… and later, if someone requests it, we could add it as an optional CF parameter later

packer/scripts/install-docker.sh
toolmantim: Yep cool, I noticed that too

templates/buildkite-elastic.yml
```diff
@@ -314,18 +320,28 @@ Resources:
...
+                hooks-path=/etc/buildkite-agent/hooks
+                build-path=/var/lib/buildkite-agent/builds
+                plugins-path=/var/lib/buildkite-agent/plugins
+                bootstrap-script=\$BOOTSTRAP_SCRIPT
+                EOF
+
+                su - buildkite-agent -c "\$BUILDKITE_AGENT_BINARY start --config /etc/buildkite-agent'
```
toolmantim: @lox is this the right way to start a thing as `buildkite-agent`?

toolmantim:
Yeah, symlink would work too…

packer/scripts/install-lifecycled.sh
toolmantim: The reason for this is because there's no longer a buildkite-agent system service at all…

lox:
So is the thinking to ditch any form of init script and just run the buildkite binary in the userdata bootstrap? My concern is that if there is ever an error or transient failure that causes the agent to crash, it leaves us with instances without agents and currently no mechanism to correct that.

lox:
I suspect you will need to add an extra IAM permission for setting instance health @toolmantim

toolmantim:
Ok, this is ready to rumble @lox @Vektah. This shouldn't have any changes for stack users… should be a simple update and provides the extra parameter.

lox:
Sorry for not giving more feedback on this one @toolmantim.

templates/buildkite-elastic.yml
```diff
@@ -307,22 +314,38 @@ Resources:
...
-                ami_id=\$(curl -sf http://169.254.169.254/latest/meta-data/ami-id)
-                docker_version=\$(docker --version | cut -f3 -d' ' | sed 's/,//')
-                meta_data_tags=\$(printf 'queue=%s,docker=%s,stack=%s,aws-buildkite-stack,aws:instance-id=%s,aws:instance-type=%s,aws:ami-id=%s' "$(BuildkiteQueue)" "\$docker_version" "$(AWS::StackName)" "\$instance_id" "\$instance_type" "\$ami_id")
+                #!/bin/bash -eu
 
-                sed -i -r "s/^(name)=.*/\\1=\"$(AWS::StackName)-\$instance_id\"/" /etc/buildkite-agent/buildkite-agent.cfg
-                sed -i -r "s/^(token)=.*/\\1=\"$(BuildkiteAgentToken)\"/" /etc/buildkite-agent/buildkite-agent.cfg
-                sed -i -r "s/^(# )?(meta-data-ec2-tags)=.*/\\2=true/" /etc/buildkite-agent/buildkite-agent.cfg
-                sed -i -r "s/^(# )?(meta-data)=.*/\\2=\"\$meta_data_tags\"/" /etc/buildkite-agent/buildkite-agent.cfg
+                ln -s /usr/bin/buildkite-agent-$(BuildkiteAgentRelease) /usr/bin/buildkite-agent
```
lox: 👍

lox:
Looked everything over, looks spot on. Nice one.

toolmantim:
@lox all good! Thanks for the look-over 👍

PR #72, opened 2016-05-27
Scale down base on idle workers

Vektah:
This scales down one agent when all agents were idle for 6 minutes, repeating every 2 minutes.  

The old logic would scale down 5 agents if there were no *waiting* for an hour, every 10 minutes.

The scheduled agents check is broken because during normal operation there often arent any scheduled builds, and during quiet (lower than usual, but still active) periods the cluster will scale to zero. This is made worse by the 10 minute cooldown after scaling, which prevents new machines from coming up.

We were seeing this scale down behaviour over lunch. To add insult to injury it means every docker cache is now totally cold, so builds after lunch can take 20+ minutes longer than they should.

A few tweaks to the scale down behaviour:
 - much lower cooldown, scaling down would prevent scaling up for 10 minutes, after scaling to zero this means builds spend a lot of time with zero agents, waiting.
 - scale down in smaller increments more often, ties into a lower cooldown

see https://github.com/buildkite/buildkite-aws-stack/issues/58

before:
annotated-bk-scaledown

after:
2016-05-27_1190x603

I don't think this is perfect, as it leaves our machines a little underutilized but I think simply tweaking the max agents will help a lot. An alternative would be to do scaling based on the percentage of busy agents, but there is also docker cache reusue to think about. keeping a fairly stable number of machines during the day is a big win for performance.

### DIFF
--- a/templates/autoscale.yml
+++ b/templates/autoscale.yml
@@ -23,14 +23,14 @@ Resources:
       ScalingAdjustment : $(ScaleUpAdjustment)
 
   AgentScaleDownPolicy:
-    Type : AWS::AutoScaling::ScalingPolicy
+    Type: AWS::AutoScaling::ScalingPolicy
     Properties:
       AdjustmentType: ChangeInCapacity
       AutoScalingGroupName: $(AgentAutoScaleGroup)
-      Cooldown : 600
+      Cooldown : 60
       ScalingAdjustment: $(ScaleDownAdjustment)
 
-  ScheduledJobsAlarmHigh:
+  AgentUtilizationAlarmHigh:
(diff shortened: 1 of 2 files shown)


pda:
Sounds good.

lox:
@bgentry thoughts?

lox:
Build failure was transient spot instance quota issue.

lox:
@toolmantim any chance you could look into the above build failure?

lox:
Otherwise this looks good to me, happy to merge 👍

bgentry:
LGTM!

toolmantim:
Fixed the build, and merging 👍

PR #76, opened 2016-07-01
Install configurations for N number of buildkite agents per instance

joffotron:
This patch adds in the ability to configure multiple buildkite-agents per EC2 instance. 

This is a configuration I've seen at 3 different companies now, although in those cases, they've either been manually configured, or taken something like the CF stack and then modified post-deployment.

Some people want to get better utilization of the hardware they are running BK on, especially in the case where bigger instances are often multi-core, but their builds may be single-threaded.

Was going to keep this addition as an 'internal' fork, but following on from discussion/help from @toolmantim on Slack, have made a PR.

### DIFF
--- a/packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
+++ b/packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
@@ -1,3 +1,21 @@
 #!/bin/bash
 
-service buildkite-agent stop
+set -eu -o pipefail
+eval "$(cat /var/lib/buildkite-agent/cfn-env)"
+
+echo "Stopping buildkite-agent gracefully"
+
+service buildkite-agent stop &
+
+if [[ ${BUILDKITE_AGENTS_PER_INSTANCE} > 1 ]]; then
+  for i in $(seq 1 $(( ${BUILDKITE_AGENTS_PER_INSTANCE} - 1 )) ); do
+    service "buildkite-agent-${i}" stop &
+  done
+fi
+
+until pgrep buildkite-agent > /dev/null; do
+  echo "Waiting for buildkite-agent processes to finish..."
+  sleep 0.5
+done
+
+echo "buildkite-agent stopped"
--- a/packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
+++ b/packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
@@ -1,17 +1,15 @@
 #!/bin/bash
 
 set -eu -o pipefail
+
+# Expose BUILDKITE_AGENTS_PER_INSTANCE
 eval "$(cat /var/lib/buildkite-agent/cfn-env)"
 
 echo "Stopping buildkite-agent gracefully"
 
-service buildkite-agent stop &
-
-if [[ ${BUILDKITE_AGENTS_PER_INSTANCE} > 1 ]]; then
-  for i in $(seq 1 $(( ${BUILDKITE_AGENTS_PER_INSTANCE} - 1 )) ); do
-    service "buildkite-agent-${i}" stop &
-  done
-fi
+for i in $(seq 1 "${BUILDKITE_AGENTS_PER_INSTANCE}"); do
+  service "buildkite-agent-${i}" stop &
+done
 
 until pgrep buildkite-agent > /dev/null; do
   echo "Waiting for buildkite-agent processes to finish..."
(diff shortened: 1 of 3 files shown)


templates/buildkite-elastic.yml
```diff
@@ -334,20 +340,23 @@ Resources:
...
-                token="$(BuildkiteAgentToken)"
-                meta-data=\$(printf 'queue=%s,docker=%s,stack=%s,buildkite-aws-stack' "$(BuildkiteQueue)" "\$DOCKER_VERSION" "$(AWS::StackName)")
-                meta-data-ec2=true
-                bootstrap-script="\$BOOTSTRAP_SCRIPT"
-                hooks-path=/etc/buildkite-agent/hooks
-                build-path=/var/lib/buildkite-agent/builds
-                plugins-path=/var/lib/buildkite-agent/plugins
-                EOF
+                for kite in {1..$(BuildkiteAgentsPerInstance)}; do
+                  cat << EOF > /etc/buildkite-agent/buildkite-agent-${kite}.cfg
```
toolmantim: we can actually keep the config out of this loop, and use `name="$(AWS::StackName)-%n-\$INSTANCE_ID"` and the buildkite server will give it a numeric number. No need to change the builds directory either, because the agent name given by the server is used for the checkout directory /within/ the builds directory.

templates/buildkite-elastic.yml
```diff
@@ -71,6 +72,11 @@ Parameters:
     Type: String
     Default: elastic
 
+  BuildkiteAgentsPerInstance:
+    Description: How Many Buildkite agents to run per instance/machine
```
toolmantim: The number of Buildkite agents to run on each instance

toolmantim: Perhaps we should put this param down with the autoscale rules?

* Number of agents per instances
* Min number of instances
* Max number of instances
* etc

Actually I can modify the param structuring once we merge it to master! Let's just get it working first :)

joffotron:
Looks like it's working, in terms of bootstrapping the machine, and starting the agents. Still need to chuck some builds at it, and test the shutdown script, but that works when invoked manually.

image

image

lox:
This will affect the auto scaling rules substantially I think. Will eye ball the code soon.

joffotron:
@lox Agreed. Better to do some maths for the user here? Or caveat emptor??

packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
```diff
@@ -1,3 +1,21 @@
 #!/bin/bash
 
-service buildkite-agent stop
+set -eu -o pipefail
+eval "$(cat /var/lib/buildkite-agent/cfn-env)"
```
lox: ?

lox:
I'm really undecided on this. On one hand, it makes for much better utilization of the hardware, but on the other it adds significant complexity to things like auto-scaling. Could we trouble you to document a bit more of the _why_ in the PR description @joffotron?

packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
joffotron: The cfn-env file now contains the env var for $BUILDKITE_AGENTS_PER_INSTANCE. 

The changes to this file are so the ASG can hold off spinning down the instance, in the case that any of the agents have a job still running.

lox:
Yeah, after some thought, it's a +1 from me, but we should update the maths for the auto-scaling rules.

templates/buildkite-elastic.yml
```diff
@@ -347,7 +354,16 @@ Resources:
 
                 chown buildkite-agent: /etc/buildkite-agent/buildkite-agent.cfg
 
+                if [[ $(BuildkiteAgentsPerInstance) > 1 ]]; then
+                  for i in \$(seq 1 \$(( $(BuildkiteAgentsPerInstance) - 1 )) ); do
+                    cp /etc/init.d/buildkite-agent /etc/init.d/buildkite-agent-\${i}
+                    service buildkite-agent-\${i} start
+                    chkconfig --add buildkite-agent\${i}
```
toolmantim: Am I loco or is this missing a dash? (`chkconfig --add buildkite-agent-\${i}`)

joffotron: Yep, missing the dash. 👍

toolmantim:
So it turns out the autoscaling rules work just fine, because they're based on `ScheduledJobs > 0` and `IdleAgents == 0` — both of which aren't affected by the number of agents per instance.

Have updated the PR and ready to merge 👍

PR #79, opened 2016-07-06
Remove 'github' from the secrets bucket SSH key name

toolmantim:
There's nothing Github specific about this key at all, so let's not include it in the name

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,8 +76,8 @@ The following environment variables can be set on the Buildkite pipeline or step
 * `BUILDKITE_SECRETS_BUCKET` - the name of the S3 bucket where secrets are stored. Default: the value set in the stack parameter when the stack was created. Example: `my-secrets-bucket`
 * `BUILDKITE_SECRETS_KEY` - the encryption key used to decrypt objects from the secrets bucket. Default: nil. Example: `w2Uzhc4kXXbW//T9zaY3neoCbR9roQ10`
 * `BUILDKITE_SECRETS_PREFIX` - the folder within the secrets bucket. Default: the build pipeline's slug. Example: `my-great-pipeline`
-* `SSH_KEY_NAME` - the filename of the SSH key inside this pipeline’s folder in the secrets bucket. Default: `id_rsa`. Example: `id_rsa_github`
-* `SHARED_SSH_KEY_NAME` - the filename of the SSH key in the root of the secrets bucket if there's no pipeline-specific SSH key present. Default: `id_rsa`. Example: `id_rsa_github_shared`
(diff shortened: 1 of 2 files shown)


lox:
Is this a breaking change for existing installs?

toolmantim:
Yep, but I was thinking of adding a fallback

toolmantim:
@lox so I've added a fallback for people still on the `_github` version, along with support for a top level SSH key for people when per-pipeline is overkill. How's that look?

README.md
```diff
@@ -2,29 +2,30 @@
...
+* CloudWatch metrics from the Buildkite API
+* Support for stable, unstable or experimental Buildkite Agent releases
 
 ## Getting Started
 
 The easiest way is to launch the latest built version via this button:
 
 [![Launch Buildkite AWS Stack](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/images/cloudformation-launch-stack-button.png)](https://console.aws.amazon.com/cloudformation/home#/stacks/new?stackName=buildkite&templateURL=https://s3.amazonaws.com/buildkite-aws-stack/aws-stack.json)
 
+> Although the stack will create it's own VPC by default, we highly recommend following best practice by setting up a separate development AWS account and using role switching and consolidated billing—see the [Delegate Access Across AWS Accounts tutorial](http://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_cross-account-with-roles.html) for more information.
```
lox: 👌

README.md
```diff
@@ -49,42 +50,73 @@ make create-stack
...
 
-You should encrypt your objects with a project-specific key and provide it in `BUILDKITE_SECRETS_KEY` which will be used to decrypt all the files found in the secrets bucket.
+## Secrets Bucket Support
 
-### Creating a new project
+The stack has a `SecretsBucket` parameter which will allow your build agents to automatically get access to SSH private keys and environment hooks for exposing environment variables to builds. The stack doesn't create the bucket for you, you need to do this yourself, but it does create a role that gives read access to the build machines. 
+
+The secrets bucket can contain the following files:
+
+* `/id_sra` - An optional private key to use Git SSH operations when there is no pipeline-specific key present
```
lox: Typo?

lox:
Thoughts on swapping to something like `ssh_private_key` rather than `id_rsa`? It could very well be dsa or ECC.

toolmantim:
Yeah that's a good idea. It's not even really specific to git is it? `ssh_key` could be fine too?

Also I noticed a lot of people use netrc and GitHub API keys with read repo instead of SSH keys. I had pondered how to support that… some other PR, just figured I'd plant the seed

README.md
toolmantim: Whoops

toolmantim:
Ok, changed it to `private_ssh_key` and fixed the Readme. Will merge when green…

PR #80, opened 2016-07-06
Fix agent log capturing for n agents

toolmantim:
In #76 we broke CloudWatch buildkite-agent logs. This fixes them again.


lox:
We should check these logs as part of the CI tests.

toolmantim:
Yep we totes should

PR #82, opened 2016-07-07
Only docker-gc images older than 7 days

### LINKED ISSUE #78: Ability to customize gc-docker rules
At the moment the hourly cleaning of Docker images is a bummer for caches… especially on our builder machine. I'd be neat to support a S3 file containing the gc-docker rules… or perhaps just paste them in as a stack parameter?

toolmantim:
Only clean up docker images once a week.

Closes #78

### DIFF
--- a/packer/conf/docker/cron.daily/docker-gc
+++ b/packer/conf/docker/cron.daily/docker-gc
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# For parameters, see:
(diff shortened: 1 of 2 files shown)


lox:
Might it make more sense to do this as part of the environment hook if disk space is low?

toolmantim:
@lox yep, I wasn't sure how to do that with the volume type differences etc… but if it's easy, sounds like a better solution to me

lox:
Yeah, fair point. I'll think about it.

PR #83, opened 2016-07-11
Rename BuildkiteAgentsPerInstance to AgentsPerInstance

toolmantim:
@joffotron when auditing the param descriptions and grouping I realised we named this incorrectly to start with.

This renames it to `AgentsPerInstance` (and it now lives down in the instance config)


toolmantim:
This is a breaking change, but this only made it in less than a week ago, so I'll reach out to people to give them a heads up

PR #85, opened 2016-07-12
Use UnfinishedJobs not BusyAgentCount

toolmantim:
I've been trying to track down why we were getting scale down events whilst we still had running jobs, and it turns out the `BusyAgentCount` metric lags behind the jobs metrics quite a bit.

metrics

Recent updates the [metrics collector](https://github.com/buildkite/buildkite-cloudwatch-metrics-publisher) and [metrics stack](buildkite/buildkite-metrics#3) means we now have an `UnfinishedJobsCount` metric we can use to represent if there's any jobs being currently worked on.

This PR updates the scaledown so it's based on this new `UnfinishedJobsCount ` metric.

### DIFF
--- a/templates/autoscale.yml
+++ b/templates/autoscale.yml
@@ -36,7 +36,7 @@ Resources:
       AlarmDescription: Scale-up if ScheduledJobs > 0 for 1 minute
       MetricName: ScheduledJobsCount
       Namespace: Buildkite
-      Statistic: Average
+      Statistic: Maximum
       Period: 60
       EvaluationPeriods: 1
       Threshold: 0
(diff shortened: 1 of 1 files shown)


toolmantim:
This has been working well in prod 👍 Merging…

mtibben:
I was under the impression this was the reason for lifecycled - shouldn't lifecycled keep instances alive when there are active jobs?

toolmantim:
@mtibben lifecycled was doing the right thing okay, but there was a lag between these two metrics of over 5 minutes, which meant they weren't great for doing different parts of the autoscaling triggers… you want your metrics to line up in this case. Lifecycled was keeping things around okay, but the instances were being torn down even when they were "busy" because the metric was incorrect (because of lag).

I was going for stability for the release… the autoscaling rules still need some love as you've seen

PR #87, opened 2016-07-13
Fix buildkite-agent starting before cloud-init has run

toolmantim:
This updates the cloud-init to not actually start the buildkite agent services, but let the rc change do it, so they don't start accepting jobs before they're ready.


sj26:
Hopefully this fixes it! :+1:

toolmantim:
Seems not to be booting on instance start. I wonder if it's ultimately the problem though. It might be worth just looking more closely at the hung/failed case to see if it was in fact starting too early…

toolmantim:
This is probably not a thing…

PR #92, opened 2016-07-15
Fix the lifecycled hook from instantly returning

### LINKED ISSUE #91: Scale down doesn't let jobs run gracefully
When a scale down happens I'm finding that a job is aborting mid-test-run with `ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?`, and the last thing in the buildkite-agent CloudWatch logs is "Gracefully shutting down" but no more logs, but the job successfully uploads artifacts etc after failing, and gracefully shuts down via our API, but is missing any more log lines in CloudWatch.

My guess is that we're not preventing `awslogs` and `docker` daemons being shutdown in the autoscale terminating event. `lifecycled` appears only to be helping `buildkite-agent` from being shut down.

toolmantim:
In #76 we introduced a bug where the lifecycled hook would instantly return, even if the buildkite-agent process had not finished gracefully shutting down. This caused the scale in event to continue, shutting down the machine mid-build… causing failing build jobs, and strange errors such as `could not connect to Docker host`

This fixes it by changing `until` to `while`.

Our tests unfortunately don't cover the the scale-in lifecycled hook.

Fixes #91

### DIFF
--- a/packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
+++ b/packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
@@ -11,9 +11,9 @@ for i in $(seq 1 "${BUILDKITE_AGENTS_PER_INSTANCE}"); do
   service "buildkite-agent-${i}" stop &
 done
 
-until pgrep buildkite-agent > /dev/null; do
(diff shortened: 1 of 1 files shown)


shuber:
Awesome, thanks!

PR #98, opened 2016-07-19
Increase scale in to 30 minutes

toolmantim:
Given instances are min 1 hour billing, and cycling agents during the middle of the day can be a bit of PITA for developers, increasing it to 30 minutes of inactivity is a bit more friendly.

Also there's no need to wait 5 full minutes from the scale in finishing to do the next scale in event, agents should pick up work almost before they're fully marked as in service.

### DIFF
--- a/templates/autoscale.yml
+++ b/templates/autoscale.yml
@@ -19,7 +19,7 @@ Resources:
     Properties:
       AdjustmentType: ChangeInCapacity
       AutoScalingGroupName: $(AgentAutoScaleGroup)
-      Cooldown: 300
+      Cooldown: 120
       ScalingAdjustment : $(ScaleUpAdjustment)
 
   AgentScaleDownPolicy:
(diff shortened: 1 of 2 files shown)


templates/autoscale.yml
```diff
@@ -49,12 +49,12 @@ Resources:
...
    Properties:
-      AlarmDescription: Scale-down if UnfinishedJobs == 0 for 15 minutes
+      AlarmDescription: Scale-down if UnfinishedJobs == 0 for 30 minutes
       MetricName: UnfinishedJobsCount
       Namespace: Buildkite
       Statistic: Maximum
-      Period: 900
-      EvaluationPeriods: 1
+      Period: 300
+      EvaluationPeriods: 6
```
lox: Doesn't this change it from `900*1` to `6 * 300`, e.g 15 minutes to 30 minutes??

lox: Oh. That was your intention. Ignore me.

PR #100, opened 2016-07-22
Allow additional managed policies for container instances

### LINKED ISSUE #47: Add an option to set the IAM of the instances
Or some other way to give the machines extra IAM permissions that the builds may need.

exidy:
Adds the ability for the user to specify zero or more managed IAM policies to be associated with the agent IAM role. Fixes #47.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -19,6 +19,7 @@ Metadata:
         - Subnets
         - AvailabilityZones
         - SecurityGroupId
+        - ManagedPolicyArns
 
       - Label:
           default: Instance Configuration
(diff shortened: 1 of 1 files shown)


lox:
I'd be tempted to reference an IAM Role, rather than IAM managed policies. It's easy enough to create a role with the managed policies in it, but this way you can also attach ad-hoc inline permissions to the Role if needed. Thoughts @exidy @toolmantim ?

exidy:
@lox not sure what you mean. Nothing about referencing a managed policy in the role stops you from also attaching ad-hoc inline policies to it.

I like this approach over simply allowing the user to specify a custom role because, as an instance can only have one role, a custom role would need to be a superset of the one the stack already creates, including the internal references to other resources.

The way I've done it the stack continues to manage the minimum policies it needs to function and the user can adds policies on top. If the set of policies in the stack changes the user can simply update. They don't need to reverse-engineer the changes into their custom role.

lox:
An instance can only have one role?

lox:
Ah right: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-instanceprofile.html:

```
Roles:
Type: List of references to AWS::IAM::Roles. Currently, a maximum of one role can be assigned to an instance profile.
```

Fair enough then :)

lox:
Ok then, I completely agree, how you've done it seems best. Sorry!

exidy:
No apologies necessary! I'm glad someone is checking my work. 😄

toolmantim:
Thanks @exidy!

PR #101, opened 2016-07-25
Have the lifecycled handler script look for just the build agent it is trying to shut down.

joffotron:
Otherwise, it gets stuck as other agents are still running, and matching
the pgrep

terminate failing

### DIFF
--- a/packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
+++ b/packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
@@ -11,7 +11,7 @@ for i in $(seq 1 "${BUILDKITE_AGENTS_PER_INSTANCE}"); do
   service "buildkite-agent-${i}" stop &
 
   # Need to ensure it's the buildkite-agent user, so it doesn't match this lifecycld handler script
-  while pgrep -u buildkite-agent buildkite-agent > /dev/null; do
+  while pgrep -u buildkite-agent buildkite-agent-${i} > /dev/null; do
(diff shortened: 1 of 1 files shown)


packer/conf/buildkite-agent/scripts/buildkite-agent-lifecycled-handler
```diff
@@ -11,7 +11,7 @@ for i in $(seq 1 "${BUILDKITE_AGENTS_PER_INSTANCE}"); do
   service "buildkite-agent-${i}" stop &
 
   # Need to ensure it's the buildkite-agent user, so it doesn't match this lifecycld handler script
-  while pgrep -u buildkite-agent buildkite-agent > /dev/null; do
+  while pgrep -u buildkite-agent buildkite-agent-${i} > /dev/null; do
```
toolmantim: Nice, thanks! Good catch. Can you add double quotes around that just for bash listing and it's good for merge party?

toolmantim:
Thanks so much @joffotron!

lox:
Hold on a min, will that work? Isn't the process name still `buildkite-agent`, regardless of whether it's a different init instance? 

I'd be tempted to put the pgrep wait outside the for loop that kills them, then it will just wait until they are all dead, which is the expected behaviour.

toolmantim:
Ugh, sorry about this slip. I should have checked it more closely.

joffotron:
Yeah, good catch @lox. Not quite sure what I was thinking there. 
Need to investigate why the script as it stands was hanging indefinitely

PR #103, opened 2016-07-25
Send all the kills before waiting

toolmantim:
This changes the `service stop` calls to all happen at once, and then do the waiting… so that agents 2, 3, 4 etc don't pick up new work whilst waiting for previous agents to stop.

Related to #101


toolmantim:
@joffotron if you'd like to test this: https://s3.amazonaws.com/buildkite-aws-stack/kill-all-agents-then-wait/aws-stack.json https://s3.amazonaws.com/buildkite-aws-stack/kill-all-agents-then-wait/mappings.yml

joffotron:
Thanks @toolmantim - I've tested it out here, and it's working much, much better :) 👍

toolmantim:
Thanks @joffotron! Will merge now

PR #104, opened 2016-07-27
Name tags for instances

exidy:
This PR adds "Name" tags for buildkite agents such that their purpose is clear in the AWS console. It also adds tags for the agent version and queue to assist with managing multiple agent stacks.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -495,6 +495,9 @@ Resources:
         - Key: Role
           Value: buildkite-agent
           PropagateAtLaunch: true
+        - Key: Name
+          Value: buildkite-agent-$(BuildkiteAgentRelease)-$(BuildkiteQueue)
+          PropagateAtLaunch: true
 
     CreationPolicy:
       ResourceSignal:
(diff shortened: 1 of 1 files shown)


lox:
Yup, probably makes sense. Another way to do this is to provide tags when creating the stack (http://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack.html)

exidy:
Tagging the stack should tag the ASG - I wonder if the ASG will then propagate the tags to the instances?

toolmantim:
It does! But from memory you have to remember to use Name and not name, and I also sometimes forget to do it. So there's some merit to making this easier… because you almost always want them just to be the same name as the stack yeah?

exidy:
@toolmantim it's of particular significance for EC2 instances created via ASG via CloudFormation. Resources created directly get an autogenerated name which usually has the name of the stack in it already. As the EC2 instances aren't directly created by the stack, they don't benefit from this feature. Hence the explicit name tags.

toolmantim:
Yep cool. I was meaning that I normally achieve this via the stack tags (if you add `Name` as the stack tag your ASGs instances will be named correctly). But this is more fool proof and obvious 👍

Issue #41, opened 2016-04-15
Stack creation fails with errors about subnet availability


lox:
Sometimes stack creation fails with:

```
Value (us-east-1a) for parameter AvailabilityZones is invalid. Subnets can currently only be created in the following availability zones...
```

![image](https://cloud.githubusercontent.com/assets/15758/14551618/d68091dc-0318-11e6-8b7b-a77cd5c58ca3.png)

This is because if no `AvailabilityZones` is provided, the list of all availability zones is used, but only of these can have subnets created. Because each account is different (the availability zones are randomly named), a general solution is very difficult.

https://stackoverflow.com/questions/21390444/is-there-a-way-for-cloudformation-to-query-available-zones-for-subnet-creation

toolmantim:
FWIW deis/deis#1601 has this issue too

tristanoneil:
I was able to work around it by specifying `us-east-1a, us-east-1b, us-east-1d` for the AvailabilityZones option.

lox:
I swear I've worked in this exact bug several times now, but I am cursed to
forget the solution. I dimmly recall it's not fixable. Will do some
searching.

lox:
As @tristanoneil said, the fix to this is to provide a list of az's for the `AvailabilityZones` template option. This differ from account to account, so we can't automate this for now.

keithduncan:
This is quite an old issue with no recent activity. From some searching it seems the expected way to fix this is to have a [default VPC](https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc.html#create-default-vpc) and [default subnets](https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc.html) which [Fn::GetAZs will use](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getavailabilityzones.html) to populate the list, otherwise it will return all the AZs including invalid ones.

PR #105, opened 2016-07-27
Change parameter help text

exidy:
Looking at the generated CloudFormation it seems the AvailabilityZones parameter is only used if creating subnets.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -112,7 +112,7 @@ Parameters:
 
   AvailabilityZones:
     Type: CommaDelimitedList
-    Description: Optional - Comma-delimited list of VPC availability zones in which to create subnets. Required if not setting Subnets.
+    Description: Optional - defines the AZs that subnets are created in if Subnets parameter is not specified.
     Default: ""
 
   InstanceType:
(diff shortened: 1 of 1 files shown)


toolmantim:
Thanks for the change. The new version might cause some people to think they need to fill it out for the basic case (no VpcId) if they didn't fill out Subnets. But this is only required if VpcId is set but Subnets is not, right?

exidy:
No, if I'm understanding the stack correctly it's needed any time subnets are not specified. There's a fallback in there where it queries the region for the available AZs, but I think @lox had identified a failure case for that situation?

toolmantim:
@lox any input?

lox:
`AvailabilityZones` is entirely optional, and only needs to be used if you encounter the dreaded https://github.com/buildkite/elastic-ci-stack-for-aws/issues/41. 

I'd be inclined to phrase it "Optional - Comma-delimited list of VPC availability zones in which to create subnets. If not provided your account defaults are used, which are occasionally incorrect."

Is that too confusing?

exidy:
It's hard to explain what's going on in the help text without making it war and peace! How about "Optional - defines which AZs subnets are created in if Subnets are not specified." That should steer someone in the right direction if they hit #41.

lox:
Feel like making that change @exidy and we'll get this merged? :)

exidy:
Updated with a slightly more grammatical version of the above.

PR #106, opened 2016-07-30
Docker 1.12.1

toolmantim:
Bump to [Docker 1.12.1](https://github.com/docker/docker/releases/tag/v1.12.1)

### DIFF
--- a/packer/conf/docker/docker.conf
+++ b/packer/conf/docker/docker.conf
@@ -1,17 +1,2 @@
-# This is the same as the standard yum-installed Docker configuration,
-# but with the `-s overlay` added to the OPTIONS, to set it to use the
-# overlay storage driver instead of devicemapper.
-#
-# https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/
-#
-# When we switch to Docker 1.12 we should probably switch to overlay2
-
-# The max number of open files for the daemon itself, and all
-# running containers.  The default value of 1048576 mirrors the value
-# used by the systemd service unit.
-DAEMON_MAXFILES=1048576
-
-# Additional startup options for the Docker daemon, for example:
-# OPTIONS="--ip-forward=true --iptables=true"
-# By default we limit the number of open files per container
-OPTIONS="--default-ulimit nofile=1024:4096 -s overlay"
+# We use the overlay2 storage driver for performance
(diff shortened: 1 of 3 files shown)


toolmantim:
It appears there's no yum package for Docker 1.12.0 just yet

toolmantim:
Still not available…

bforchhammer:
The install-script on https://get.docker.com installs packages from `https://yum.dockerproject.org/`, which is probably up to date? Maybe we can borrow something from that script... or just run `curl -sSL https://get.docker.com/ | sh` of course, but then you won't be able to control the installed version, I think.

toolmantim:
@bforchhammer that's a good idea. We could just start an instance, run that installer, copy out the yum repo config, and just use that!

toolmantim:
Turns out that script just installs it from Amazon's yum repos. I managed to install it via the centos 6 yum repo package, but I'm not sure that is such a good idea…

```
cat >/etc/yum.repos.d/docker.repo <<-EOF
[dockerrepo]
name=Docker Repository
baseurl=https://yum.dockerproject.org/repo/main/centos/6
enabled=1
gpgcheck=1
gpgkey=https://yum.dockerproject.org/gpg
EOF
yum install docker-engine
```

ans0600:
Could we also upgrade to overlay2 here: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/conf/docker/docker.conf#L7

toolmantim:
I've just changed to performing a manual install of Docker, and switched to overlay2 whilst we were there. The stack json URL is https://s3.amazonaws.com/buildkite-aws-stack/docker-1-12-0/aws-stack.json if you wanted to test  — I'm going to test this on our own CI instances for a day before merging

toolmantim:
It's looking mostly stable, although I did get one `ssh: connect to host github.com port 22: Connection timed out` error out of 1100ish job runs so far. I'm not sure if this is Docker networking weirdness, or just a regular old GitHub/network timeout.

bforchhammer:
Sounds promising :-)

toolmantim:
Ok, this looks stable. 🚀 town

bforchhammer:
I've just updated our stacks as well... looks good so far ;-)

PR #110, opened 2016-08-04
Fix #109 - Remove s3 debugging output

### LINKED ISSUE #109: Adding a password-less SSH key fails
I have a put a password-less key in my secret bucket. The file is correctly downloaded from S3, but adding the key fails. Here's my build log.

```
Agent pid 2946
2016-08-04 08:01:33    1.6 KiB private_ssh_key
Downloading ssh key from s3://ss-bk-elastic-secrets-us-east-1/private_ssh_key
Enter passphrase for /dev/fd/63:
```

It hangs waiting for standard in.

ahawkins:


toolmantim:
Thanks! Good catch. I've actually seen this in production as well.

PR #112, opened 2016-08-10
More tags for EC2 instances

porty:
We have a company directive to add standardised tags to EC2 instances, including the Buildkite agent/metrics boxes. The existing `Role` and `Name` tags for the agent stack is 👌  but is missing for the metrics stack, plus we also require a few more for both boxes.

1. Tags the _metrics_ stack with appropriate Name/Role tag (follow on from https://github.com/buildkite/elastic-ci-stack-for-aws/pull/104 which only named the _agent_ boxes)
2. Adds ability for user-specified tags via `extra_tags.json`, with provided example.

## Questions

Is `extra_tags.json` overkill? The other option is to have organisations document the incantation to create the new stack (`aws cloudformation create-stack ...`) and add the tags directive there. You'd lose the Makefile goodness, but then again you could just document `make config.json build/aws-stack.json` as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,5 @@ templates/mappings.yml
 vendor/
 .bundle/
 *.bak
+extra_tags.json
 
--- a/Makefile
+++ b/Makefile
@@ -39,14 +39,18 @@ upload: build/aws-stack.json
 config.json:
 	test -s config.json || $(error Please create a config.json file)
 
-create-stack: config.json build/aws-stack.json
+extra_tags.json:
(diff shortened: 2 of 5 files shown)


toolmantim:
I think providing a standard way of doing this is AOK! Especially because the stack was birthed within your AWS environment, so I think following your lead on this makes sense. And the Name/Tag on the metrics instance is much appreciated!

Is this ready to merge in?

porty:
Yep, good to go - we're working with it already

PR #113, opened 2016-08-12
Grant sudo access to apt-get

ahawkins:
I opened this PR because projects for my team need various other software installed (make, zip, unzip, a few other things). Currently the elastic pipeline only includes docker, docker-compose, and the aws CLI. This is enough for 90% of projects but not 100%. Now I can add `apt-get install -y FOO` in to environment hook in the secrets bucket.

However the question is should we just grant sudo access to everything or only a subset.

### DIFF
--- a/packer/conf/buildkite-agent/sudoers.conf
+++ b/packer/conf/buildkite-agent/sudoers.conf
@@ -1,1 +1,1 @@
-buildkite-agent ALL=NOPASSWD: /usr/bin/fix-buildkite-agent-builds-permissions
+buildkite-agent ALL=NOPASSWD: /usr/bin/fix-buildkite-agent-builds-permissions, /usr/bin/apt-get


dgoodlad:
IMHO this breaks one of the basic assumptions of the stack: that a build will not affect the host instance. I'd suggest that, if you're using this stack, you might be better off doing one of two things:

1. use that other software _inside docker containers_
2. build a custom AMI based on the public one, and install whatever software you need there.

toolmantim:
Hey @ahawkins! Yeah, we can't really put this change in there I'm afraid.

You're welcome to customize the cloudformation stack to run your own commands on cloud-init, or alternatively build your AMI like Dave suggested. And if there's any generic improvements we could make to the build process (ala #112) that'd be fine!

But we can't just whitelist `apt-get` I'm afraid.

Is there any reason why Docker isn't a good case for this? I do [something similar in the `Makefile` for the stack](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/Makefile#L65) where I use a container just to run a npm bin command on a file, and it's quite snappy and completely self-contained.

ahawkins:
> You're welcome to customize the cloudformation stack to run your own commands on cloud-init

Would you support a PR that allowed some sort of custom user data passed in? I'd like to avoid building an AMI if possible. I prefer this project because I can skip building an AMI which takes an age.

lox:
@ahawkins could you explain a bit more about why you can't just do the customizations in a docker container? That's the approach that the elastic-ci stack was designed for. If anything, I'd prefer to default to running things in a container, modifying the host is incredibly non-deterministic.

toolmantim:
Sorry that link I posted earlier broke because master changed. It was meant to point to: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/627d5fd53c525500f3be5ce75e93fbb29abc3a5a/Makefile#L68-L69

ahawkins:
We do all of our build and test runs with Docker containers. We still
need a few host level programs to coordinate deploys (like zip to create
Elasticbeanstalk source bundles) and other external interactions.


On Fri, Aug 12, 2016, at 12:01 PM, Lachlan Donald wrote:
> @ahawkins[1] could you explain a bit more about why you can't just do
> the customizations in a docker container? That's the approach that the
> elastic-ci stack was designed for. If anything, I'd prefer to default
> to running things in a container, modifying the host is incredibly non-
> deterministic.
> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].
>


Links:

  1. https://github.com/ahawkins
  2. https://github.com/buildkite/elastic-ci-stack-for-aws/pull/113#issuecomment-239371370

toolmantim:
Actually, AWS does use zip a lot.

I had to do it here for EB deployments via this stack (it took a lot of messing around):
https://github.com/buildkite/polyglot-co-demo-backend/blob/39036297cfb26e462c8cb71901d1c5e886182436/Makefile#L16-L21
https://github.com/buildkite/polyglot-co-demo-backend/blob/39036297cfb26e462c8cb71901d1c5e886182436/.buildkite/steps/deploy.bash#L7

Adding `zip` the base image isn't necessarily a bad idea given some AWS APIs need it.

PR #114, opened 2016-08-12
Basic support for paravirtualized images

Vektah:
This is by no means complete, but I wanted to see if anyone else is interested using the older paravirtualized instance types?

spot pricing for the c4.large instances is going crazy:
image

These huge price spikes cause instance terminations regardless of lifecycled which look like this in a job:
```
$ docker-compose -f .buildkite/docker-compose.yml -p buildkite135479143c984704a4046342d74118a4 kill
ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?
 
If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.
$ docker-compose -f .buildkite/docker-compose.yml -p buildkite135479143c984704a4046342d74118a4 rm --force --all -v
ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?
 
If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.
$ docker-compose -f .buildkite/docker-compose.yml -p buildkite135479143c984704a4046342d74118a4 down
ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?
 
If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.
```

Older instance types seem to have much more capacity, and lower more stable prices:
image

These are the bare minimum changes required to use the PV instances like m1.large, if there is interest perhaps the mappings should support these instance types automatically?

### DIFF
--- a/packer/buildkite-ami.json
+++ b/packer/buildkite-ami.json
@@ -3,9 +3,9 @@
     {
       "type": "amazon-ebs",
       "region": "us-east-1",
-      "source_ami": "ami-6869aa05",
+      "source_ami": "ami-2a69aa47",
       "region": "us-east-1",
-      "instance_type": "c4.large",
+      "instance_type": "m1.large",
       "ssh_username": "ec2-user",
       "ami_name": "buildkite-stack-{{isotime | clean_ami_name}}",
       "ami_description": "Buildkite CloudFormation Stack base image (Amazon Linux, buildkite-agent, docker, docker-compose, docker-gc, jq, lifecycled)",
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -311,7 +311,7 @@ Resources:
         $(ImageId)
       ]
       BlockDeviceMappings:
-        - DeviceName: /dev/xvda
+        - DeviceName: /dev/sda1
           Ebs: { VolumeSize: $(RootVolumeSize), VolumeType: gp2, Iops: $(RootVolumeIops) }
       UserData: !Base64 |
         #!/bin/bash -xv


toolmantim:
Interesting…

lox:
@Vektah I haven't actually added support for spot pricing notifications to lifecycled yet btw.

lox:
https://github.com/lox/lifecycled/issues/3

lox:
So, is the aim just to do this for the packer builder @Vektah? Or agent instances?

Vektah:
> @Vektah I haven't actually added support for spot pricing notifications to lifecycled yet btw.

I doubt we would be able to delay them even if we were listening to them? Once your spot price has been exceeded I imagine AWS forcefully kill that instance pretty quickly.

> So, is the aim just to do this for the packer builder @Vektah? Or agent instances?

I dont think the packer builder needs to change as long as whatever instance its built on support both PV and HVM.

Ideally there would be pre built paravirtualized AMIs pushed up setting an instance type that is PV only would use one of a PV image. I think HVM is preferred where ever its available.

toolmantim:
So for spot instances, we're going add some kind of signal support to say "I'm shutting down, so cancel this job but feel free to retry" and a pipeline step can say `retryable` or something similar, and it'll just reschedule it to run again.

As for PV support, it'd be great to add support for this with some kind of mapping. Assuming it's not massive overhead?

lox:
Well, it's certainly nicer to get the 2 minutes extra lead time that termination notifications provide. Would allow for a graceful shutdown in most cases.

lox:
My vote would be 👎 for adding support for PV instance types. I think they are a dying breed and not worth the complexity it would introduce to the stack .

PR #115, opened 2016-08-12
Install zip/unzip

ahawkins:
These tools are used when integrating with AWS services. These should be
included given the AWS CLI is installed. This is a follow up from #113.


toolmantim:
Yep, I think this makes sense given AWS uses them.

PR #118, opened 2016-08-15
Remove provisioned IOPS from the template

### LINKED ISSUE #116: EBS gp2 doesn't support provisioned IOPS
The template tried to provision IOPS - which is doing nothing, since it specifies general purpose SSD (gp2) which doesn't support provisioned IOPS. Did you intend for a "volume type" argument to go with this?

holmesjr:
Provisioned IOPS only applies to io1 volumes, not gp2. Fixes #116

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -33,7 +33,6 @@ Metadata:
         - ArtifactsBucket
         - AuthorizedUsersUrl
         - RootVolumeSize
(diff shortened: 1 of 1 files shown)


toolmantim:
Yeah I don't think PIOPS is necessary most of the time either…

toolmantim:
And one less config to have to figure out

toolmantim:
Thanks for the PR @holmesjr!

PR #120, opened 2016-08-18
Add optional BootstrapScriptUrl parameter

ahawkins:
This provides a similar solution like UserData. Ideally this script
would be used to install supporting packages or run other one-time
configuration (such as setting AWS CLI profiles).

Users can provide this to customize the instances in some way _without_
having to worry about building a new AMI.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -32,6 +32,7 @@ Metadata:
         - SecretsBucket
         - ArtifactsBucket
         - AuthorizedUsersUrl
+        - BootstrapScriptUrl
         - RootVolumeSize
         - RootVolumeIops
 
@@ -97,6 +98,11 @@ Parameters:
 
   AuthorizedUsersUrl:
     Description: Optional - S3 url to periodically download ssh authorized_keys from
+    Type: String
+    Default: ""
+
+  BootstrapScriptUrl:
+    Description: Optional - S3 url to run on each instance during boot
     Type: String
     Default: ""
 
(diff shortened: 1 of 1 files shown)


toolmantim:
This is something I'm also interested in, rather than building an AMI for us internally as well. For example, I want to install DataDog onto the instances. I'm keen to get some other people's opinion on it too though. @lox @joffotron @sj26?

If we don't do this, we probably need to make it easier for people to run their custom AMI + CF stack via pipelines?

ahawkins:
> If we don't do this, we probably need to make it easier for people to run their custom AMI + CF stack via pipelines?

Yes definitely. That is a significantly higher barrier to entry then `curl | bash`. This issue is my final blocker before adopting this project at Saltside.

joffotron:
People that have more complex requirements can always still build their own AMIs. The setup we had was to add an 'ansible-local' step to the AMI Packer file, and then do everything in Ansible from there on in. 

The nice part of that approach is that you know everything is set up just how you want it on the box before it even boots / starts accepting agent jobs, etc etc. 

This does need more DevOpsKnowHow™, so the option of just running a script from S3 might cover a lot of cases.

toolmantim:
Yeah, I feel like this will give people enough rope to get something working, and if it causes instability with their instances then the pain of building their own AMI will be apparent (and worthwhile)

ahawkins:
> Yeah, I feel like this will give people enough rope to get something working, and if it causes instability with their instances then the pain of building their own AMI will be apparent (and worthwhile)

However if you build your own AMI, you also start maintaining your own fork. I view this PR as a means to avoid building AMI's and maintaining my fork (which my customizations) to apply on top of upstream base.

toolmantim:
Yeah, keeping people on the latest version of the stack is a good thing

verveguy-slave:
allowing a "bootstrap" ansible playbook to be specified via URL (from S3 or other..) during cloudformation stack creation would be very helpful.

toolmantim:
Ok, let's go with this and if it turns out a terrible headache for people (or us) we can reevaluate.

PR #123, opened 2016-08-21
Don't stop all other Docker containers

### LINKED ISSUE #121: Cannot use docker and multiple buildkite-agents
Hi, 

I have discovered that it is currently impossible to reliably run multiple buildkite-agents with docker on the same machine. If a couple of tasks start up shortly after each other, the following lines in the buildkite "environment" hook will lead to running docker-containers (from earlier tasks) being killed... In the build output this leads to failed tasks with exit status 137.

    if [[ $(docker ps -q | wc -l) -gt 0 ]] ; then
      echo "~~~ Stopping existing :docker: containers"
      docker stop $(docker ps -q)
    fi

I guess solutions would be to either rely on build-scripts to clean up after themselves (i.e. no cleanup in the environment hook at all), or to scope the killing to the specific buildkite-agent somehow...

toolmantim:
If you run multiple agents, or you run other daemons on your instance, just stopping all Docker containers is bad. This assumes that people will stop their Docker containers, and that Buildkite's default Docker gear behaves properly.

Closes #121

### DIFF
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -58,11 +58,6 @@ fi
 echo "~~~ Waiting for :docker:"
 timeout 30 docker ps
 
-if [[ $(docker ps -q | wc -l) -gt 0 ]] ; then
-  echo "~~~ Stopping existing :docker: containers"
-  docker stop $(docker ps -q)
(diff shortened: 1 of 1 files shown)


bforchhammer:
Awesome, thanks! :+1:

PR #125, opened 2016-08-21
Trigger rolling updates for key params

### LINKED ISSUE #84: Updating stack parameters doesn't trigger a rolling update
If you update the buildkite agent queue name, API key, agents per instance or release stream (from stable to beta) you need to manually set the ASG desired size to 0 and force a scale down for it to apply changes, because it's all done in cloud-init. And some people might try just deleting the stack and recreating.

It'd be preferable if you could just update the stack parameters and it's handled for you. This could mean performing a rolling update somehow?

toolmantim:
Updating the agent queue name, and a bunch of other keys, doesn't trigger an update of instances. This outputs those params into UserData so an update is triggered.

> Update requires: Some interruptions for Amazon EBS-backed instances.
> For EBS-backed instances, changing the UserData stops and then starts the instance; however, Amazon EC2 doesn't automatically run the updated UserData. To update configurations on your instance, use the cfn-hup helper script.
> Update requires: Replacement for instance store-backed instances.

Closes #84 

This doesn't fix updating the API key, that'll have to be done in the metrics stack.


toolmantim:
@exidy does this look correct?

exidy:
LGTM.

PR #127, opened 2016-08-22
Clean up docker login

### LINKED ISSUE #111: AWS ECR login creds are clobbered by pre-command
At the moment there's no way to login to ECR to use the `docker-compose` build/push support.

If you add an `env` hook that performs:

```
$(aws ecr get-login --region us-west-2)
```

then because of [our pre-command hook](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/76990cb0ca0422b0d3c61e6ca0bdf335243ea56c/packer/conf/buildkite-agent/hooks/pre-command#L5) the creds get removed before any commands are run.

The pre-command hook basically undoes the env hook.

I propose we:

* Don't `rm` the docker config files (but if we must keep it, at least move it into the conditional)
* Add built-in support for `AWS_ECR_LOGIN=<region>` which performs the login for you

We could also add support for logins in https://github.com/buildkite-plugins/docker-compose-buildkite-plugin itself?

toolmantim:
This removes `DOCKER_HUB_*` configs in favour of generic `DOCKER_LOGIN_*` environment variables, and adds support for AWS ECR login.

This also removes the `rm` of the Docker login creds, a file that is readable by all builds. This means that if one step performs a docker login, subsequent steps can read that file. Is this a problem? I'm not sure what the answer is for a world where more than one agent runs on the machine at once. The current implementation isn't much more secure at the moment in this situation anyway.

- [x] First cut for review
- [x] Decide on whether it's okay to leave Docker login creds around between job runs
- [x] Add back support for `DOCKER_HUB_USER` and `DOCKER_HUB_PASSWORD` with a deprecation warning

Closes #111

### DIFF
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -21,7 +21,7 @@ s3_download() {
   env -i aws s3 cp --quiet ${aws_s3_args[@]} "$1" /dev/stdout
 }
 
-echo "~~~ :house_with_garden: Setting up the environment"
+echo "~~~ Setting up the environment"
 
 eval "$(cat ~/cfn-env)"
 eval "$(ssh-agent -s)"
@@ -55,7 +55,7 @@ if [[ -n "${BUILDKITE_SECRETS_BUCKET:-}" ]] ; then
   fi
 fi
 
-echo "~~~ Waiting for :docker:"
+echo "~~~ Waiting for Docker"
 timeout 30 docker ps
 
 unset BUILDKITE_SECRETS_BUCKET
--- a/packer/conf/buildkite-agent/hooks/pre-command
+++ b/packer/conf/buildkite-agent/hooks/pre-command
@@ -4,11 +4,13 @@ set -eu -o pipefail
 
 : "${AWS_ECR_LOGIN:-}"
 : "${AWS_ECR_LOGIN_REGISTRY_IDS:-}"
-: "${DOCKER_HUB_USER:-}"
-: "${DOCKER_HUB_PASSWORD:-}"
 : "${DOCKER_LOGIN_USER:-}"
 : "${DOCKER_LOGIN_PASSWORD:-}"
 : "${DOCKER_LOGIN_SERVER:-}"
+
+# Legacy envs
+: "${DOCKER_HUB_USER:-}"
+: "${DOCKER_HUB_PASSWORD:-}"
 
 # AWS ECR login support
 
--- a/packer/conf/buildkite-agent/hooks/pre-command
+++ b/packer/conf/buildkite-agent/hooks/pre-command
@@ -3,7 +3,6 @@
 set -eu -o pipefail
 
 [[ -f ~/.docker/config.json ]] && rm ~/.docker/config.json
-[[ -f ~/.dockercfg ]] && rm ~/.dockercfg
 
 if [[ -n "${DOCKER_HUB_AUTH:-}" ]] ; then
   DOCKER_HUB_USER=$(jq -r '."https://index.docker.io/v1/".auth' <<< "$DOCKER_HUB_AUTH" | base64 --decode | cut -d: -f1)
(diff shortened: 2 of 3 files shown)


toolmantim:
@joffotron I just stumbled upon another gotcha for multiple-agents-per-instance… leaked `docker login` credentials.

packer/conf/buildkite-agent/hooks/environment
```diff
@@ -21,7 +21,7 @@ s3_download() {
   env -i aws s3 cp --quiet ${aws_s3_args[@]} "$1" /dev/stdout
 }
 
-echo "~~~ :house_with_garden: Setting up the environment"
+echo "~~~ Setting up the environment"
```
bforchhammer: No emoticons anymore? :crying_cat_face:

packer/conf/buildkite-agent/hooks/pre-command
```diff
@@ -2,24 +2,32 @@
...
+# Generic docker login support
+
+: "${DOCKER_LOGIN_USER:-}"
+: "${DOCKER_LOGIN_PASSWORD:-}"
+: "${DOCKER_LOGIN_SERVER:-}"
+
+if [[ -n "${DOCKER_LOGIN_USER:-}" && -n "${DOCKER_LOGIN_PASSWORD:-}" ]] ; then
+  echo "~~~ Authenticating with Docker server"
+
+  docker login --username="${DOCKER_LOGIN_USER}" --password="${DOCKER_LOGIN_PASSWORD}" "${DOCKER_LOGIN_SERVER}"
```
bforchhammer: :+1:

bforchhammer:
lgtm :rocket:

packer/conf/buildkite-agent/hooks/environment
toolmantim: Maybe I should put them back. I've seen people removing them in their own forks, to give their own emojis more prominence! But perhaps I need to apply a CSS b&w filter to emojis that come from `~~~` lines (which are already subdued)

bforchhammer: Hehe, I don't really mind either way, I was just wondering... ;-)

joffotron:
Looks Good!
Sadly, I don't have access to the multi-agent install any more, but we were using ECR, so the above changes are :+1:

toolmantim:
I think to merge this we should add back in support for `DOCKER_HUB_USER` and `DOCKER_HUB_PASSWORD` with a deprecation warning.

bjeanes:
How do you specify the region? I am getting the following in the ECR login step:

> You must specify a region. You can also configure your region by running "aws configure".

bjeanes:
Looking at the code, it doesn't look like `AWS_ECR_LOGIN=<region>` would work, as you described in #111. Is the AWS CI stack supposed to set a default region with `aws configure`?

toolmantim:
@bjeanes sorry you've been having problems getting this going.

So [the pre-command hook](http://docs.aws.amazon.com/cli/latest/reference/ecr/get-login.html) simply calls [aws ecr get-login](http://docs.aws.amazon.com/cli/latest/reference/ecr/get-login.html) — the syntax changed between this issue being discussed, and it shipping, hence the lack of `AWS_ECR_LOGIN=region` (because that used the be the old cli syntax, `aws ecr login <region>`).

A quick fix is to set the `AWS_DEFAULT_REGION` environment variable (which is standard for all AWS cli stuff).

I thought it would have inferred the region? If not, we could add `AWS_ECR_REGION` which adds `--region`, or just document `AWS_DEFAULT_REGION` (but it'll apply to all `aws` cli commands), or we make `AWS_ECR_LOGIN` a bit smarter and allow the `AWS_ECR_LOGIN=<region>` syntax. What do you think?

bjeanes:
I'll have to try it to have an opinion worth any weight, but in general I prefer explicitness, so would be inlined to have `AWS_ECR_REGION`/`AWS_ECR_LOGIN=<region>` instead of (or in addition to) `AWS_DEFAULT_REGION`. It definitely doesn't infer, which is also what I expected, but `AWS_DEFAULT_REGION` will work in my case since I am not running any other AWS commands against different regions at the moment. I'll give it a try and report back, I suppose.

PR #128, opened 2016-08-24
Follow bootstrap URL redirects

ahawkins:
This works nicely with S3 urls since they will give redirects if the region is not included. This change is not AWS specific. There is not a reason to skip redirects on these urls.


toolmantim:
Thanks for the PR! 👍👍👏👌

Issue #130, opened 2016-08-30
Rolling update fails with "Failed to receive 5 resource signal(s)"


toolmantim:
When performing a stack update with running instances it's very common for it to fail, making for a long 50 minutes. #126 was meant to help this, but it appears not really to help at all, but only to make the wait longer.

The stack event log:

```
21:07:41 UTC+10 UPDATE_ROLLBACK_COMPLETE  AWS::CloudFormation::Stack  elastic-runners 
21:07:41 UTC+10 UPDATE_COMPLETE AWS::CloudFormation::Stack  MetricsStack  
21:07:30 UTC+10 DELETE_COMPLETE AWS::AutoScaling::LaunchConfiguration AgentLaunchConfiguration  
21:07:29 UTC+10 UPDATE_IN_PROGRESS  AWS::CloudFormation::Stack  MetricsStack  
21:07:29 UTC+10 DELETE_IN_PROGRESS  AWS::AutoScaling::LaunchConfiguration AgentLaunchConfiguration  
21:07:26 UTC+10 UPDATE_ROLLBACK_COMPLETE_CLEANUP_IN_PROGRESS  AWS::CloudFormation::Stack  elastic-runners 
21:07:25 UTC+10 UPDATE_COMPLETE AWS::CloudFormation::Stack  MetricsStack  
21:07:02 UTC+10 UPDATE_COMPLETE AWS::AutoScaling::AutoScalingGroup  AgentAutoScaleGroup 
21:06:58 UTC+10 UPDATE_IN_PROGRESS  AWS::AutoScaling::AutoScalingGroup  AgentAutoScaleGroup Received SUCCESS signal with UniqueId i-2e3ce11f
21:06:58 UTC+10 UPDATE_IN_PROGRESS  AWS::AutoScaling::AutoScalingGroup  AgentAutoScaleGroup Received SUCCESS signal with UniqueId i-293ce118
21:06:58 UTC+10 UPDATE_IN_PROGRESS  AWS::AutoScaling::AutoScalingGroup  AgentAutoScaleGroup Received SUCCESS signal with UniqueId i-3502df04
21:06:58 UTC+10 UPDATE_IN_PROGRESS  AWS::AutoScaling::AutoScalingGroup  AgentAutoScaleGroup Received SUCCESS signal with UniqueId i-0b77820a
[...]

PR #131, opened 2016-08-30
Send cfn-signal/init logs to CloudWatch

toolmantim:
To make it easier to debug autoscaling/cfn signals (for example, for #130), this adds the cfn logs to CloudWatch


sj26:
Neat!

PR #132, opened 2016-08-31
Updated CreationPolicy/ResourceSignal count to 1

keithpitt:
> The number of success signals AWS CloudFormation must receive before it sets the resource status as CREATE_COMPLETE. If the resource receives a failure signal or doesn't receive the specified number of signals before the timeout period expires, the resource creation fails and AWS CloudFormation rolls the stack back.

After re-reading the docs, we believe the `Count` parameter here is per-instance. The ASG is marked as complete immediately, where it should be waiting for 1 signal from each instance.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -523,7 +523,7 @@ Resources:
     CreationPolicy:
       ResourceSignal:
         Timeout: PT5M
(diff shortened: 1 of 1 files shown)


keithpitt:
```

toolmantim:
Probs don't need this after all!

PR #133, opened 2016-08-31
Start agents after boostrapping

ahawkins:
This fixes a race condition where the agents may accept jobs before bootstrapping completes.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -423,14 +423,6 @@ Resources:
 
                 service awslogs restart
 
-                # Setup and start services
-
-                for i in \$(seq 1 $(AgentsPerInstance)); do
-                  cp /etc/buildkite-agent/init.d.tmpl /etc/init.d/buildkite-agent-\${i}
-                  service buildkite-agent-\${i} start
-                  chkconfig --add buildkite-agent-\${i}
-                done
-
             05-fetch-authorized-users:
               test: test -n "$(AuthorizedUsersUrl)"
               command: |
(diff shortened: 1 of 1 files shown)


shuber:
@ahawkins What happens when these agents accept jobs before bootstrapping? Just curious what kind of errors or log output you're seeing in these cases.

ahawkins:
Depends on what you do in the bootstrap script. I install vagrant in my
bootstrap script. My env hook invokes "vagrant --version". Thus
depending on when the agent accepts the job that build will pass/fail
depending if the bootstrap completed.


On Wed, Aug 31, 2016, at 11:33 PM, Sean Huber wrote:
> @ahawkins[1] What happens when these agents accept jobs before
> bootstrapping? Just curious what kind of errors or log output you're
> seeing in these cases.
> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].
>


Links:

  1. https://github.com/ahawkins
  2. https://github.com/buildkite/elastic-ci-stack-for-aws/pull/133#issuecomment-243848989

shuber:
Nice I think I've seen similar behavior before where some agents can't parse our secrets from the S3 bucket when they first boot up. It only happens occasionally and I suspect this could be what's causing it.

dgoodlad:
👍

ahawkins:
Freshly rebased for your merging pleasure :bow:

toolmantim:
Danke @ahawkins!

toolmantim:
@shuber I've seen that very occasionally too, it should be related to this though…

toolmantim:
@shuber *shouldn't be related

PR #134, opened 2016-09-01
[WIP] Per-instance SQS autoscaling message queues

toolmantim:
One solution for #130 is to switch to per-instance SQS queues, still using lifecycled, and each one subscribes to the SNS topic with the autoscaling events.

Instances will create their SQS queue on init, and then delete the SQS queue once lifecycled has completed.

- [x] Initial cut of demo code
- [ ] Finish code & roles & testing

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -246,8 +246,8 @@ Resources:
             Resource: "*"
           - Effect: Allow
             Action:
-              - sqs:*
-            Resource: $(AgentLifecycleQueue[Arn])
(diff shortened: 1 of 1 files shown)


lox:
Hmmmm, I wonder if there is some other way to do this without the complexity of creating and managing all these extra queues. 

If we cut the timeouts down massively in https://github.com/lox/lifecycled/pull/4 and write some load tests in there to see what the timeframes look like, will that address the issue?

toolmantim:
This isn't needed due to #135

PR #135, opened 2016-09-01
Replace lifecycled without SNS/SQS

sj26:
We've had some trouble with SQS lifecycle handling, so we can try something a little less sophisticated and _POLL FOR GREAT JUSTICE_. Maybe. Haha.

It definitely works:

screenshot 2016-09-01 18 09 23


templates/buildkite-elastic.yml
```diff
@@ -442,22 +466,28 @@ Resources:
             06-run-bootstrap-script:
               test: test -n "$(BootstrapScriptUrl)"
               command: |
-                #!/bin/bash -euo pipefail
```
sj26: (This option caused an error for me)

toolmantim:
Fixes #130

toolmantim:
Stack JSON URL: `https://s3.amazonaws.com/buildkite-aws-stack/lifecycled-the-second/aws-stack.json`

lox:
:(

toolmantim:
Sorry @lox! Turns out you can do a `CONTINUE` without a lifecycle action token, and just the instance id. The docs/guide/examples don't make that obvious at all.

Which means you can just poll your own state to see the state change.

templates/buildkite-elastic.yml
```diff
@@ -428,7 +428,7 @@ Resources:
                                        --query AutoScalingInstances[].LifecycleState --output text)"
                   logger -t lifecycled "Lifecycle state for \$INSTANCE_ID: \$LIFECYCLE_STATE"
                   if [[ "\$LIFECYCLE_STATE" == "Terminating:Wait" ]]; then
-                    logger -t lifecycled "Time to shut it all down!"
```
sj26: Oh sorry, this would have broken. Woops!

sj26: Oh no, it did work in the example. Perhaps it's too enthusiastic, though.

toolmantim:
Rolling updates work AOK, scaling up and down works AOK. I think this is good to 🚀

toolmantim:
This seems to be working well 👍

PR #141, opened 2016-09-13
use correct parameter name for ECR login

AdamNowotny:
When using AWS_ECR_LOGIN_REGISTRY_IDS variable I'm getting an error
```
Unknown options: --registry_ids,[xxxxxxxxxxxxx]
```
The parameter used by aws-cli is "registry-ids", not "registry_ids"


toolmantim:
Thanks Adam!

PR #143, opened 2016-09-14
Prevent chmod collisions in multi-agent per instance stacks

sj26:
If we chmod the whole builds directory at the beginning of every build then we might chmod files which are currently in use by another agent's build.

The builds path has a first component which is the agent name. We can safely chmod recursively inside there. But we don't know this path, and don't expose it anywhere. And it's not as simple as `builds-path/agent-name` because the agent name is sanitized, but how it is sanitized depends on the agent version.

So we can grab the checkout path and work backwards. It's kinda hacky, but it works.

Like @tim suggested, it might be nice to extract the munging from the fix script into the environment hook and just pass in a prepared agent directory name and just make sure it has no slashes and isn't a traversal component in the script.

### DIFF
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -61,4 +61,4 @@ timeout 30 docker ps
 unset BUILDKITE_SECRETS_BUCKET
 unset BUILDKITE_SECRETS_KEY
 
-sudo /usr/bin/fix-buildkite-agent-builds-permissions
+sudo /usr/bin/fix-buildkite-agent-builds-permissions "${BUILDKITE_BUILD_CHECKOUT_PATH}"
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -23,10 +23,17 @@ s3_download() {
 
 echo "~~~ Setting up the environment"
 
+echo "Sourcing CloudFormation environment..."
 eval "$(cat ~/cfn-env)"
+echo
+
+echo "Starting an SSH Agent..."
 eval "$(ssh-agent -s)"
+echo
 
 if [[ -n "${BUILDKITE_SECRETS_BUCKET:-}" ]] ; then
+  echo "Downloading secrets..."
+  echo
 
   # Allow environment vars set in Buildkite to override paths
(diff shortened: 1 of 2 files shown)


toolmantim:
Thanks! I'll give this a test once it's built.

shuber:
Will this fix issues like this?
```
git clean -fxdq
warning: failed to remove tmp/cache/assets/sprockets/v3.0/xJ/xJp0BBSfoEY_xAaV-vNHd_oy26vDwZJQIIawkxXEoqk.cache
warning: failed to remove tmp/cache/assets/sprockets/v3.0/FH/FHdUqc7Xi1zv_N7zZOP2by38NrLESpyeiKuFuCLkdD0.cache
```

We're getting these occasionally and we only have one agent running per instance.

sj26:
screenshot 2016-09-15 11 25 02

sj26:
@shuber oh dear. If you're having that problem with a single agent then I don't think so. Are you running docker with the build directory mapped in? If it writes files to the volume they're probably owned by root, which is what this sudo script is meant to fix up at the beginning of the next run. This is an unfortunate side effect of using Docker. We're also looking into a better fix using [user namespaces](https://integratedcode.us/2015/10/13/user-namespaces-have-arrived-in-docker/) to make sure files on host filesystem end up owned by buildkite-agent.

shuber:
@sj26 Correct we're running docker and mounting volumes for stuff like `node_modules` right in the build directory itself. Our environment hooks run the sudo script that you're describing but occasionally we still get those errors. We run like 80 agents per build and more often then not it wont just be one agent that has that issue. Instead like 20-40 of them will fail within 10 seconds with that error, all in the same build.

sj26:
@shuber definitely only one agent per instance? That does sound like the issue this PR is intended to fix, but with multiple agents per instance.

keithpitt:
Anything I can do to help get this over the line? :)

shuber:
@sj26 yea we use
```json
{
    "ParameterKey": "AgentsPerInstance",
    "ParameterValue": "1"
  }
```

sj26:
@tim did you end up giving it a test?

sj26:
@shuber I don't think it's related to this change then I'm afraid. I'll reach out separately and maybe we can figure it out!

Issue #81, opened 2016-07-06
Reducing wait time for cold caches with large apps


toolmantim:
Currently the best way to handle a large app that takes ages to build is to have two separate instances of the stack: builders and runners, and https://github.com/buildkite-plugins/docker-compose-buildkite-plugin to have a `build` step that runs before the steps.

The builder doesn't autoscale, sits there with a warm docker cache, and pushes to a registry. The test steps run on the `runners` queue, and pulls from docker hub without doing any git clone at all.

The pipeline looks like this:

``` yaml
steps:

  - name: ":docker: :package:"
    plugins:
      docker-compose:
        build: app
        image-repository: index.docker.io/myorg/myapp
    agents:
      queue: elastic-builders

  - wait

  - name: ":spec: %n"
    command: ".buildkite/steps/knapsack"
    plugins:
      docker-compose:
        run: app
    agents:
      queue: elastic-runners
    parallelism: 75
```

A simpler setup would be to use something like http://blog.runnable.com/post/145362675491/distributing-docker-cache-across-hosts to make the cache mostly invisible, and require no separate queue or pipeline config changes.

lox:
Distributing cache across hosts would be great. For that to work, you'd need a list of images to save and then some code in the `environment` and `post-command` hooks. Thoughts on how you'd specify which images to cache?

toolmantim:
So @tonistiigi just released https://github.com/tonistiigi/buildcache (https://github.com/docker/docker/issues/20316#issuecomment-240531816) which we could potentially use

mtibben:
Codeship has a blog post explaining how they tackled this problem https://blog.codeship.com/building-a-remote-caching-system/

lox:
This will be solved by https://github.com/buildkite/elastic-ci-stack-for-aws/pull/234.

It could also be solved by using the new docker 1.13 `--cache-from` stuff, which needs some investigation.

jdotjdot:
I'm seeing some recursion here between this PR and #234.  Is the shared layer cache support being handled?

lox:
@jdotjdot you should be able to use the new `--cache-from` stuff in docker-compose 1.14, which is on the latest stack version. I'm going to close this for now, do you have a specific issue that you are currently seeing?

jdotjdot:
We don't use docker-compose is the issue. Is there another way?

lox:
Ah, right, you don't use docker at all? Or just not docker-compose?

[...]

PR #145, opened 2016-09-19
Implement shared docker layer cache

Vektah:
@porty and I have been hacking on a fix for https://github.com/buildkite/elastic-ci-stack-for-aws/issues/81

It adds an s3 bucket that contains a tarball for each pipeline that can be `docker load`'ed. Its very much batteries included, all a project should do is nominate a step to save all built layers by setting `DOCKER_CACHE_IMAGES_PUSH=true`. Ideally this is done on master.

pre-commit:
- every step will fetch s3:://cache-bucket/$pipeline.images, which is a list of layer hashes contained in s3://cache-bucket/$pipeline.tar.gz
- if any of the layer hashes are missing, fetch the tarball and `docker load`

post-commit in one step, probably on master only
- work out all the layers used by all docker containers tagged with the buildkite id (should be everything built in this step)
- compare that with the contents of s3://cache-bucket/$pipeline.images
- if the image list is different `docker save` all of the images into s3://cache-bucket/$pipeline.tar.gz and update s3://cache-bucket/$pipeline.images

On both sides if the machine already has fetched all of the required layers both the save and load steps skip, so the fast path is really fast. The slow path, where the cache needs to be fetched is quite a bit faster, fetching from s3 is pretty speedy compared to pulling and building.

_Work in progress_: Still need to do a bunch of testing.

### DIFF
--- a/packer/conf/buildkite-agent/hooks/post-command
+++ b/packer/conf/buildkite-agent/hooks/post-command
@@ -1,3 +1,27 @@
 #!/bin/bash
 
+set -eux -o pipefail
+
 ssh-agent -k
+
+if [[ -n "${BUILDKITE_DOCKER_COMPOSE_CONTAINER:-}" && -n ${DOCKER_CACHE_IMAGES_PUSH:-} && BUILDKITE_COMMAND_EXIT_STATUS -eq 0 ]] ; then
+    echo "~~~ Pushing cached docker images"
+    (
+        # clear any aws env vars so the machine profile is used
+        unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_REGION
+
+        slug=/tmp/docker-cache/$BUILDKITE_PIPELINE_SLUG.tar.gz
+        images_file=s3://$BUILDKITE_IMAGE_CACHE_BUCKET/$BUILDKITE_PIPELINE_SLUG.images
+        images=$(echo $(docker images -a | grep $(echo $BUILDKITE_JOB_ID | sed 's/-//g') | awk '{print $1}' | xargs -n 1 docker history -q | grep -v '<missing>'))
+
+        if [[ -n $images ]] && ( ! aws s3 ls $images_file || [[ "$images" != $(aws s3 cp $images_file -) ]]) ; then
+            rm -rf /tmp/docker-cache
+            mkdir -p /tmp/docker-cache
+
+            docker save $images | gzip -c > $slug
+
+            aws s3 cp $slug s3://$BUILDKITE_IMAGE_CACHE_BUCKET/$BUILDKITE_PIPELINE_SLUG.tar.gz
+            echo "$images" | aws s3 cp - s3://$BUILDKITE_IMAGE_CACHE_BUCKET/$BUILDKITE_PIPELINE_SLUG.images
+        fi
+    )
+fi
--- a/packer/conf/buildkite-agent/hooks/pre-command
+++ b/packer/conf/buildkite-agent/hooks/pre-command
@@ -41,3 +41,21 @@ if [[ -n "${DOCKER_LOGIN_USER:-}" && -n "${DOCKER_LOGIN_PASSWORD:-}" ]] ; then
   unset DOCKER_LOGIN_PASSWORD
   unset DOCKER_LOGIN_SERVER
 fi
+
+
+# Docker compose layer caching
+
+if [[ -n "${BUILDKITE_DOCKER_COMPOSE_CONTAINER:-}" ]] ; then
+    echo "~~~ Fetching cached docker images"
+
+    (
+        # unset any aws keys so we use the machine policy
+        unset AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_REGION
+
+        # see if we are missing any of the images locally, and load them if we are
+        images_file=s3://$BUILDKITE_IMAGE_CACHE_BUCKET/$BUILDKITE_PIPELINE_SLUG.images
+        if aws s3 ls $images_file && ! docker inspect $(aws s3 cp $images_file -) > /dev/null ; then
+            aws s3 cp s3://$BUILDKITE_IMAGE_CACHE_BUCKET/$BUILDKITE_PIPELINE_SLUG.tar.gz - | gunzip -c | docker load
+        fi
+    )
+fi
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -250,6 +250,27 @@ Resources:
       Roles:
         - $(IAMRole)
 
+  ImageCacheBucket:
+      Type: AWS::S3::Bucket
+
+  ImageCacheBucketPolicy:
+    Type: AWS::IAM::Policy
+    Properties:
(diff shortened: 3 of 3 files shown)


lox:
Awesome! I'll dig into this tomorrow. I dimly recall that having a CFN managed s3 bucket results in stack destroy failing if the bucket has contents, have you tried?

Vektah:
The docs say it should delete by default, but I havent tried science yet.

lox:
Perhaps it was just if you made changes to the bucket, or perhaps that was fixed in old times.

porty:
The docs also say:

> Note
> For Amazon S3 buckets, you must delete all objects in the bucket for deletion to succeed.

Which seems to indicate manual work required 😢

Vektah:
Ah bummer, I guess we will have to hoist it out of the cfn template like the other buckets.

toolmantim:
Wow, thanks @Vektah & @porty. Would love to see this in! 🎱

lox:
Yup, that's basically why I didn't make the secrets bucket stack managed originally. Very annoying.

lox:
Are you guys running this on builds yet? Does it work out quicker?

Vektah:
The speedup we've seen so far is 25-50% depending on how heavily docker is used. I think there are more optimisations to do once there is a cache like npm / app deps in. I think the big win will be shared layers being reused in deploys but we have to done one yet.

lox:
Btw, I assume you guys have read http://blog.runnable.com/post/145362675491/distributing-docker-cache-across-hosts?

lox:
And relevant issue https://github.com/docker/docker/issues/20316

packer/conf/buildkite-agent/hooks/post-command
```diff
@@ -1,3 +1,27 @@
...
+
+        slug=/tmp/docker-cache/$BUILDKITE_PIPELINE_SLUG.tar.gz
+        images_file=s3://$BUILDKITE_IMAGE_CACHE_BUCKET/$BUILDKITE_PIPELINE_SLUG.images
+        images=$(echo $(docker images -a | grep $(echo $BUILDKITE_JOB_ID | sed 's/-//g') | awk '{print $1}' | xargs -n 1 docker history -q | grep -v '<missing>'))
+
+        if [[ -n $images ]] && ( ! aws s3 ls $images_file || [[ "$images" != $(aws s3 cp $images_file -) ]]) ; then
+            rm -rf /tmp/docker-cache
+            mkdir -p /tmp/docker-cache
+
+            docker save $images | gzip -c > $slug
```
lox: Given that the layers are content addressable, might it be better to pull them apart into different files so that they can be synced (and therefore deduplicated)? Seems to me like this would have huge savings in terms of transfer time.

lox: Or this would work nicely: https://github.com/tonistiigi/buildcache

lox:
One thing to think about is that this opens up a vector for cache poisoning attacks, for instance where a build could inject a layer into `ubuntu:latest` or `alpine:latest` and subsequent builds would get it. Probably only an issue in terms of third-party builds, which is still theoretical.

lox:
It looks like this is the way docker is going to solve the problem: https://github.com/docker/docker/issues/26065

I'm still not sure how we'd apply that to a CI environment.

lox:
Would it not be better to expose the build cache as a build artifact?

Vektah:
> Btw, I assume you guys have read http://blog.runnable.com/post/145362675491/distributing-docker-cache-across-hosts?

I didn't read that exact post, this one is pretty similar https://blog.codeship.com/building-a-remote-caching-system/ docker 1.11+ seems fine.

> One thing to think about is that this opens up a vector for cache poisoning attacks, for instance where a build could inject a layer into ubuntu:latest or alpine:latest and subsequent builds would get it. Probably only an issue in terms of third-party builds, which is still theoretical.

I'm not sure thats true, I'm pretty sure if there is a pull for your base layer the cache will be pulled, the new base will be pulled and then it will be rebuilt.

> Would it not be better to expose the build cache as a build artifact?

Can they be shared across builds? I thought the whole point was for a build to create and reference its own artifacts?

packer/conf/buildkite-agent/hooks/post-command
Vektah: We talked about it briefly, but it adds a whole lot of complexity:
 - walking the graph (in bash :scream: )
 - determining what to download would be hard, you basically need to rewrite docker build...
 - cleaning up old hashes

It also probably wont save as much as you thing, there are really only two common cases:
 - fetch all the layers (new machine came up)
 - fetch all the layers and rebuild a couple (branch has changed something in the dockerfile)
 - fetch nothing, all the layers are already on the machine

I haven't dug into the source, but buildacache seems like its just a docker save and load with history. Its still outputting a single tar right?

Vektah: https://github.com/tonistiigi/buildcache/blob/master/getcache.go#L75-L89 looks very much like:

```
docker save $(docker history $image)
```

lox:
Regarding cache poisoning https://github.com/docker/docker/pull/24711#issuecomment-237666954.

Vektah:
I think I'm missing the point here, any build can just build and tag anything as ubuntu:latest and other builds would pick it up right?

lox:
I don't believe so, no. Content trust is supposed to prevent that, but load/save works around that. That said, I think it's generally better to assume that it's like you say, that is, no security model.

Vektah:
```
 » cat Dockerfile 
FROM ubuntu

RUN echo "foo" > /test

 » docker build . -t ubuntu
Step 1 : FROM ubuntu
 ---> 948f6753a706
Step 2 : RUN echo "foo" > /test
 ---> Running in 1edc2ebfdda0
 ---> f94e2dcaf192
Removing intermediate container 1edc2ebfdda0
Successfully built f94e2dcaf192

 » docker run -it ubuntu cat /test
foo
```

packer/conf/buildkite-agent/hooks/post-command
tonistiigi: Just happened to be reading this, `buildcache` doesn't save the whole images, only the configuration json. Not sure if it fits better or worse for you because of it. Also `save $(docker history)` was buggy in docker `v1.11` and buildcache works around that.

gugahoi:
@Vektah As far as I understand the build cache can be shared across jobs that run on the same host but that might not be easy to guarantee so there's benefits into sharing the docker build cache across all hosts in the stack.

Vektah:
@gugahoi this PR shares the cache across all hosts by dumping the layers into an s3 bucket and loading them on the other machines as needed

lox:
Looks like `--cache-from` will address this: https://github.com/docker/docker/issues/26065

Shall we close this until that time?

Vektah:
This dosent look like a solution to cache distribution, just the fix required to allow this to work in the next version of docker, regardless of how the images are distributed.

They mention rsync for distribution but it sounds so racey and you would probably end up with an s3 bucket that grows forever with initial sync times that follow.

Vektah:
Looks like there is more hanging off `--cache-from` in https://github.com/docker/docker/pull/26839.

Lets wait and see :)

PR #151, opened 2016-10-03
Correct scaling up logic 

tnh:
It looks like the scale up logic is incorrect. We currently scale if even if there is an agent that is free but has yet to pick the job off the queue.

We inject metrics every 15 seconds by default into cloudwatch as per:

https://github.com/buildkite/buildkite-cloudwatch-metrics-publisher/blob/master/templates/cloudformation.yml#L45

Having an alarm of maximum over that 60 seconds will always trigger an alarm and a scale up event. 

What this will mean is that agents will spin up to max-agent size quickly even though those agents are in fact idle.

Changing the scale up to minimum will mean that we will only scale up if there is a scheduled job that hasnt been picked up by an agent because the agent is busy doing other builds.

### DIFF
--- a/templates/autoscale.yml
+++ b/templates/autoscale.yml
@@ -36,7 +36,7 @@ Resources:
       AlarmDescription: Scale-up if ScheduledJobs > 0 for 1 minute
       MetricName: ScheduledJobsCount
       Namespace: Buildkite
-      Statistic: Maximum
+      Statistic: Minimum
       Period: 60
       EvaluationPeriods: 1
       Threshold: 0


stramel:
@tnh I am testing this in my stack will see if it scales better.

tnh:
buildkitescaling

The aforementioned in the cloudwatch metrics for last Wednesday using the out-of-the-box metrics with the out-of-the-box buildkite metrics agent + stack. 

Our scaling is
+1 scale up, -1 for scale down. Min agents 0, Max agents 10.

lox:
Yup, fair point, it does seem like it's incorrect to me.

stramel:
Hmmm... I was seeing 2 agents always spinning up :/

lox:
What scaling increment @stramel?

tnh:
screen shot 2016-10-06 at 11 27 40 am

Here is a quick update for Irexchange - we're scaling a little better and not spinning up agents that dont get used.

stramel:
@lox I was trying the minimum that @tnh suggested.

tnh:
bumb

lox:
I don't object to merging this in, @sj26 @toolmantim ?

toolmantim:
Thanks for the pull request, explanation and awesome graphs. Makes it MUCH easier to reason about.

So there's one situation I have a question about… for the case of starting with 0 agents and waiting for the first one, instead of waiting 15 seconds this would require then to wait 60 seconds yeah? Because scheduled count needs to be a non-zero number for a full minute for a scale up to occur?

If this is the case, I don't know if it's be a deal breaker, but it would add another 45 worrying seconds to an already quite long first-start time. But maybe there's another solution to that problem, whilst keeping this much more sane scaling behavior.

toolmantim:
Either way, I think this is probably a good thing to add. An extra 45 seconds on initial boot probably won't even be noticed, and is worth it for a probably large reduction in EC2 instance bill.

toolmantim:
Alright, let's go with that. And we can try to optimise the "time to first instance" afterwards.

PR #152, opened 2016-10-07
New ENV variable BUILDKITE_USE_KMS to allow the use of KMS when downloading from secrets bucket

gugahoi:
This will allow users to encrypt their sensitive files in `secretsBucket` with KMS instead of using a passphrase that needs to be pasted as environment variables in plain text in a pipeline. I have tested using the default `aws/s3` key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,19 @@ The secrets bucket can contain the following files:
 
 The files in your secrets bucket should be encrypted with server-side object encryption to ensure they are reasonably secure. See the [Security](#security) section for more details.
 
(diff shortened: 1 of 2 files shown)


toolmantim:
Thanks for this, this sounds better than passwords for sure. This looks like the preferred solution yeah? We should update the Readme to tell people how to do this. Would you be able to do that in this PR too, as you've more KMS experience? Happy to check the words for you.

gugahoi:
Docs updated, let me know if you'd like further modifications

lox:
This looks good to me.

PR #154, opened 2016-10-07
Added feature to copy authorized_users from specified s3 bucket path

wei-xu-myob:
Solved 2 problems:
1. Instead of using http(s), I can use the existing bucket to store authorized_users, e.g. use the secrets bucket
2. Merge with the local authorized_users instead of override, to allow login with the private key which used to create the EC2 instance.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -32,6 +32,7 @@ Metadata:
         - SecretsBucket
         - ArtifactsBucket
         - AuthorizedUsersUrl
+        - AuthorizedUsersS3Path
         - BootstrapScriptUrl
         - RootVolumeSize
 
@@ -96,7 +97,12 @@ Parameters:
     Default: ""
 
   AuthorizedUsersUrl:
-    Description: Optional - S3 url to periodically download ssh authorized_keys from
+    Description: Optional - http(s) url to periodically download ssh authorized_keys from
+    Type: String
+    Default: ""
+
+  AuthorizedUsersS3Path:
+    Description: Optional - S3 path to periodically copy ssh authorized_keys from
     Type: String
     Default: ""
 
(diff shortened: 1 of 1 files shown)


lox:
Am I right in understanding that the reason you'd want to do this is to avoid having `AuthorizedUsersUrl` publicly accessible?

lox:
Might it be simpler to just support s3 urls in `AuthorizedUsersUrl`? I'm not sure I'm keen on part 2 of this, it seems to add a lot of complexity and is a breaking change. If we were going to do this, I'd probably be inclined to use `AuthorizedKeysCommand` with a custom script.

wei-xu-myob:
I'm fine if you change the current AuthorizedUsersUrl logic, in that case you need to download the file based on protocol in the url(http(s), s3), because it should be backward compatible.

The reason to use S3 bucket is for reuse of the existing bucket. It's fine to make the authorized_keys public, but which means I need to setup a bucket only for this file. Because I need to create a secrets bucket for private keys to access github repository, I think it's simple to reuse the same bucket.

Another change I made is using merge instead of override. The existing logic is to override the local authorized_keys, which makes the KeyPair for this instance doesn't work.

wei-xu-myob:
I have changed my request to reuse the AuthorizedUsersUrl parameter, now it's compatible with http(s) and s3

lox:
Thanks! I'll review the changes shortly.

toolmantim:
This looks good to me!

pda:
Makes sense to be able to reuse the secrets bucket for `authorized_keys`.

I'm not a fan of merging the fetched list into existing `authorized_keys`, it could easily be unexpected in a way that is really bad in this security context, and it seems impossible to revoke the original key from the machine. Perhaps that merge aspect could be split out into a separate PR for discussion?

lox:
I've thought about this a bit, and I'm happy to accept the s3 path origin, but I think the merging of `authorized_users` with the launch key is a bad idea, for these reasons:

 * It's harder to reason about what will happen, and it adds more surface area for breakage and unexpected side effects. 
 * Currently `AuthorizedUsersUrl` provides a way to revoke the key used to launch the machine. This is a regression on that capability.

wei-xu-myob:
For me, I think the problem about override is: I created an instance using a key, but could not login by the same key if I specified the AuthorizedUsersUrl.

Could you please provide more details for side effects and security problem for merging? 
Why do we need to revoke the key? and what would happen if we don't?

pda:
> I created an instance using a key, but could not login by the same key if I specified the AuthorizedUsersUrl.

Just make sure `AuthorizedUsersUrl` includes that key. It becomes the single authoritative list of authorized keys, which is a simpler situation than `AuthorizedUsersUrl` merged with whatever authorized key(s) the machine was launched with. Simplicity is pretty important in security.

> Why do we need to revoke the key?

e.g. a build agent / stack is launched by an employee on their last day with the company. Removing their key from `AuthorizedUsersUrl` should cause it to be revoked from the build agents. (This is an imperfect example, because new instances will permit their key between boot and first cron run, but I think it illustrates the issue anyway).

So, it's not about always revoking the launch key, it's about being able to.

> Could you please provide more details for side effects and security problem for merging?

I think the first-time-only backup to `/home/ec2-user/.ssh/original_authorized_keys` adds complexity to the implementation, makes it harder to reason about what's happening, and introduces a new vector for security mistakes/attacks.

lox:
Hah, I just typed out something very similar, but yup, what @pda said :)

lox:
Are you able to move the merging behaviour to a separate PR @wei-xu-myob? If so, we can get this merged.

wei-xu-myob:
Thanks for your explanation. I will create a separate PR for merging, because the following reasons:
- Overriding file changes the default behaviour of EC2 instance: The key should able to login
- In case of revoke the Key, if I put the key in s3 bucket, then I need to remove it in two places.
- If somebody leave the company, and we are using his key to create instance, we **DEFINITELY** should remove that key.

wei-xu-myob:
any update on this pull request?

PR #155, opened 2016-10-12
Add support for EBSOptimized option

ans0600:


templates/buildkite-elastic.yml
```diff
@@ -126,6 +127,11 @@ Parameters:
     Default: t2.nano
     MinLength: 1
 
+  EBSOptimized:
+    Description: Optional - Enabled EBSOptimized for the instance
```
lox: "Enabled Optimized EBS for instance types that don't have it automatically enabled" ?

Vektah: >  that don't have it automatically enabled" ?

What happens if you specify `EBSOptimized: false` for instances that get it by default? are they unoptimized?

ans0600: @lox @Vektah We noticed it will not enable EBSOptimized for you by default. A good example would be m4 family.

lox: > The M4 instances are EBS-Optimized by default, with additional, dedicated network capacity for I/O operations. 

Are you saying that's not the case @ans0600?

ans0600: @lox try launch a spot m4 instance use this template and see if it is set to true. It is not set for us.

lox: @ans0600 are you able to follow up with Amazon and enquire? I'm reluctant to make this change without some more feedback on how it should work.

lox:
Hey, you didn't need to close it :) My concern was that you are passing a false value where there was none previously. 

If you were able to use an `If` and pass `AWS::NoValue` rather than false for when the parameter wasn't provided, would be happy to merge.

PR #157, opened 2016-10-27
Only use two subnets/azs

Vektah:
us-west-1 only has two availability zones, so attempting to create a stack there fails. This drops back to two subnets, which should be deployable everywhere.

the exact error is:

```
    Subnet2 Template error: Fn::Select cannot select nonexistent value at index 2
```

I'm probably being really dumb and there is some super obvious solution to this that I'm not seeing.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -473,7 +473,7 @@ Resources:
     Properties:
       VPCZoneIdentifier: !If [
         "CreateVpcResources",
-        [ $(Subnet0), $(Subnet1), $(Subnet2) ],
+        [ $(Subnet0), $(Subnet1) ],
         $(Subnets)
       ]
       LaunchConfigurationName: $(AgentLaunchConfiguration)
--- a/templates/metrics.yml
+++ b/templates/metrics.yml
@@ -18,7 +18,7 @@ Resources:
         ]
         Subnets: !If [
           "CreateVpcResources",
-          "$(Subnet0),$(Subnet1),$(Subnet2)",
+          "$(Subnet0),$(Subnet1)",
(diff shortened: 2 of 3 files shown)


lox:
Hrm, that would be unfortunate. Could we just have the number of subnets conditional on what is passed in for `Subnets`?

gugahoi:
What about using this instead? http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getavailabilityzones.html?tag=synccit0e-20

lox:
We use that already @gugahoi. The issue is it's very difficult in CF syntax to figure out how many elements the result of that has.

lox:
The best bet I can think of is having a map of regions to whether it has 3 AZs. 

Alternately, after some thought, I could be talked into simply keeping it 2 AZ/Subnets.

toolmantim:
I think dropping down to 2 AZ's to support a wider variety of regions, makes sense… given we've got no way of reflecting on the API from within CF templates

lox:
Do we need to make any documentation changes to support this?

SomeoneWeird:
Could we not have a map for each region that mentions how many AZs in each? We'd have to manually keep it up to date but given how little (they basically don't...) AZs change then I don't think that would be a big deal?

gugahoi:
Personal experience: different users in the same account might have access to different AZs - yes I know it sounds crazy but it has troubled me before. So a map might not work best.

SomeoneWeird:
Oh, interesting. Didn't know you could do that.

lox:
@SomeoneWeird we absolutely could, it's just that it seems simpler to cut down to two AZs.

PR #161, opened 2016-11-03
Allow Custom instance roles for ASG.

pvdvreede:
It would be nice if I could use a single IAM Role across multiple buildkite agent stacks, and also create a role outside of cloudformation so that it has a nicer name and live longer than any one stack. This is useful if a role is given to another account for access.

So this PR provides an optional parameter to supply a IAM Role ARN and use that role for instances rather than create one.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -34,6 +34,7 @@ Metadata:
         - AuthorizedUsersUrl
         - BootstrapScriptUrl
         - RootVolumeSize
+        - AsgRoleArn
 
       - Label:
           default: Auto-scaling Configuration
@@ -160,6 +161,10 @@ Parameters:
     Default: 250
     MinValue: 10
 
+  AsgRoleArn:
+    Description: Optional - A precreated custom IAM role to use for the Buildkite agents.
+    Default: ""
+
   SecurityGroupId:
     Type: String
     Description: Optional - Existing security group to associate the container instances. Creates one by default.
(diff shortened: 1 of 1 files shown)


lox:
Do we need both a `CustomRole` and `SpecifiedIamPolicies`? I'd like to standardize on one way to get IAM customizations in.

lox:
Oh, I just saw that this was replacing the internal IAMRole entirely. Why would you want to do that as opposed to adding an additional role?

pvdvreede:
I wasnt aware you could add more than one role to an instance profile (even though the property is an array). If you could add multiple roles to an instance then I could just do that.

The main use case for this is when we have cross account policies owned by other teams that refer to my Buildkite role in my account by name and it is a real hassle to get the policy changed if i have to delete the Buildkite stack. I suppose another option would be to use the new `name` property that has been added to Cloudformation to hard code a role name (http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#cfn-iam-role-rolename), but then i cannot reuse the same role if i have more than one Buildkite stack (eg with different instance types or jobs).

Hopefully that make sense as an explanation...

i understand you would want to standardise a way rather than having a lot of if elses everywhere, so not really sure what another solution might be, but perhaps i can do a new PR to use the iam role name property instead, would that be better? So set the role name as `buildkite-<stack name>-<queue name>-role`...

lox:
> i understand you would want to standardise a way rather than having a lot of if elses everywhere, > so not really sure what another solution might be, but perhaps i can do a new PR to use the iam 
> role name property instead, would that be better? So set the role name as buildkite-<stack 
> name>-<queue name>-role...

That sounds like a good approach, I'm for anything that reduces parameters :)

lox:
I just noticed that despite being an array, you can only have 1 role per InstancePolicy.

lox:
The IAM Role now has a predictable name `${StackName}-Role`. I'm going to close this one out, let me know if that doesn't solve your problem.

pvdvreede:
Love it! 👍 

Is it worthwhile specifying in the docs that you need to add a new IAM_CAPABILITY when creating the stack now?

pvdvreede:
Ignore me, all good, didnt see the README commit :facepalm:

PR #162, opened 2016-11-07
added support for global env file

caarlos0:
Docker credentials, for example, are usually the same across all repos.

Same with Nexus credentials and others.

Right now I'm duplicating the file across all pipelines, this is painful, but not as painful as it will be if the credentials change one day for whatever reason...

I think this might be helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ The stack has a `SecretsBucket` parameter which will allow your build agents to 
 The secrets bucket can contain the following files:
 
 * `/private_ssh_key` - An optional private key to use for Git SSH operations when there is no pipeline-specific key present
+* `/env` - An optional bash script to use as a global [agent environment hook](https://buildkite.com/docs/agent/hooks)
 * `/{pipeline-slug}/env` - An optional bash script to use as an [agent environment hook](https://buildkite.com/docs/agent/hooks)
 * `/{pipeline-slug}/private_ssh_key` - An optional pipeline-specific private key to use for Git SSH operations
 
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ The stack has a `SecretsBucket` parameter which will allow your build agents to 
 
 The secrets bucket can contain the following files:
 
-* `/private_ssh_key` - An optional private key to use for Git SSH operations when there is no pipeline-specific key present
 * `/env` - An optional bash script to use as a global [agent environment hook](https://buildkite.com/docs/agent/hooks)
+* `/private_ssh_key` - An optional private key to use for Git SSH operations when there is no pipeline-specific key present
 * `/{pipeline-slug}/env` - An optional bash script to use as an [agent environment hook](https://buildkite.com/docs/agent/hooks)
 * `/{pipeline-slug}/private_ssh_key` - An optional pipeline-specific private key to use for Git SSH operations
 
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -60,14 +60,14 @@ if [[ -n "${BUILDKITE_SECRETS_BUCKET:-}" ]] ; then
   fi
   echo
 
-  if s3_exists "$env_url" ; then
-    echo "Downloading env from $env_url"
-    eval "$(s3_download $env_url)"
-    echo
-  fi
   if s3_exists "$shared_env_url" ; then
     echo "Downloading shared env from $shared_env_url"
     eval "$(s3_download $shared_env_url)"
+    echo
+  fi
+  if s3_exists "$env_url" ; then
+    echo "Downloading env from $env_url"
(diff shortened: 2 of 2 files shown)


lox:
What's the difference between the shared env url and the env url just about it?

caarlos0:
`env_url` accounts for the slug of the pipeline, eg: `my-bucket/my-pipeline/env`
`shared_env_url` in that case would be just `my-bucket/env`...

lox:
Ah right, in which case, wouldn't you just set `BUILDKITE_SECRETS_PREFIX=/`? I could see it getting fairly confusing either way. I wonder whether some better way of doing secrets management might be sensible.

caarlos0:
@lox hmm, makes sense... I haven't tried that, will check if it works...

caarlos0:
@lox can I change that config without re-creating everything?

caarlos0:
@lox oh, just realized, this config is for the pipeline itself, not the stack.

So, that's the problem, we have **a lot** of repos (therefore a lot of pipelines), I don't want to set this env for each one... a default global would be nice 😃

sj26:
I like this idea! But I do agree it might get a little confusing. My expectation before looking at the code was that the shared env would always be fetched and sourced for every build on every pipeline, then a pipeline-specific env would additionally be fetched and sourced if present.

What are your thoughts?

caarlos0:
@sj26 Makes sense, I'll make that change. (I was thinking about doing it since the beginning actually 🤔 )

caarlos0:
@sj26 done :)

sj26:
@caarlos0 great, thanks!

One final bit of feedback: I'd expect the pipeline-specific env to override the shared env if both are present. I think your change does the opposite, where shared env will override pipeline env. Do you think it should be reversed?

caarlos0:
@sj26 agreed, done

lox:
Are you able to update the README to describe this?

caarlos0:
sure

caarlos0:
@lox done

README.md
```diff
@@ -130,6 +130,7 @@ The stack has a `SecretsBucket` parameter which will allow your build agents to
 
 The secrets bucket can contain the following files:
 
+* `/env` - An optional bash script to use as a global [agent environment hook](https://buildkite.com/docs/agent/hooks)
```
lox: Are these docs actually correct? Looks like the contents of these files are just eval'd, which is slightly terrible. Thoughts @sj26?

lox: Happy to merge this though, it is at least consistently incorrect.

caarlos0: > Happy to merge this though, it is at least consistently incorrect.

yeah, I saw all those evals and I was thinking on fixing them in another PR... 

FWIW I think that replacing `eval $(s3_download file)` for `source <(s3_download file)` should work the same way as before (assuming no eval-dependent hacks)

caarlos0:
ping

caarlos0:
@sj26 ping

README.md
caarlos0: FYI I replaced the  `eval` for `source`

caarlos0:
ping?

lox:
Apologies for the delay, been pondering on this one. Always like to reflect before adding more config :) 

It's interesting, my expectation would be the opposite @sj26, I would have thought both env would get loaded if a global and pipeline specific existed. 

Thoughts @toolmantim? Otherwise I am happy to merge. Thanks for your hard work @caarlos0!

README.md
sj26: I'm curious about this change — why is source more secure than eval?

sj26:
@lox sorry that's what I was trying to say: I expect the global env to get loaded _and_ the pipeline specific env to get loaded, but I expect pipeline-specific env vars to override global env vars. So a global env with `A=1`, `B=2` and a pipeline env with `B=3`, `C=4` would result in `A=1`, `B=3`, `C=4`. I believe this is what we have now, which is I think most correct

sj26:
Sorry for the delayed response! Things have been a bit hectic. I'm on board with the change, although I'm curious about the source change for downloaded files. sourcing an implicit fd should have little security difference to evaling a variable?

lox:
Yup, I'd be tempted to save that for another change in case it has unexpected side effects. I recall choosing that part pretty carefully, I just don't recall why :D

caarlos0:
In theory, no differences: http://unix.stackexchange.com/questions/123063/difference-between-eval-and-source-dev-stdin

lox:
Thoughts on this approach @caarlos0? https://github.com/buildkite/elastic-ci-stack-for-aws/pull/171

caarlos0:
@lox a minor comment, but it's a good refactor, I very much liked it 🍺

lox:
I ended up sticking with the existing eval @caarlos0, will address this in future.

lox:
Thanks for the contribution and code review!

PR #163, opened 2016-11-07
Fix example command line.

jeffreyctang:
<( command ) is different and would be used in place of a filename.


lox:
Oooops! You are absolutely correct, thanks for the catch.

PR #167, opened 2016-11-15
Easier ECR integration

lox:
At the minute it's a little bit tricky to get ECR integration working, this makes it more seamless by adding a parameter called `ElasticContainerRegistryAccess` that is either `none`, `readonly`, `poweruser` or `full`. If this is set, everything else will be setup for ECR to work out of the box. 

Remaining todo:

 * [x] Document the new parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ rm id_rsa_buildkite*
 
 ## Docker Registry Support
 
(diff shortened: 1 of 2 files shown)


lox:
Any objections to me merging this @toolmantim / @sj26?

PR #168, opened 2016-11-15
Convert YAML to native CloudFormation format

lox:
This will remove the cfoo dependency and get us back to native Cloudformation YAML templates.

This changes the file that is published from .json to .yml. Aside from that it's basically backwards compatible. Next step will be to merge the files, but that makes merging changes hard so holding off.


lox:
This initial implementation is a bad idea, it will make merges really hard. Going to try this a different way.

sj26:
CloudFormation supports YAML now? Interesting!

lox:
I'm going to close this, its just getting out of date and adds minimal value at this stage.

PR #174, opened 2016-11-22
Remove any mention of `BUILDKITE_SECRETS_KEY`

ticky:
We no longer recommend this as an option (it’s best not to confuse people!)


lox:
Awesome, many thanks @ticky!

lox:
Should we go so far as to show a deprecation warning for people still using it?

ticky:
Possibly! But I’m not sure where that’d live (@toolmantim was doing some back-seat editing 😜)

lox:
Cool, let's merge this for now and address a deprecation warning at some later stage. (Typical @toolmantim)

Issue #177, opened 2016-11-23
Choosing an ECRAccessPolicy w/ causes stack update/create failure


toolmantim:
![failure](https://cloud.githubusercontent.com/assets/153/20551148/da379ff6-b190-11e6-9989-d5a1758beac4.png)

The error is "Value of property ManagedPolicyArns must be of type List of String"

denmat:
I was about to raise the same question - this did not fix it @lox https://github.com/buildkite/elastic-ci-stack-for-aws/pull/176.

I have tried a few combinations but I can't get it to work - that is messing around with these lines:
https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/buildkite-elastic.yml#L245

lox:
This is fixed in master, could you let me know if this fixes your problem @denmat /  @toolmantim?

denmat:
Fixed for me, thank you.

PR #180, opened 2016-11-23
Fix issue #177

lox:
Another attempt at fixing #177. This breaks BC by replacing `ManagedPolicyArns` with `ManagedPolicyARN`. This is because there is no mechanism to merge a list with a single item in CloudFormation.

To mitigate this, I've added a predictable RoleName for the IAMRole (as suggested in #161).

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -253,6 +253,13 @@ Resources:
     Condition: UseECR
     Properties:
       ManagedPolicyArns: !If [ UseECR, ['$(ECRManagedPolicy[$(ECRAccessPolicy)][Policy])'], '$(AWS::NoValue)' ]
+      AssumeRolePolicyDocument:
+        Statement:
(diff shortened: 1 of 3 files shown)


templates/buildkite-elastic.yml
```diff
@@ -170,16 +170,16 @@ Parameters:
...
-    Default: ""
-
   ImageId:
     Type: String
     Description: Optional - The AMI to use, otherwise uses the mapping built in
     Default: ""
 
+  ManagedPolicyARN:
+    Type: String
+    Description: Optional - ARN of an existing managed IAM policies to associate with instances
```
toolmantim: Should "policies" be singular here?

lox: That's basically correct, except they wouldn't have had to create their own policy for ECR, they just would have had to reference one of the managed AWS policies, passed in via the previous ManagedPolicyArns parameter.

PR #181, opened 2016-11-24
Default to KMS for secrets

### LINKED ISSUE #160: Set KMS Download of Secrets to be the default approach
Now that we have KMS integration into the BuildKite agents for secret management I would like to have that as the default way of copying from the S3 secrets bucket instead of having to specify the environment variable `BUILDKITE_USE_KMS`. 

This implies that by default (without having to configure pipelines) BuildKite agents would attempt to decrypt the secrets from S3 using KMS, we could still allow the use of PASSPHRASE for encryption with the same approach it has now (if the ENV variable exists than use it).

Happy to create the pull request if you would like.

lox:
Makes the stack default KMS for all the secrets bucket. Closes #160.

### DIFF
--- a/packer/conf/buildkite-agent/hooks/pre-command
+++ b/packer/conf/buildkite-agent/hooks/pre-command
@@ -5,7 +5,7 @@ set -eu -o pipefail
 # AWS ECR login support
 
 # For logging into the current AWS account’s registry
-if [[ -n "${AWS_ECR_LOGIN:-}" ]] ; then
+if [[ "${AWS_ECR_LOGIN:-true}" =~ ^(true|1)$ ]] ; then
   echo "~~~ Authenticating with AWS ECR"
   eval "$(aws ecr get-login)"
 fi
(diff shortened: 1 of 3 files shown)


packer/conf/buildkite-agent/hooks/pre-command
```diff
@@ -5,7 +5,7 @@ set -eu -o pipefail
 # AWS ECR login support
 
 # For logging into the current AWS account’s registry
-if [[ -n "${AWS_ECR_LOGIN:-}" ]] ; then
+if [[ "${AWS_ECR_LOGIN:-true}" =~ ^(true|1)$ ]] ; then
```
toolmantim: I think for a little while we supported `AWS_ECR_LOGIN=<region>`, which would break this. Something to keep in mind for the release notes!

gugahoi:
Awesome! Sorry I didn't get around to this. Thanks!

PR #171, opened 2016-11-17
Cleanup ssh key check and env loading

lox:
This cleans up the SSH key checks and env loading by iterating over a set of possible keys. Makes it much easier to reason about and add consistent error checking.

PR #183, opened 2016-11-24
using source instead of eval

caarlos0:
refs #171 and #162

### DIFF
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -42,7 +42,7 @@ s3_download() {
 echo "~~~ Setting up the environment"
 
 echo "Sourcing CloudFormation environment..."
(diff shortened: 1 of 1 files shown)


lox:
I'm going to let the current stack settle down before this change. From what I can tell there isn't much in the way of actual security improvement?

caarlos0:
@lox I don't think we get the security advantages of not using `eval` in this particular case, but, either way, using `source` is more correct AFAIK.

There is a good explanation [here](http://stackoverflow.com/questions/17529220/why-should-eval-be-avoided-in-bash-and-what-should-i-use-instead).

lox:
I'm going to close this one for now, I plan on rewriting some of the env and ssh cred loading stuff and I'm going to incorporate it at that point. Thank you very much for the PR though and the research!

PR #187, opened 2016-11-25
Fix issue with s3 cp command source location in environment hook

holmesjr:
The signature for s3_copy() changed but the aws s3 cp command call didn't change to match. This fixes that issue. It also removes the broken bucket_region local var.

### DIFF
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -15,8 +15,7 @@ s3_exists() {
 s3_download() {
   local bucket="$1"
   local key="$2"
-  local bucket_region
-  local aws_s3_args=("--quiet" "--region=$bucket_region")
+  local aws_s3_args="--quiet"
 
   if [[ -n "${BUILDKITE_SECRETS_KEY:-}" ]] ; then
     echo "~~~ :warning: Deprecated BUILDKITE_SECRETS_KEY env variable"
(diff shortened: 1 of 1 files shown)


packer/conf/buildkite-agent/hooks/environment
```diff
@@ -15,8 +15,7 @@ s3_exists() {
 s3_download() {
   local bucket="$1"
   local key="$2"
-  local bucket_region
-  local aws_s3_args=("--quiet" "--region=$bucket_region")
```
lox: Why the drop of region?

holmesjr:
It's in line with the change at https://github.com/buildkite/elastic-ci-stack-for-aws/commit/ea06ca158b9ec26fdaad8c18db6575493cad0f90. The bucket_region var is never defined and causes an error.

lox:
Oh of course. that should be `$AWS_DEFAULT_REGION`, man I can't figure out how I so badly merged this changeset.

lox:
I think because initially I was getting the region from the bucket for https://github.com/buildkite/elastic-ci-stack-for-aws/pull/170, but then in master I used the simpler `$AWS_DEFAULT_REGION`.

lox:
The merge must have gotten half of each.

lox:
Let's merge this as is and address the region stuff subsequently.

PR #188, opened 2016-11-25
Fail if a secret key isn't found but a secrets bucket is set

lox:
Currently if all checks for a private ssh key fail, the checkout process kicks off anyway. This changes that so that it fails.

### DIFF
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -65,7 +65,7 @@ if [[ -n "${BUILDKITE_SECRETS_BUCKET:-}" ]] ; then
     fi
   done
 
-  if [[ -z $key_found ]] ; then
+  if [[ -z "${key_found:-}" ]] ; then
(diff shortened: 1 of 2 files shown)


packer/conf/buildkite-agent/hooks/environment
```diff
@@ -60,15 +69,23 @@ if [[ -n "${BUILDKITE_SECRETS_BUCKET:-}" ]] ; then
     if s3_exists "${BUILDKITE_SECRETS_BUCKET}" "$key" ; then
       echo "Downloading ssh-key $key"
       SSH_ASKPASS="/bin/false" ssh-add <(s3_download "${BUILDKITE_SECRETS_BUCKET}" "$key")
+      key_found=1
       break
     fi
   done
 
+  if [[ -z "${key_found:-}" ]] && [[ "${BUILDKITE_REPO:]}" =~ ^git ]] ; then
```
lox: Good point, think this covers it @toolmantim?

PR #195, opened 2016-12-05
Skip creating autoscaling resources if not required

lox:
This allows MaxSize=MinSize clusters which don't scale. If this is the case, then the autoscaling resources aren't created.

### DIFF
--- a/templates/autoscale.yml
+++ b/templates/autoscale.yml
@@ -16,6 +16,7 @@ Resources:
 
   AgentScaleUpPolicy:
     Type: AWS::AutoScaling::ScalingPolicy
+    Condition: UseAutoscaling
     Properties:
       AdjustmentType: ChangeInCapacity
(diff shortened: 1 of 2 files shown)


toolmantim:
I updated a fixed-sized stack successfully to this master build, without changing it's params (min 1, max 2, scale up 0, scale down 0, and blank org slug and API key)

I then tried to update again, to set min and max to the same value of 1, but got this error:

failed

Is that expected?

lox:
Nope, that was the error I was trying to fix.

toolmantim:
Ah. Strangely it didn't affect new stacks, just the one that previously existed. Could have also been operator error!

lox:
No, it was all me :)

lox:
I've addressed this in #201.

lox:
It only happens on update btw @toolmantim.

Issue #117, opened 2016-08-13
Agent instances have public IPs


holmesjr:
It's not a major issue, but since I deploy my BK agents in a private subnet with NAT gateways, public IPs are unnecessary (and potentially confusing to anyone looking at the agents). Perhaps a configuration option as a stack parameter?

toolmantim:
Makes sense as a checkbox parameter! Would be happy to accept a PR

PR #198, opened 2016-12-10
Create a parameter to determine whether public IPs are assigned to instances

holmesjr:


lox:
I suspect that if you don't associate public IP's then they won't be able to access the external internet without a NAT Gateway.

holmesjr:
Yep, that's the intent. For security, we delpoy our agents exactly that way. I should have mentioned that this fixes #117.

lox:
Interesting, how does the agent contact buildkite.com without access to the internet?

holmesjr:
We have a NAT gateway in our public subnet. Since the VPC provisioned in this template doesn't, perhaps a note in the readme that the option is only suitable when we have existing VPC resources is appropriate. I can add that to the PR if you like.

lox:
Ah right, that makes sense. I was going to basically introduce this underlying VPC structure:

https://github.com/lox/cloudformation-templates/blob/master/vpc.yml

holmesjr:
Awesone, I'm all for a default structure that enforces better security. There's little reason to run instances in a public subnet.

lox:
Could we just defer the associating of public ip's to the VPC/Subnet? That way a config like yours would work as expected.

holmesjr:
That would be a reasonable change, as long as people know that it's up to their subnet to dictate the behaviour. In that case, just removing the attribute from the launch config should be enough.

lox:
I wonder how CloudFormation copes with updating underlying VPC pieces.

holmesjr:
A change to subnet behaviour would require no interruption: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-subnet.html#cfn-ec2-subnet-mappubliciponlaunch

lox:
My concern with merging this is that if people select "false" for this and don't provide their own vpc/subnets, they will have non-functioning agents. I think we need to get a NAT Gateway into the default VPC before we could do this.

holmesjr:
Agreed, it's a better pattern and a good example of how people can set up secure VPCs.

lox:
Have decided to merge this in, the other PR will take some more thought. Thanks for the contribution @holmesjr.

jsierles:
Did the other PR for adding a NAT Gateway ever surface? If not, is the best option here to setup a VPC with a NAT gateway and assign it to the stack?

PR #202, opened 2016-12-11
WIP: Add a public subnet and a nat gateway to VPC

lox:
As per #198, the default configuration for the agents should be to not have a public ip address. To support this, a NAT gateway is needed, and a public subnet for the NAT gateway to sit on.

### DIFF
--- a/templates/vpc.yml
+++ b/templates/vpc.yml
@@ -80,7 +80,7 @@ Resources:
     Condition: CreateVpcResources
     DependsOn: NatGatewayEIP
     Properties:
-      AllocationId: $(NatGatewayEIP.AllocationId)
+      AllocationId: !GetAtt [ NatGatewayEIP, AllocationId ]
       SubnetId: $(Subnet2)
 
   RoutesPublic:
--- a/templates/vpc.yml
+++ b/templates/vpc.yml
@@ -1,11 +1,20 @@
 
+Mappings:
+  SubnetConfig:
+    VPC: { CIDR: 10.0.0.0/16 }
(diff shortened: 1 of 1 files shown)


lox:
If you get a chance @holmesjr, would be great to get you to cast an eye over this.

holmesjr:
We have more than one NAT gateway, so that if an AZ goes down, we still have working instances. We route each private subnet to a NAT in the same AZ's public subnet.

lox:
Good point @holmesjr. Thus far I haven't been designing much redundancy into elastic stack, but it's worth considering.

holmesjr:
Yes, people would need a bastion instance in the public subnet (which is what we do now).

toolmantim:
We do the same for a lot of our own stuff, but for lots of people using the stack I doubt they'll have one setup at the moment they need to do some debugging. Is there an easier solution to that?

holmesjr:
None that spring to mind. The private subnet is designed to keep people out.

Maybe the template than can have a condition that deploys things into public or private subnets based on a param?

lox:
Couldn't they just update their stack to add re-enable the `AssociatePublicAddresses` parameter?

holmesjr:
Only if the instances have been deployed in the public subnet. There's no way to get to a private subnet instance from outside the NAT gateway, since the private subnet's default route goes through that gateway.

Or did I misunderstand the suggestion?

lox:
Nope, I think I was just under the illusion that `AssociatePublicIpAddress` performs some routing magic to enable access to it.

lox:
Closing for now.

PR #206, opened 2016-12-20
Move runtime installation from Cloudformation metadata to to Packer/AMI

lox:
As part of the move towards multiple ASG's in one stack, this moves configuration from the `LaunchConfiguration` into a script on the AMI. 

This should have effectively no impact on functionality, it's just moving code from cfn-init into a shell script.

### DIFF
--- a/packer/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/conf/bin/bk-install-elastic-stack.sh
@@ -24,7 +24,7 @@ AWS_DEFAULT_REGION=$AWS_REGION
 AWS_REGION=$AWS_REGION
 EOF
 
-if [[ -n "$BUILDKITE_USE_ECR" ]] ; then
+if [[ "${BUILDKITE_ECR_POLICY:-none}" != "none" ]] ; then
 	printf "AWS_ECR_LOGIN=1\n" >> /var/lib/buildkite-agent/cfn-env
 fi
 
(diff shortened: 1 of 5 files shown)


packer/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -0,0 +1,99 @@
...
+[plugins]
+cwlogs = cwlogs
+[default]
+region = $AWS_REGION
+EOF
+
+# cfn-env is sourced by the environment hook in builds
+cat << EOF > /var/lib/buildkite-agent/cfn-env
+BUILDKITE_STACK_NAME=$BUILDKITE_STACK_NAME
+BUILDKITE_SECRETS_BUCKET=$BUILDKITE_SECRETS_BUCKET
```
sj26: We still need `$BUILDKITE_AGENTS_PER_INSTANCE` here for `/etc/terminationd/hook` to work. I've re-added it in 3935a9857b28bbecc2cd5cbb38cd58a315675a3f.

We noticed when our builds started seeing docker shut down mid-job, and found this in the logs:

```
Jan 12 23:37:00 ip-10-0-1-182 terminationd: Stopping buildkite-agent gracefully
Jan 12 23:37:00 ip-10-0-1-182 terminationd: /etc/terminationd/hook: line 12: BUILDKITE_AGENTS_PER_INSTANCE: unbound variable
Jan 12 23:37:00 ip-10-0-1-182 terminationd: Waiting for all buildkite-agent processes to have stopped...
```

Issue #208, opened 2016-12-22
BuildkiteMetricsFunction Lambda prevents stack being created other regions


MattFisher:
This issue affects the master branch, but not v1.1.1

Trying to create the stack in `ap-southeast-2` (Sydney) fails because according to [CloudFormation docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html#cfn-lambda-function-code-s3bucket), when creating a lambda from an s3 bucket, "The S3 bucket must be in the same region as the stack".

The `buildkite-metrics` bucket, where `buildkite-metrics-v1.4.1-lambda.zip` is stored, is in `us-east-1`.

Error message during CloudFormation stack creation:
```
2016-12-22: 10:54:21 UTC+1100
Status: CREATE_FAILED
Type: AWS::Lambda::Function
Logical ID: BuildkiteMetricsFunction
Status reason: Error occurred while GetObject. S3 Error Code: PermanentRedirect. S3 Error Message: The bucket is in this region: null. Please use this region to retry the request
```

lox:
Apologies for the breakage! The lambda stuff is new, will get a fix out for this soon.

lox:
The fix has taken some more time than usual, waiting for the BK team to come back to work. In the meantime, #210 provides params for the lambda function, which you could copy to your own bucket and reference.

PR #210, opened 2016-12-28
Params for bucket and key for the metrics lambda

lox:
Provides a workaround for #208 until we get lambda's published in all regions.


lox:
@toolmantim https://github.com/buildkite/buildkite-metrics/pull/26

Issue #172, opened 2016-11-18
Adding SSH key from S3 hangs in console


mattlevin:
I am having an issue similar to https://github.com/buildkite/elastic-ci-stack-for-aws/issues/109 where the ssh key downloads correctly from the default secrets s3 bucket, but then is hanging on user input when adding the key. My build output is below. I have also tried setting BUILDKITE_SECRETS_KEY and BUILDKITE_USE_KMS and it has given the same result.

```
Sourcing CloudFormation environment...
 
Starting an SSH Agent...
Agent pid 2940
 
Downloading secrets...
 
2016-11-09 09:15:29    2.1 KiB private_ssh_key
Downloading ssh key from s3://signpost-buildkite-trial-secrets/private_ssh_key
Enter passphrase for /dev/fd/63:
```

Thanks for your help.

Matt

holmesjr:
Your key will need to be generated without a passphrase. When you do so ssh-keygen will whine,  so just ignore it and hit enter.

mattlevin:
@holmesjr 
I generated without a passphrase using the following commands and the behavior is the same:

```
ssh-keygen -t rsa -b 4096 -f id_rsa_buildkite

aws s3 cp --acl private --sse aws:kms id_rsa_buildkite "s3://${SecretsBucket}/private_ssh_key" 
```

mattlevin:
Alright I figured this out. Because its running quiet when copying from s3 (https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/conf/buildkite-agent/hooks/environment#L23) we don't see any of the errors that come up if this copy goes wrong. So for me it was failing for a reason related to kms. However, the result of this command is getting run as the parameter for ssh-add (https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/conf/buildkite-agent/hooks/environment#L50) so its the ssh-add complaining about the fact that the key didn't download from s3. 

Seems like there needs to be a way to see the s3 copy failure messages.

lox:
Sorry you had to dig through this! I've run into it myself a few times,
it's because the s3 call is wrapped in an if statement in the environment
hook. Fix coming soon!

lox:
Hrm, I was sure I fixed this. Re-opening.

PR #215, opened 2017-01-05
Enable S3 v4 Signatures by Default

lox:
Seems like more people are seeing issues with s3 requiring v4 signatures in certain circumstances. This defaults to v4. 

/cc @benschwarz


benschwarz:
When this master, I can verify it 👁👃 👁

benschwarz:
Ref: #172

PR #218, opened 2017-01-10
Create the secrets bucket if left blank

lox:
This creates a secrets bucket if none is provided. This bucket will be left in place when the stack is deleted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,7 @@ aws cloudformation create-stack \
 
 ## Build Secrets
 
-The stack refers to a `SecretsBucket` parameter which will allow your agents to access SSH private keys for source control, and environment hooks to provide secrets to your builds.
-
-> You will need to create this S3 bucket yourself, but the stack will automatically create permissions for access to it. Make sure to create the bucket in the same region as your stack.
+The stack will have created an S3 bucket for you (or used the one you provided as the `SecretsBucket` parameter). This will be where the agent will fetch your SSH private keys for source control, and environment hooks to provide other secrets to your builds.
 
 The following paths in the bucket are checked. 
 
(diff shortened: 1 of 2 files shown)


lox:
I'm not 100% sure this is useful. Thoughts @toolmantim?

toolmantim:
I think it is! I think the "Getting Started" guide will be improved, because it's just "Add a key to this bucket and run your GitHub project's build" instead of "Create a bucket, edit the stack, etc etc"

lox:
With the caveat that you need to explain how to figure out the bucket name

toolmantim:
Yeah… it's still better I think. "Find your stack’s secrets bucket, add your key, run a build" vs "Create a secrets bucket, add the key, edit your stack, update your agents, run a build"

lox:
Yup, true.

lox:
Would it be going to far to have a policy on the bucket that prevented non-encrypted contents?

```
{
   "Version":"2012-10-17",
   "Id":"PutObjPolicy",
   "Statement":[{
         "Sid":"DenyUnEncryptedObjectUploads",
         "Effect":"Deny",
         "Principal":"*",
         "Action":"s3:PutObject",
         "Resource":"arn:aws:s3:::YourBucket/*",
         "Condition":{
            "StringNotEquals":{
               "s3:x-amz-server-side-encryption":"aws:kms"
            }
         }
      }
   ]
}
```

toolmantim:
Oh, if it's not standard I think that'd be a very nice option!

If it is standard, can you upload it as encrypted via the AWS web console? Or `cli` only?

lox:
Thoughts on getting this in as 2.1.0 @toolmantim? It should be non-breaking.

PR #225, opened 2017-01-12
Restart awslogs instead of just start for config changes.

pvdvreede:
I was getting my logs coming through in us-east-1.

After I restarted the process they moved to the right region. I was wondering if its because awslogs may already have been started before this scripts run (perhaps a race condition?), so that starting it is a no op?

Instead I have changed it to restart, so that it will either start if stopped, or restart if started to pickup any config changes.


lox:
Makes sense, thanks for the contribution!

PR #226, opened 2017-01-12
Allow agent to get objects from its artifacts bucket.

pvdvreede:
Not sure if there is a reason not to allow this by default, but sometimes I have pipelines that create an artifact in once stage, and then need it in the next, so it would be good to be able to pull down something from the artifacts bucket during a step.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -340,6 +340,7 @@ Resources:
             Action:
               - s3:Put*
               - s3:List*
+              - s3:Get*
             Resource:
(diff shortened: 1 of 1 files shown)


lox:
I think this was an oversight. Frankly, we were looking at removing support for the artifact bucket in aws-stack because it's not very well integrated. Have you found it's much work to get working?

pvdvreede:
I use terraform wrapped around the cloudformation, so that makes it quite easy to create the bucket for artifacts and secrets. So that has been good.

PR #229, opened 2017-01-15
Set env vars for custom artifact buckets as needed

lox:
Whilst we supported a `ArtifactsBucket` parameter, none of the environment variables mentioned in https://buildkite.com/docs/agent/cli-artifact#using-your-own-s3-bucket were added. This does that.


lox:
I'm torn on this. On one hand I agree, but on the other, people are presently using this functionality, and it works ok, just not great.

packer/conf/buildkite-agent/hooks/environment
```diff
@@ -41,6 +41,12 @@ export AWS_REGION
 export AWS_DEFAULT_REGION
 export AWS_ECR_LOGIN
 
+if [[ -n "${BUILDKITE_ARTIFACTS_BUCKET:-}" ]] ; then
+  echo "Setting up custom artifacts bucket in ${BUILDKITE_ARTIFACTS_BUCKET}"
+  export BUILDKITE_ARTIFACT_UPLOAD_DESTINATION="s3://${BUILDKITE_ARTIFACTS_BUCKET}/$BUILDKITE_JOB_ID"
+  export BUILDKITE_S3_DEFAULT_REGION="${AWS_DEFAULT_REGION}"
```
toolmantim: Is the artifact bucket region always going to be the same as the stack region?

toolmantim:
It's really new stack users I'm talking about.

Without the envs I found it confusing. With the envs, I *know* the next question will be "how do we access our artifacts?".

I'm still tempted to remove it, and work on adding proper support via `BUILDKITE_S3_URL_SIGNING_ROLE_ARN`. And in the meantime put a section in the readme that links to the custom artifacts bucket guide, and tells them how to update their instance roles and env hook.

lox:
Yup, cool, I agree.

lox:
Still very reluctant to let this one go....

PR #233, opened 2017-01-16
Force native cgo resolver for docker

lox:
See https://github.com/docker/docker/issues/22673 for more details.

/cc @toolmantim


lox:
I'm not sure if this will work, but reading the crappy init script for docker, it seems like `/etc/sysconfig/docker` is just sourced, so the exported ENV should make it to docker.

PR #234, opened 2017-01-17
WIP: Dedicated instances for docker-builders

lox:


dalberto:
I'm currently setting up a `builders` stack in addition to a `runners`, per the note in the README about setting up dedicated Docker instances. 

If I'm not mistaken, this PR simply allows an existing deployed stack to have 2 agent queues. If so, besides simplicity and ease, is there an added benefit this PR introduces on top of the multiple-stack solution? 

I ask because it may be worthwhile to wait for this PR to land before setting up agents.

lox:
I actually suspect that we might end up ditching this and make use of the new `--cache-from` directive that docker 1.13 introduces. I'm accordingly going to close this. Keep an eye on #81.

PR #237, opened 2017-01-20
Add Name tags to more AWS resources

SomeoneWeird:


toolmantim:
Nice! Should we do the same with the EC2 instances?

SomeoneWeird:
Currently the EC2 instances [get tagged statically with `buildkite-agent`](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/buildkite-elastic.yml#L443-L444), it would be cool to maybe prepend the stack name so you can differentiate them if you have multiple stacks?

[](url)

toolmantim:
Yes! If we could add the stack name that'd be ace. Potentially even before `buildkite-agent`?

lox:
I'm going to merge this as-is. We can talk more tags later.

SomeoneWeird:
Oops! I totally forgot about this, sorry folk!

PR #244, opened 2017-02-10
Add ASG Name output.

pvdvreede:
I use terraform wrapped around this cfn template, and would like to pull out the ASG name by using the cloudformation data source.

I have created a new file on the thought that people might like other output values and that this list my grow.


pvdvreede:
FYI relates to #243 .

PR #248, opened 2017-02-11
Move agents to docker containers

lox:
This moves the buildkite-agent to a docker container rather than on the host. This paves the way for moving to a different base image (such as RancherOS) or ECS/Swarm.

**TODO**:
 * [x] Decide on approach for `BootstrapScriptUrl`
 * [x] Do we still have concurrency issues with access to the build checkouts? Need to figure out how `docker-compose scale` works with volumes.
 * [ ] Determine how to make this work with the docker-compose and docker plugins


holmesjr:
Oh gods, please don't.

lox:
That is not a constructive comment @holmesjr :)

lox:
But keen to hear why you object.

holmesjr:
Now, I have your attention :-)

Are you proposing to run docker inside docker, or alternatively to spawn builds alongside the agent container?

lox:
Nah, docker-in-docker is dead to me. This is just replacing running the agents on the host with the agent in a docker container. The host docker socket will be passed in, so the performance and overall experience should be close to identical as currently.

holmesjr:
Phew. Please carry on :-)

lox:
Basically it removes a lot of fragile code from elastic stack and lets us look a things like ECS.

lox:
o7

lox:
If there were performance regressions, it would be a deal-breaker for me.

holmesjr:
It's a great idea. For anyone reading along who wonders why docker in docker is bad, this is a start: https://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/

lox:
This is pretty much done. I'd love some feedback/review on approach and implementation.

lox:
Also how to release this. Is this a 2.1.0 or is this a 3.0.0? It's designed so there are no breaking changes, although I'm sure I've forgotten something.

toolmantim:
One thing I thought of is, is that I know some people are adding custom agent hooks via the bootstrap script. We could think about how to support that properly?

And anything else related to the Buildkite agent running on different paths to before (build dir, plugin dir, etc).

Either way, I reckon we should move forward with this in a 3.0rc release?

lox:
The hooks get mounted in from the host, so that should still work.

packer/conf/buildkite-agent/docker-compose.yml
```diff
@@ -0,0 +1,18 @@
...
+    restart: on-failure:10
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - buildkite-builds:/buildkite/builds
+      - /etc/buildkite-agent/hooks:/buildkite/hooks
+      - ./entrypoint.sh:/docker-entrypoint.d/elastic-stack
+    environment:
+      - BUILDKITE_AGENT_TOKEN
+      - BUILDKITE_AGENT_META_DATA
+      - BUILDKITE_VOLUME=buildkite-data
```
toolmantim: What's this?

toolmantim:
Oh nice, I didn't realise that.

SomeoneWeird:
If i'm reading this correctly, this would be a breaking change for us. We install a few things on the host that pipelines use, so they wouldn't be available inside of the containers we run.

packer/conf/buildkite-agent/docker-compose.yml
lox: I think it's a hangover from a previous iteration, will remove it.

lox:
@SomeoneWeird yup, need to figure out how to handle that. I'm thinking either:

*  Running `BootstrapScriptUrl` in the agent containers, so that you could install stuff, or
*  Invoke all builds in their own docker container, allow a custom image to be specified

Suspect option 1 is easiest step forward, I'd like to put pluggable "executors" in the agent to accomplish 2.

lox:
After some thought, I think we'll keep `BootstrapScriptUrl` as it is, but I've added an easy way to specify an alternate docker image for the agents. I think this is the best way to customize the run-time environment for builds anyway. Thoughts @SomeoneWeird?

SomeoneWeird:
@lox Totally missed this! He exactly do you see `BootstrapScriptUrl` working?

lox:
`BootstrapScriptUrl` will run in the context of the host, not your agent containers. If you wanted to add things to your agent environment, you could build a custom image in that BootstrapScript and then set that to be the image used for running the agent.

lox:
Closing for now.

Issue #236, opened 2017-01-18
Intermittent ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?


piotrb:
Having a bit of an odd random occuring issue ..

setup is elastic cloud 2.rc1-9, beta agent, docker compose

occasionally .. steps fail to connect to docker with:

```
Creating network "buildkite5736e56df5c2478db2635348bf21e4d2_default" with the default driver
ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?
```

except that .. earlier in that step it was able to docker pull just fine .. and later in that step in cleanup its able to try to clean up without connection errors (it said no network to delete, etc etc, which means it connected and got inventory right?) ..

Not sure if this is a elastic-ci issue, or agent, or plugin or what ..

piotrb:
This is quite annoying since its causing random failures ..

lox:
I'm sorry you've been having issues, we will investigate. I believe
@bjeanes might have had similar issues.

keithpitt:
@piotrb whats your agent count / server setup like? Are you running lots of agents on 1 machine? Or 1 agent per machine?

bjeanes:
I get this occasionally. One of my stacks is 4 agents/server but I can't recall if this error was isolated to that stacks or not.

piotrb:
Running 4 agents per m4.xlarge instance

kbaum:
I also get this error when running 4 docker instances on a t2.xlarge.

lox:
Any idea if these instances are scaling down when they error?

piotrb:
I don't know if I can ever really catch that .. 

Oh yea .. I guess I havn't seen this issue in a while now .. so somehing has changed

Although it seems like docker comes back up since it tries do do the cleanup steps .. so its not like its shut down already or something

kbaum:
In my situation, the instances were just launched.

lox:
That makes sense too, it would happen either before docker starts or after
it shuts down. I will do some research.

kbaum:
Cut down to 4 docker instances on t2.xlarge ec2 instances but still regularly seeing this error.  Any known workarounds?

lox:
Hrm, this could be from a variety of things. I'd love to see the docker log after this happens. Any chance you'd have a look on the next failure @kbaum? Would like to find a solution to this for you folks.

kbaum:
[...]

PR #250, opened 2017-02-15
Use ECS-Optimized AMI for base image

lox:
Docker considers Amazon Linux an "[unsupported distro](https://github.com/docker/docker/pull/23480)" and it seems like the relationship between the two is becoming increasingly strained. 

Currently we are using the [latest Amazon Linux AMI](https://aws.amazon.com/amazon-linux-ami/2016.09-release-notes/) and then installing Docker on it. An official Amazon package isn't available for 1.13, and tends to lag behind because kernel changes are often required in the underlying distro. 

This PR moves to the newer [ECS Optimized image](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html) from Amazon, which is basically a supported bundle of Amazon Linux, Docker 1.12.6 (currently) and ECS agent.

My hope is that this will be more stable.

### DIFF
--- a/packer/buildkite-ami.json
+++ b/packer/buildkite-ami.json
@@ -3,12 +3,12 @@
     {
       "type": "amazon-ebs",
       "region": "us-east-1",
-      "source_ami": "ami-9be6f38c",
+      "source_ami": "ami-b2df2ca4",
       "region": "us-east-1",
       "instance_type": "c4.large",
       "ssh_username": "ec2-user",
-      "ami_name": "buildkite-stack-{{isotime | clean_ami_name}}",
-      "ami_description": "Buildkite CloudFormation Stack base image (Amazon Linux, buildkite-agent, docker, docker-compose, docker-gc, jq)",
+      "ami_name": "buildkite-for-aws-{{isotime | clean_ami_name}}",
+      "ami_description": "buildkite/elastic-ci-stack-for-aws",
       "ami_groups": ["all"]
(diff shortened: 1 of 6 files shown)


toolmantim:
I considered this long ago when I more closely followed the ECS project, and I think I even copy and pasted bits from their image into our stack at some point.

I'm not sure if this is worth it though. It's definitely a more battle-hardened setup, but unless we're moving the Elastic CI Stack to ECS why wouldn't we just switch to CoreOS?

lox:
Yup, agreed @toolmantim. Beyond that, the ECS image seems to insist on devicemapper backend, which seems like a dead-end to me.

lox:
After all the issues we've been seeing in #236 I've been really reconsidering this.

toolmantim:
It could be worth experimenting with, especially if @piotrb has a repeatable environment for testing. I'm just not sure how many other bugs it might create in the process? Or if there's something we should be using instead of ECS? CoreOS maybe? I remember CoreOS Docker versions lagging behind a little bit?

piotrb:
"repeatable environment for testing" .. eh .. I'd like that to be a semi stable CI cluster since its kinda our main one ;)

toolmantim:
Sorry, I meant you seem to be able to reproduce the problem!

lox:
Apologies @piotrb, we are working on a fix!

PR #251, opened 2017-02-21
Download git-credentials files from secrets bucket

lox:
This checks the secrets bucket for a `git-credentials` file to support git over https. 

The format of the file (from https://git-scm.com/docs/git-credential-store#_storage_format) is:

```
https://user:pass@example.com
```


lox:
> Is this the default way of doing things? Because from what I can tell, the Downloading git-credentials files header will always be output?

It will be, but it's consistent with the other sections. They output that header regardless of whether they actually do anything.

lox:
IMO this is good to go @toolmantim.

Issue #256, opened 2017-03-07
Builds started before docker is running


deppy:
A few times per day we see tests fail because the buildkite agent has registered and taken a job before docker is started up.  Looking into this, the boot order does indeed first load the buildkite agents and then load docker daemon.

I'll submit a PR to change this ordering so that docker gets started before the buildkite agents.

Logs from a failed run:
```
$ docker-compose -f docker-compose.yml -p buildkite4b994829764f4e23ae2d356ff0803d3b run --no-deps web prospector
Creating network "buildkite4b994829764f4e23ae2d356ff0803d3b_default" with the default driver
ERROR: Couldn't connect to Docker daemon at http+docker://localunixsocket - is it running?
 
If it's at a non-standard location, specify the URL with the DOCKER_HOST environment variable.
```

lox:
Hi, sorry to hear you are having this problem. I wish it was as simple as docker not starting, but from what I can tell this happens randomly. It's currently being tracked in #236 and is very difficult to reproduce. 

If you are able, seeing the docker logs for the instance that this happened on would be very helpful.

lox:
Could I trouble you to downgrade to 2.0.0-rc2 @deppy? https://s3.amazonaws.com/buildkite-aws-stack/master/7ddb893abf326592a8a6bf7053891d8af991544f.aws-stack.json

I'm in the #aws-stack branch of the Buildkite Slack if you want to chat.

deppy:
At least the last 2 times that I saw it happen this morning it was because 2 new instances had just started.  I've also seen the more random ones in #236 and at least those seem less frequent on v2.0.0-rc5.

It very well could be that those 2 times were related to the issues in #236 but it just seems less likely since those 2 agents had just come up and the buildkite-agent does in fact start before docker :)

I'll talk to my team if we can downgrade to rc2, we had been on that one and I forget already why but we upgraded to rc5 due to some problems with the S3 encryption of secrets on rc2.

lox:
This is in master, tagged as 2.0.0-rc6. Let me know if you keep seeing the issue @deppy, at which point it's probably #236.

PR #257, opened 2017-03-07
Make sure that docker starts before launching the buildkite-agent

deppy:
This changes order of services to fix issue #256 so that docker starts up before buildkite-agent.


lox:
Thanks! Not sure why I haven't done this before, but fingers crossed it improves things. I will cut a new release candidate.

PR #258, opened 2017-03-08
Include outputs template when building stack

gugahoi:
It seems the outputs from `template/outputs.yml` were never included when generating the aws-stack.json. This addresses the issue.


lox:
Doh!

gugahoi:
That was quick, thanks!

lox:
We don't mess around :) Thanks for finding this issue and the PR!

Issue #260, opened 2017-03-21
Updating the stack always puts us back to the MinSize of the ASG


deppy:
The CloudFormation template looks like it was going to allow people to configure the DesiredCapacity (parameter added), but then was hard coded to use the MinSize.  Our MinSize is 0, so any upgrades during the middle of the work day always sets the ASG back to 0 nodes until the CloudWatch alarms slowly have it scale up.  

I'll submit a PR to fix this as it should be configurable by parameter to prevent these cases.

gugahoi:
if you have a desired capacity that is bigger than the minimum size, when will it ever go down to minimum?

gugahoi:
Why would it matter so much? Just let it scale up when workload is needed. Whats the point of a desired capacity in this case other than the first time the ASG starts?

deppy:
Desired capacity is how many instances it should currently have, which is why it matters so much.  If I have `MinSize: 0`, `MaxSize: 20` and the stack currently has `DesiredCapcity: 18` and I update the stack to make any kind of change, the current setting would drop DesiredCapacity back to 0, meaning that I have no agents.  With our current scaling policies, it would take about an hour to go back to the DesiredCapacity of 18.  This means a lot of unhappy developers for an hour (or me manually going in and setting DesiredCapcity on the ASG to 18, when I should be able to do it through the stack).

lox:
I think this is probably being driven by https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/buildkite-elastic.yml#L455-L467:

```
    UpdatePolicy:
      AutoScalingRollingUpdate:
        MinInstancesInService: 0
        MaxBatchSize: 5
        # On rollback this might have to wait for an agent to finish
        # a job. The agent's init script waits 30 minutes, so we wait
        # a maximum of 40 minutes to be safe.
        PauseTime: PT40M
        WaitOnResourceSignals: true
        # Ignore scaling alarms during update
        SuspendProcesses:
          - AlarmNotification
          - ScheduledActions
```

I need to do some more research on this, I suspect we are missing something here.

lox:
This was closed by #261, with some follow-up work pending in #270.

PR #261, opened 2017-03-21
Allow DesiredCapacity to be configured via parameter

deppy:
Allow the DesiredCapacity to be configured via parameter to prevent always scaling down to the MinSize during stack updates. (Issue #260)

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -39,7 +39,7 @@ Metadata:
       - Label:
           default: Auto-scaling Configuration
         Parameters:
-        - DesiredSize
+        - DesiredCapacity
         - MinSize
         - MaxSize
         - ScaleUpAdjustment
@@ -138,6 +138,12 @@ Parameters:
     Description: Spot bid price to use for the instances. 0 means normal (non-spot) instances
     Type: String
     Default: 0
+
+  DesiredCapacity:
+    Description: Desired number of instances to start with
+    Type: Number
+    Default: 0
+    MinValue: 0
 
   MaxSize:
     Description: Maximum number of instances
@@ -420,7 +426,7 @@ Resources:
         $(Subnets)
       ]
       LaunchConfigurationName: $(AgentLaunchConfiguration)
-      DesiredCapacity: $(MinSize)
+      DesiredCapacity: $(DesiredCapacity)
       MinSize: $(MinSize)
       MaxSize: $(MaxSize)
       MetricsCollection:


toolmantim:
@deppy I just found a problem/side-effect of this new addition. If you set `MinSize` to a non-zero value (say, 1), but don't forget to set `DesiredCapacity` then you get a stack creation failure after it goes and creates a bunch of resources. And because it's not a validation error, you have to go through the whole stack creation process again.

I've poured over the CloudFormation documentation to try to do:

```
DesiredCapacity: $(Maximum of (DesiredCapacity || MinSize)
```

or:

```
DesiredCapacity: $(DesiredCapacity < MinSize ? MinSize : DesiredCapacity)
```

or just to add support for validating that `MinSize >= DesiredCapacity`

… but I don't think any of these are possible.

Given we can't validate it, this makes the stack setup process much less easy-to-use :(

Is there another solution we could switch to, to have the same affect? I think this new parameter could add a lot of confusion.

deppy:
Fair Point.  DesiredCapacity is an optional field.  I believe the behavior would be that the initial value would be set to MinSize, and then on upgrade it shouldn't modify it (ie. solving the problem that I was running into where an update would downscale to MinValue).  I've put in Support Request to amazon to see what the default value would be (I have no time to test that at the moment to find out the answer).

If it does indeed behave that way, I'd be up for just removing DesiredCapacity from the template since it would revert back to the previous behavior of initially being the MinSize, while fixing the downscale issue we were facing.  

Will let you know when I get feedback from AWS.

toolmantim:
Oh yeah, I hadn't considered removing it altogether! That could work. Thanks for reaching out to AWS to figure it out.

SomeoneWeird:
@deppy Did you hear back from AWS?

lox:
So shall we just remove it? @SomeoneWeird points out it's actually a breaking change as it stands.

deppy:
I had a few messages back and forth with AWS.  They wouldn't directly answer my questions when I gave them multiple scenarios but this is what I've gathered so far....

If you don't include DesiredCapacity, on initial stack creation it will use the MinSize value.  On stack update it will leave it as is the current value for DesiredCapacity unless the stack update requires a replacement of the ASG, at which point it will revert back to the MinSize (which was the case that was breaking for me, because we often change the MaxSize).

As for the other part of verifying that `MinSize <= DesiredCapcity >= MaxSize`, they basically said there is no way to do that.  I've given them a feature request for this, that would require that they allow you to use a parameter as a value for another parameters MinValue and MaxValue.

Basically there's no way to validate that the value someone entered was in the correct range and the only way to not revert back to the MinSize for DesiredCapacity is do declare it when replacing the ASG in the stack update.  So feel free to revert this and I can just keep handling it the way I have been, which is manually updating the DesiredCapacity after a stack update.  Less than ideal, but it will at least prevent others from putting invalid values and breaking their stack.

SomeoneWeird:
```
If you don't include DesiredCapacity, on initial stack creation it will use the MinSize value.
```

This doesn't seem to be the case, currently. The stack exits with an error if `DesiredCapacity` is not defined.

deppy:
@SomeoneWeird By "If you don't include DesiredCapacity", I mean that if you completely pull it out of the template.  The way it exists now is because it doesn't have a default because it must it must bet MinSize <= DesiredCapacity >= MaxSize.

toolmantim:
What if we set `DesiredCapacity` to `MaxSize`? That way when you update or create the stack it scales out, but then scales in afterwards if there's no jobs. And stack updates will be much smoother than reverting to the old behavior.

deppy:
@toolmantim That works for me :-)

lox:
https://github.com/buildkite/elastic-ci-stack-for-aws/pull/270 changes things so that if you leave it at zero it will be the empty value. I think that captures the above?

PR #266, opened 2017-04-06
Start docker manually to avoid corrupting init

lox:
A current theory for issue #236 is that the restart is happening during dockers initialization, which causes corruption.

This moves the first docker start from packer to when the instance boots, as well as ensuring it happens before the agents are started.


lox:
@toolmantim IMO this is worth deploying to master for testing.

toolmantim:
@lox agreed! Let's give it a go.

PR #269, opened 2017-04-08
Support spot fleets

lox:
Very early thinking for spot fleets.

At this stage I'm thinking we will provide a built-in list of instances and weights and the only parameter that will be customisable is the spot price param we currently have. 

Because of how Spot Fleet scaling works, we need to implement scaling and launch configurations both in the old ASGs and the new Spot Fleets. It's a pain, but should be largely invisible to the user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,8 @@ STACK_NAME ?= buildkite
 SHELL=/bin/bash -o pipefail
 TEMPLATES=templates/description.yml \
   templates/buildkite-elastic.yml \
-  templates/autoscale.yml \
+  templates/asg.yml \
+  templates/spot-fleet.yml \
   templates/vpc.yml \
   templates/metrics.yml \
   templates/outputs.yml
--- a/templates/asg.yml
+++ b/templates/asg.yml
@@ -2,7 +2,7 @@
 Resources:
   AgentScaleUpPolicy:
     Type: AWS::AutoScaling::ScalingPolicy
-    Condition: UseAutoscaling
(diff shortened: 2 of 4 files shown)


chendo:
Would this also provide the ability to specify the duration requirements? More expensive than normal spot instances, but having it for an hour/six hours should increase build reliability.

lox:
@chendo defined duration spot instances are not overlapping with spot fleets unfortunately.

shuber:
@lox Would love to get this back on your radar to prioritize if possible 🤞

tspacek:
We just got hit by this: ap-southeast-2b spot prices spiked, and even though our ASG was in 2c as well it blocked on the failing 2b request.

arturopie:
What's the status of this PR @lox?

lox:
Unfortunately still waiting for us to have some time to work on it @arturopie, I'm hoping soon.

screamish:
How does the launch of EC2 Fleet https://aws.amazon.com/blogs/aws/ec2-fleet-manage-thousands-of-on-demand-and-spot-instances-with-one-request/ change your thinking re: this?

lox:
Closing this for the time being.

donnoman:
@lox What's the status for this, we are having problems with spot availability with a newer instance type, we want to back the spot requests up with an on-demand option which should be available from ec2 fleet.

lox:
We are still figuring out what direction to head in, I'll post some thoughts in https://forum.buildkite.community/c/elastic-ci-stack-for-aws

PR #270, opened 2017-04-12
Allow DesiredCapacity to be unset

lox:
This addresses the issue that if a value of zero is set (the default) and the MinSize is greater than that then a validation error occurs.

See #261 for more discussion.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -140,7 +140,7 @@ Parameters:
     Default: 0
 
   DesiredCapacity:
-    Description: Desired number of instances to start with
+    Description: Desired number of instances to start with, or zero for unset
     Type: Number
     Default: 0
     MinValue: 0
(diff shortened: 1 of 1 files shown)


toolmantim:
Oh nice! It might be nice to clarify the documentation on this.

Some questions I had, reading it like a first-time-user:

* Do I need to set it?
* When/why would I need to set it?
* When do I _not_ need to set it?

I'm happy to fix the docs, if we can help answer those questions

toolmantim:
Perhaps it should be called InitialDesiredCount and say "The initial number of agents to start when the stack is created or updated. The default value of 0 will leave the desired number unset, meaning it will…"

Having written that, can we just leave it as `NoValue`? That should give us 0 for a new stack, and maintain the current value for a stack update?

I think that was the intention of the talk in #261, but it was assumed you couldn't set it as `NoValue`.

lox:
I'm not sure I follow @toolmantim, as in make it not configurable and always `NoValue`? I'd frankly be fine with that.

toolmantim:
Yep, that's exactly what I'm suggesting. I think that'd be perfect.

lox:
Is #271 what you meant @toolmantim?

toolmantim:
That's it! And it looks like #271 passed! 👏🏼 Let's merge and test.

Issue #107, opened 2016-08-02
Suggestion for smoother autoscaling


mtibben:
The current rules seem to give a hard scale up to max, and a hard scale down to min when idle. This might fit a typical workday in one office, but because we have a few developers in different timezones using the elastic stack, this doesn't really fit because we have a lot of unused resources when this hard scale up occurs in the middle of the night. I'd prefer a more reactive stack that could scale to fit the demand at the time.

Typically the way I've used autoscaling is to peg average CPU utilisation, adding instances at 60%, and removing instances at 40%. Agents are a little different as we've literally got a queue being processed, which means our stack stats are very spiky. Builds that come in spawn multiple jobs, so the `RunningJobsCount` and `UnfinishedJobs` bounces around quite a bit. 

So the metric I _think_ will give us better results would be a "agents busy" percentage averaged over 5 minutes or so. Then we could trigger a more gentle scale up maybe at 70%, and a gentle scale down at maybe 20%. Hopefully that would give the elastic stack room to bounce around between 20-70%, but will nudge the cluster size in the right direction once utilisation increases or decreases past those points. And we would keep the existing scale up trigger if the `ScheduledJobsCount` rises above 0 to make sure we scale in the first place and on demand.

Thoughts @toolmantim @lox ?

toolmantim:
How hard/soft the scales are is configurable though yes? I find it hard to find a good number with a mix of pipelines when some need a large number of agents per build, and others need less.

So that's one other thing to consider.

There's also stepped scaling as well? Which could do much more context-sensitive scaling depending on the queue size. But nothing's been yet done to spike that.

toolmantim:
Relates to #93

lox:
Is this addressed by https://github.com/buildkite/elastic-ci-stack-for-aws/pull/151?

lox:
Ignore that, I don't think it is. How has this played out over time @mtibben?

mtibben:
The current auto-scaling rules are wasteful for our use-case. I haven't put any time into trying to resolve this recently. This chart is total agents vs busy agents over the course of day

![screen shot 2016-11-11 at 1 58 40 pm](https://cloud.githubusercontent.com/assets/980499/20202914/330ad81a-a817-11e6-80c6-e6cb0a1f6be1.png)

toolmantim:
@mtibben have you tried out the latest master? It's changed the rules since this issue was opened, and I think it might be better for you now.

[...]

PR #274, opened 2017-04-19
Add a ScaleDownPeriod param to control scale-down timeframe

lox:
Based on discussion in slack, this adds the ability to tune the timeframe around which scale down occurs.

It also slightly increases both the scale up and down cooldown timers (60 seconds to 300 seconds), because I've been meaning to for ages. The current stack will scale up again before the new instance has even finished booting.

### DIFF
--- a/templates/autoscale.yml
+++ b/templates/autoscale.yml
@@ -6,7 +6,7 @@ Resources:
     Properties:
       AdjustmentType: ChangeInCapacity
       AutoScalingGroupName: $(AgentAutoScaleGroup)
-      Cooldown: 120
+      Cooldown: 300
       ScalingAdjustment : $(ScaleUpAdjustment)
 
   AgentScaleDownPolicy:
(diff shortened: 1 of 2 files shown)


lox:
@mtibben this might also help the scale up issues you were talking about.

gugahoi:
Does this mean scaling up will be less aggressive? It used to work quite well for us using `m3.mediums`.

mtibben:
the issues are more with scale down @lox. As in it never gets a chance to scale down because there is always something active in the queue. That's why I suggested in https://github.com/buildkite/elastic-ci-stack-for-aws/issues/107 that the scale up and down rules be based on the percentage of active agents

mtibben:
Here's an example again of Total vs Busy agents. You can see it only scales down when `busy == 0`, when I think it should start scaling down when `busy/total == 0.2` or similar

screen shot 2017-04-24 at 3 28 36 pm

lox:
For what it's worth, that graph beautifully shows the overly aggressive scale ups too :)

lox:
I'd love to see how it looks after we are scaling up properly and then look at tweaking the scale down @mtibben.

lox:
@gugahoi it just means that scaling up will wait for the last batch of instances to spin up before firing again. You can make it more aggressive by increasing `ScaleUpAdjustment`.

lox:
I'm merging this, would love to get you to test the latest master once it's built @mtibben.

PR #276, opened 2017-04-25
Default the agent to beta

lox:
Seeing as this is the advice I tend to give in the slack channel anyway, default the agent to beta. :rocket:

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -61,7 +61,7 @@ Parameters:
       - stable
       - beta
       - edge
-    Default: "stable"
+    Default: "beta"
 
   BuildkiteAgentToken:
     Description: Buildkite agent token


SomeoneWeird:
👌

toolmantim:
I wonder if instead of these release channels, we should switch to "2.x" and "3.x" as the parameters? Because taking a step back, I think that's probably preferable for declaring dependencies of hooks, pipelines, etc. Same for people using our Docker image.

gugahoi:
makes more sense to go to a specific version because when selecting `beta` in cloudformation I have no idea what version of the agent thats going to be. Also why not just release v3 stable 😝 ?

toolmantim:
Also in that world, the elastic stack can always drop support for specific versions of Buildkite Agent too.

lox:
I agree with all of the above, but that is a breaking change. I'm tempted to ship this in the 2.x series of the stack and then look to make the above change for v3.x?

PR #281, opened 2017-05-15
Move hooks to plugins

lox:
I expect this to be controversial, but I'd like to begin moving the logic that exists in the hooks in this repo into plugins so that people can mix and match.

To "always use" the secrets and ecr plugin without having to declare them in your pipeline (for backwards compat and sanity), the stack still has hooks that defer to the plugins.

This also makes it easy to disable certain features you don't want with env vars.


ozbillwang:
Not get it. Do we support plugins in v2.x?

lox:
@ozbillwang technically, no, but this manually calls the plugin hooks in the relevant spots, so they work.

PR #284, opened 2017-06-01
Add PutObject permission for managed secrets bucket

lox:
This fixes a bug where the permissions were incorrect for managed secrets bucket.


toolmantim:
This seems to have broken updating of existing stacks?

boo

lox:
Hrm, @toolmantim, it looks like it failed because Subnet1 changed? I can't think of anything I have done that would impact that.

toolmantim:
Hrmm… perhaps this bug is because of a previous change then?

Issue #286, opened 2017-06-02
Windows support


lox:
Would be great to get support for Windows AMI's to support Windows builds. 

This would require:

 - Adding a parameter for switching between windows / linux builds
 - Build a Windows AMI in parallel to the linux one (plus all the scripts)
 - Port lifecycle monitoring scripts to windows or golang

ozbillwang:
Thanks, that will be better to have one CFN template to cover both windows and linux systems. 

I need autoscaling, schedule scaling, spot instances or not, cloudwatch logs/alarms, etc.

lox:
It would be quite a big piece of work @ozbillwang!

ozbillwang:
I can contribute my time if required.

ozbillwang:
Another feature is, can we support Docker in this windows ami?

lox:
What version of windows would you expect?

lox:
https://aws.amazon.com/blogs/aws/run-windows-server-2016-on-amazon-ec2/ seems like a pretty good bet.

ozbillwang:
Do we need any GUI in windows? Can we have `Server Core` only?

ozbillwang:
And how about to install `docker for windows` as well? Otherwise, we can only pull/run windows containers.

lox:
Basically it should be at parity with the linux stuff. I'd be pretty tempted to try and put Ubuntu for Windows on it so that the bash scripts don't need to be ported.

ozbillwang:
@lox 

Not sure if you are working on this. Start preparing the `packer` template for windows. Hope we don't work on the same case.

I made a windows template (https://github.com/BWITS/elastic-ci-stack-for-aws/blob/windows_support/packer/buildkite-ami.json) in my fork (https://github.com/BWITS/elastic-ci-stack-for-aws/commit/1dfcd9fe6ae6f9c5d3a68d6be3cdbb9a9c678eb4)  as stage 1

- build from AMI image `Windows_Server-2016-English-Full-Containers-2017.05.10`
- install chocolatey
- use chocolatey to install 7zip, git, later we can add more applications. If you can package buildkite via chocolatey, that will be simpler as well. 

Please take a look.

ozbillwang:
@lox 

What I will do:

1. Set packer template for windows AMI
2. Set AWS CFN templates base on windows AMI

I need your decision on how to organize the files to for linux/windows. We should use same CFN templates as below.  

```
autoscale.yml
[...]

PR #289, opened 2017-06-05
[WIP] Windows support - #286

ozbillwang:
- build from AMI image Windows_Server-2016-English-Full-Containers-2017.05.10
- install chocolatey
- use chocolatey to install 7zip, git, later we can add more applications. If you can package buildkite via chocolatey, that will be simpler as well.

This PR is still in WIP, please don't merge it.

Let me know how to arrange the templates for linux and windows. Currently I didn't update `Makefile` and use the same template name to easily build the new image.

### DIFF
--- a/packer/buildkite-ami.json
+++ b/packer/buildkite-ami.json
@@ -14,6 +14,13 @@
     }
   ],
   "provisioners": [
+      {
+           "type": "powershell",
+           "inline": [
+              "iex ((new-object net.webclient).DownloadString('https://chocolatey.org/install.ps1'))",
+              "choco install -y 7zip git"
+           ]
+       },
        {
            "type": "powershell",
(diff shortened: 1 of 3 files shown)


lox:
Awesome start @ozbillwang! Sorry for the slow reply, I've been pondering how to handle this. I'm concerned about getting into a situation where we have parallel version of all the scripts in this repo. Pretty much the entirety of the code in here is bash scripts, and at present it assumes *nix support. 

I think the only way we could consider something like this would be if we could use something like https://blogs.windows.com/buildingapps/2016/03/30/run-bash-on-ubuntu-on-windows/ and make minor changes to the bash scripts.

ozbillwang:
sure, will raise or re-open ticket if I can enable bash on Windows 2016.

PR #294, opened 2017-06-12
Improve documentation for uploading env files to s3

### LINKED ISSUE #292: Example for setting secrets in the secret bucket
I was wondering if there's a gist, documentation or an example for the S3 object to set environment variables. Is it a key/value pair with the object name being the key name or just a `ENV_NAME=ENV_VALUE` in a text file? Was hoping there'd be something to reference rather than figuring it out via trial and error.

Thanks

lox:
Closes #292

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ aws cloudformation create-stack \
 
 The stack will have created an S3 bucket for you (or used the one you provided as the `SecretsBucket` parameter). This will be where the agent will fetch your SSH private keys for source control, and environment hooks to provide other secrets to your builds.
 
-The following s3 objects are downloaded and processes:
+The following s3 objects are downloaded and processed:
 
 * `/env` - An [agent environment hook](https://buildkite.com/docs/agent/hooks)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -91,7 +91,17 @@ pbcopy < id_rsa_buildkite.pub # paste this into your github deploy key
 aws s3 cp --acl private --sse aws:kms id_rsa_buildkite "s3://${SecretsBucket}/private_ssh_key" 
 ```
 
-If you really want to disable KMS encryption, you can set `BUILDKITE_USE_KMS=false`.
+If you want to set secrets that your build can access, create a file that sets environment variables and upload it:
+
+```bash
+echo "export MY_ENV_VAR=something secret" > myenv
```
andrewoh531: Does `myenv` need to be named `env`?

lox: It doesn't :) It can be called anything you want locally.

andrewoh531: You're right :) thanks

PR #295, opened 2017-06-12
TMPDIR is unbound in the environment hook

anoldguy:
Let's set it like [the plugin does](https://github.com/buildkite-plugins/s3-secrets-buildkite-plugin/blob/master/hooks/environment#L10).


lox:
Many thanks @anoldguy! Now to figure out why the tests didn't catch this.

anoldguy:
Wow, merged already! I can't wait to let it rip tomorrow! ❤️❤️❤️

PR #296, opened 2017-06-17
Add some basic bats unit tests

lox:
This was largely an exercise for me to understand the fix-buildkite-agent-builds-permissions stuff.

PR #298, opened 2017-06-17
WIP: Tidy up the fix permissions script

lox:
This allows it to be more easily used other places

### DIFF
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -8,6 +8,9 @@ source ~/cfn-env
 echo "~~~ :llama: Setting up elastic stack environment ($BUILDKITE_STACK_VERSION)"
 cat ~/cfn-env
 
+echo "Fixing permissions for '${BUILDKITE_AGENT_NAME}'..."
+sudo --preserve-env /usr/bin/fix-buildkite-agent-builds-permissions
+
 echo "Configuring built-in plugins"
 
 [[ ! ${SECRETS_PLUGIN_ENABLED:-true} =~ (on|1|true) ]] && PLUGINS_ENABLED=${PLUGINS_ENABLED/secrets/}
@@ -41,7 +44,3 @@ if [[ -n "${BUILDKITE_SECRETS_BUCKET:-}" &&  "${SECRETS_PLUGIN_ENABLED:-}" == "1
   # shellcheck source=/dev/null
   source /usr/local/buildkite-aws-stack/plugins/secrets/hooks/environment
 fi
-
-
-echo "Fixing permissions for '${BUILDKITE_AGENT_NAME}'..."
-sudo --preserve-env /usr/bin/fix-buildkite-agent-builds-permissions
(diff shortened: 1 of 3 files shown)


packer/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions
```diff
@@ -1,81 +1,59 @@
...
+  /usr/bin/awk -F "/" '{print $1}' <<< "${BUILDKITE_BUILD_CHECKOUT_PATH#${BUILDKITE_BUILD_PATH}/}"
+}
+
+# Returns => `/var/lib/buildkite-agent/builds/my-agent-1`
+buildkite_build_agent_checkout_path() {
+  echo "${BUILDKITE_BUILD_PATH}/$(buildkite_build_agent_name_stub)"
+}
+
+# Allow a custom chown passed in for testing
+chown_bin=${1:-/bin/chown}
```
sj26: This seems like a security hole. In any build I can now run:

```
/usr/bin/fix-buildkite-agent-builds-permissions /path/to/my-pwn-script
```

packer/conf/buildkite-agent/hooks/environment
```diff
@@ -8,6 +8,9 @@ source ~/cfn-env
 echo "~~~ :llama: Setting up elastic stack environment ($BUILDKITE_STACK_VERSION)"
 cat ~/cfn-env
 
+echo "Fixing permissions for '${BUILDKITE_AGENT_NAME}'..."
+sudo -E /usr/bin/fix-buildkite-agent-builds-permissions
```
sj26: Passing in all env means I can override PATH and things for nefearious purposes. I think the interface should be kept very minimal and require passing in the variables as arguments.

sj26: I think this also needs to remain in `environment` (i.e. not move to pre-checkout like #296) because otherwise I could do things like insert a plugin hook which changes BUILDKITE_CHECKOUT_PATH to something targeting /etc/passwd so it gets chowned and I can escalate out of my build.

packer/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions
```diff
@@ -1,81 +1,59 @@
...
 
-# We also need to make sure it's not empty so we don't collide with other agents:
-if [[ -z "${AGENT_BUILDS_NAME}" ]]; then
-	exit 3
-fi
+# This script can be run in a Buildkite environment hook to reset permissions for an agents
+# builds directory to avoid errors.
 
-# We know the builds path:
-BUILDS_PATH="/var/lib/buildkite-agent/builds"
```
sj26: The reason to put this in here was to absolutely make sure that there was no way for this script to be used anywhere but inside this known part of the filesystem.

sj26:
I love the idea of making this modular and reusable. But because it is a script which can be sudo'd I tried to keep it with the most minimal interface and as locked down as possible to prevent exploits. I think maybe there's some middle ground. When you've got a moment, maybe tomorrow in HQ, let's have a chat? :-)

lox:
Yup, you are totally right, this was ill considered.

PR #303, opened 2017-06-23
Backport SECRETS_PLUGIN_ENABLED for disabling secrets in 2.1.x

lox:
This allows for the secrets functionality to be entirely disabled. It backports `SECRETS_PLUGIN_ENABLED` like in the forthcoming 2.2.x series. 

You can then just set this in env and it will skip the secrets code.


SomeoneWeird:
Looks good, much thanks 🎉

PR #306, opened 2017-06-30
Add shudder rather than bash terminationd

lox:
Currently we use a lovely bash script for polling AWS lifecycle status to catch scale up and scale down events and shutdown gracefully.

This works nicely, for the most part, except it tends to exceed API limits on large clusters and it doesn't support spot instance termination notifications. I started to build https://github.com/lox/lifecycled to work around this, but in the meantime https://github.com/scopely/shudder got there. 

This PR replaces `terminationd` with the shudder daemon, which calls our custom shutdown script which will gracefully shutdown the agents. 

The only major caveat is that we go back to using an SNS topic to publish lifecycle changes, so shudder will create an SQS queue per instance. These are free, and unlimited. I can't see any good reason why this is a bad idea.


sj26:
I think we only moved away from lifecycled because it was using a single SQS queue and messages weren't getting to the correct instances in time for graceful termination, a problem which got worse as instance numbers went up. But queue per instance sounds like it could work! Give it a go?

PR #309, opened 2017-07-05
Revert "Docker 17.06.0-ce, was 17.05.0-ce."

lox:
Reverts buildkite/elastic-ci-stack-for-aws#307

There are some breaking changes around ecr-login that I would prefer not to deal with yet, and the URL is broken.


lox:
Thanks anyway @pda :)

Issue #63, opened 2016-05-10
Instance should be marked as unhealthy if buildkite-agent it stopped


toolmantim:
I think you were right on this one @lox — seeing as there's no monit/upstart, if you remotely stop the buildkite-agent process then you're going to get the instance just sitting there, but without any agent running on it.

I imagine what should happen is that the instance is marked as unhealthy and then replaced as soon as that happens.

lox:
A few options: 

 * add a generic shutdown hook to the agent
 * setup a cron or some sort of process monitor baked into the AMI
 * add a process monitor feature to lifecycled that then called the shutdown API when the process died
 * add an aws-aware feature to the agent that marks and instance as unhealthy on shutdown, perhaps distinguishing between remote and local shutdown

toolmantim:
Related: buildkite/agent#111 buildkite/agent#224

lox:
Option 1 would pretty neatly solve a lot of those.

toolmantim:
What if we just changed the command that starts the agent to `buildkite-agent start; report-instance-as-unhealthy`?

gugahoi:
I know this is not directly related however I have also seen agents spin up when scalling and fail to register with buildkite. This results in the agent being up and running but not actually doing anything. Would it be sensible for the metrics stack to periodically check that an agent is registered with buildkite? In the case the agent is not registered then it should be shut down.

deppy:
I'd +1 this, but also take it a step further.  We've been seeing a number of cases where containerd restarts and then the job just continues to run and run (we've also seen cases where the job just stops output and fails, but I'm still trying to see if that's the same issue).  So I'd have it checking the health of any of the common services required to be a healthy node and terminate the instance when it doesn't meet those requirements.

keithduncan:
We now [supervise](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/linux/conf/buildkite-agent/systemd/buildkite-agent.service) buildkite-agent with systemd and have an `ExecStopPost` which terminates the instance. Hopefully this isn’t happening any more! 😄

PR #317, opened 2017-07-26
WIP: Add a instance health check script

lox:
This adds `/usr/local/bin/check-instance-health` which checks if less than 1GB of disk is free and if so marks an instance as unhealthy.

Currently this happens in the `environment` hook, but the intention is to trigger it via cron too.

This will also probably address https://github.com/buildkite/elastic-ci-stack-for-aws/issues/63.


toolmantim:
This is awesome!

I'm curious that's it an environment hook. Is that just because we haven't dealt with cron in Packer yet?

lox:
Yeah @toolmantim, perhaps cron would be better. My thinking was that this way rather than kill the instance that it would clear it's disk then and there and preserve recent cache.

lox:
Replacing this with #326

PR #325, opened 2017-08-14
Prevent unused images from filling up agent disk

patrobinson:
Using system prune works better than the previous `docker-gc` but we found after 10 days the unused images were enough to fill our 80GB disk.

This prunes unused images as well as "dangling" images (the default) older than 4 hours.

NOTE: This can have a nasty side effect that images are not available for use by the docker build cache, which results in slower build times. We have multiple ways to work around this internally, but it may not be desirable for all users.
Having the build cache around "forever" (as long as the instance is around) can be rather convenient.

### DIFF
--- a/packer/conf/docker/cron.daily/docker-gc
+++ b/packer/conf/docker/cron.daily/docker-gc
@@ -2,4 +2,4 @@
 set -euo pipefail
 
 # prune stopped containers and images
-docker system prune -f
+docker system prune -af --filter "until=4h"


lox:
Perfect, thanks :) 

Similar approach to what I was looking at in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/317.

lox:
Now that I've merged it, I wonder if 4 hours is too aggressive? Perhaps we should only prune if disk space is low?

patrobinson:
The reason I chose 4 hours is, I'd rather this expose issues straight away not randomly. i.e. every 5th or 6th day Josephine's build times out and she just retries it. But actually if she properly setup the docker cache it would be faster and less flakey.

This makes it easier to understand why this is happening.

lox:
My concern is more that it kills cache hit rates for anything that runs less frequently than 4 hours.

PR #326, opened 2017-08-16
Only run docker-disk-gc on low-disk

lox:
This riffs on #325 and supersedes #317. It adds a disk free check to the cron script that cleans up docker things so that older images/containers/volumes are only purged if disk or inodes are low. 

It also increases the frequency of this to hourly (cron.hourly). I was almost tempted to run this in the `environment` hook too, so that issues are surfaced during the build if possible.

Feels @patrobinson?


lox:
Yup, agreed @patrobinson. I've made those changes.

lox:
The other thing this does is marks the instance as unhealthy if it has low disk and an image purge won't fix it.

packer/conf/docker/cron.hourly/docker-network-gc
```diff
@@ -0,0 +1,11 @@
...
+set -euo pipefail
+
+DOCKER_PRUNE_UNTIL=${DOCKER_PRUNE_UNTIL:-4h}
+
+## ------------------------------------------
+## Prune stuff that doesn't affect cache hits
+
+docker network prune -af --filter "until=${DOCKER_PRUNE_UNTIL}"
+docker container prune --all --force --filter "until=${DOCKER_PRUNE_UNTIL}"
+docker volume prune --all --force --filter "until=${DOCKER_PRUNE_UNTIL}"
```
patrobinson: The name of this file is no longer relevant

patrobinson:
Sounds pretty sweet. I like it! 👍

PR #331, opened 2017-08-23
Fix incorrect description

meganemura:
`30 mins` is old description since there is `ScaleDownPeriod`.


toolmantim:
Good catch! Thanks @meganemura

PR #336, opened 2017-09-02
Add a dependency on the Gateway for the DefaultRoute

lox:
Fixes a bug raised by @pda where there is a race condition between the gateway creation and the default route. Something something eventual consistency. 

image

### DIFF
--- a/templates/vpc.yml
+++ b/templates/vpc.yml
@@ -58,6 +58,7 @@ Resources:
 
   Routes:
     Type: AWS::EC2::RouteTable
+    DependsOn: Gateway
     Condition: CreateVpcResources
     Properties:
       VpcId: { Ref: Vpc }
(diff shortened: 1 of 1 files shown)


templates/vpc.yml
```diff
@@ -58,6 +58,7 @@ Resources:
 
   Routes:
     Type: AWS::EC2::RouteTable
+    DependsOn: Gateway
```
pda: I have a feeling these need to be `DependsOn: GatewayAttachment` rather than `Gateway`.

The dependency on `Gateway` is implicit to the the `GatewayId: { Ref: Gateway }` below, whereas there's no way for CloudFormation to know about the dependency on `GatewayAttachment`.

The `AWS::EC2::VPCGatewayAttachment` is the thing that associates the `AWS::EC2::InternetGateway` with the `AWS::EC2::VPC`. Until the attachment is made, the Gateway is just “floating”, explaining the “[route table and internet gateway] belong to different networks” error message.

pda: *(PR inbound)*

PR #338, opened 2017-09-05
VPC: fix explicit DependsOn: GatewayAttachment

pda:
Follow-up to https://github.com/buildkite/elastic-ci-stack-for-aws/pull/336

`DependsOn` is required for implicit dependencies that aren't represented by things like `Ref: Foo`. I think the only implicit dependency here is `RouteDefault` depending on `GatewayAttachment`; the dependency on the `Gateway` itself is explicit via `GatewayId: { Ref: Gateway }`, but that `Gateway` must be attached to the `Vpc` before the `Route` can be created, so there's an implicit dependency on `GatewayAttachment`.

See similar use-case in https://github.com/awslabs/ecs-refarch-cloudformation/blob/8a10c808/infrastructure/vpc.yaml#L137-L143

### DIFF
--- a/templates/vpc.yml
+++ b/templates/vpc.yml
@@ -20,7 +20,6 @@ Resources:
 
   GatewayAttachment:
     Type: AWS::EC2::VPCGatewayAttachment
-    DependsOn: Gateway
     Condition: CreateVpcResources
     Properties:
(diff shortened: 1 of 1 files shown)


lox:
Thanks!

PR #340, opened 2017-09-08
Optimize the permissions check script to only fix the current pipeline’s build dir

toolmantim:
Currently the [fix-buildkite-agent-builds-permissions](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions) script, that's run by our [environment hook](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/conf/buildkite-agent/hooks/environment), `chown`'s the agent’s entire builds directory on every job run. This means the chown operation gets slower every time a different pipeline is checked out onto disk.

This PR changes the permissions fixing script to `chown` only the running pipeline’s checkout dir (if it exists), rather than the entire builds dir. The permissions tests have also been moved from the comment in the bottom of the file into their own executable `.bats` tests.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -8,6 +8,14 @@ steps:
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
 
+  - name: ":bash: Unit tests"
+    agents:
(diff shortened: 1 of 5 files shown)


lox:
Looks legit to me! I wish I could remember why we didn't do this in the first place.

PR #341, opened 2017-09-08
Run docker as buildkite agent with userns-remap 

lox:
This uses the `--userns-remap` feature in recent docker daemons to remap the uid of root in docker containers to the buildkite-agent users on the host.


lox:
I'm going to merge this after I put out a point release. I'm thinking of doing it as `2.3.0` @toolmantim, feels?

toolmantim:
The release plan sounds good, testing aside!

lox:
Possibly an rc1 so we can test if before it hits 2.3.0.

Vektah:
What happens to other user ids in the container? eg nobody writing doctrine cache files out to a volume mount? I think will land in following user ids so there will be stuff owned by 501+uid in that directory, so you probably still need the permission fixing scripts.

:+1: for an RC, this is a whole lot of change.

lox:
Yeah, that is a pretty epic list huh.

lox:
I might try and make this opt-in.

toolmantim:
Maybe a v3.0 milestone feature?

templates/buildkite-elastic.yml
```diff
@@ -237,6 +237,14 @@ Parameters:
       - false
     Default: "true"
 
+  EnableDockerUserNamespaceRemap:
+    Type: String
+    Description: Enables experimental feature to run docker as buildkite-agent
```
toolmantim: What if we let this feature be turned on using an ENV var?

If we do add it as another option on the stack setup screen, might need to make it clear that it's not "buildkite-agent" the binary we're taking about here in the description?

If it's a "this could be buggy, so you better know what you're doing" then perhaps we just say:

> Enables experimental Docker userns-remap support

lox: The issue is that it needs to have the system docker daemon started with the setting

toolmantim: Ah of course! 🙃

lox:
If this is an optional setting, are you ok with a 2.3.0 release @toolmantim?

PR #343, opened 2017-09-11
Namespace and rename the CloudWatch log groups

### LINKED ISSUE #342: CloudWatch Logs Not Namespaced
image

We use CloudWatch Logs for a variety of applications, not just BuildKite CI. It would be ideal if the BuildKite logs used a `/buildkite` (configurable, or default empty for backwards compatibility) Log Group prefix in order to:

- enhance readability and maintainability
- allow prefix based IAM permissions for developers

For example AWS CodeBuild uses the `/aws/codebuild` prefix for Log Groups.

toolmantim:
As reported in #342, the CloudWatch log groups are currently quite roughly named, and not namespaced to Buildkite. If you’re using the stack in an existing AWS account this can create a bit of mess.

This PR namespaces everything under `/buildkite` and renames the groups to be consistent and a little more friendly.

Closes #342

### DIFF
--- a/packer/conf/awslogs/awslogs.conf
+++ b/packer/conf/awslogs/awslogs.conf
@@ -1,9 +1,9 @@
 [general]
 state_file = /var/awslogs/state/agent-state
 
-[/buildkite/systemd]
+[/buildkite/system]
 file = /var/log/messages
-log_group_name = /buildkite/systemd
+log_group_name = /buildkite/system
 log_stream_name = {instance_id}
 datetime_format = %b %d %H:%M:%S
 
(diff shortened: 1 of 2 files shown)


lox:
@toolmantim thoughts on whether this is a backwards compatibility break?

toolmantim:
@lox I don't think so! It's just logs. In the release notes we'll just mention they have a new path, and that you're welcome to delete the following old log groups: "..."

toolmantim:
I thought about it, but I figured in support and Slack we'd probably say "Can you check the system logs?" or "Can you check the agent logs?" and then they'd have to know that internally the agent log is stored at `/var/log/elastic-stack.log` and that's the CloudWatch log group to look at.

I'd say I'm a 9/10 on the rename. Because personally, I never know which one to pick… and I'm a maintainer.

In general I think people will deal with logs long before they SSH into an instance or look at the Packer source code, so trying to be more descriptive is probably better? And if you're a maintainer, you'll know to look in the `awslogs.conf` to find the path a group maps to.

packer/conf/awslogs/awslogs.conf
```diff
@@ -1,38 +1,38 @@
 [general]
 state_file = /var/awslogs/state/agent-state
 
-[/var/log/messages]
+[/buildkite/systemd]
```
lox: /buildkite/system

packer/conf/awslogs/awslogs.conf
```diff
@@ -1,38 +1,38 @@
...
-[/var/log/cloud-init-output.log]
+[/buildkite/cloud-init-output]
 file = /var/log/cloud-init-output.log
-log_group_name = /var/log/cloud-init-output.log
+log_group_name = /buildkite/cloud-init-output
 log_stream_name = {instance_id}
 datetime_format = %Y-%m-%d %H:%M:%S,%f
 
-[/var/log/elastic-stack.log]
+[/buildkite/buildkite-agent]
```
lox: Let's call this `/buildkite/elastic-stack-init`

toolmantim:
Thanks @lox! I've made some changes, what's it look like now?

PR #351, opened 2017-10-05
Revert "Spin up new ASG and replace on update"

lox:
In commit 35a23b3295bfe21b8ce4a78fd35c0cb1da781883 we changed the stack to spin up a new ASG in parallel, which speeds up stack updates, but at the expense of not giving the old instances that still might be running builds time to finish.


lox:
Closing, as I've lost the context on this.

PR #353, opened 2017-10-10
Login as buildkite user

patrobinson:
## Context

Recently we found `/usr/local/bin` was not in `buildkite-agent` user's path. I tried adding it to the bash_profile for the user, but that didn't work because we're calling sudo and it defaults to the `secure_path`, which by default is very restrictive

## Change

Emulate user login when launching buildkite-agent so we get all the env variables sourced from `.bash_profile`

### DIFF
--- a/packer/conf/buildkite-agent/init.d/buildkite-agent
+++ b/packer/conf/buildkite-agent/init.d/buildkite-agent
@@ -37,7 +37,7 @@ case "$1" in
         echo "Already started"
     else
         echo "Starting $name"
-        sudo -u "$user" $cmd >>"$log" 2>&1 &
+        sudo -iu "$user" $cmd >>"$log" 2>&1 &
         echo $! > "$pid_file"
         if ! is_running; then
             echo "Unable to start, see log"


lox:
Might it be clearer to just do that in the `environment` hook? I'd prefer to keep the env/defaults inherited from the system environment as minimal as possible.

patrobinson:
Yeah that sounds like a logical place to put it. My only problem is I'm trying to keep drift from this stack to a minimum, so how do I add something (i.e. add `/usr/local/bin` to `$PATH`) without modifying the existing scripts? Should we source another file (i.e. `~/user-env`) where users can place their own environment paths?

lox:
Yeah, fair point. Perhaps we should just explicitly set a path that includes `/usr/local/bin` in the stack `environment` hook? 

I'm reluctant to add to much stuff to customize the host system environment as I really want to move to a "agents run in docker" approach as soon as possible.

patrobinson:
Yeah given the headache this has caused I'm inclined to just stop supporting anything not running in docker

PR #354, opened 2017-10-10
Support assuming IAM roles for job execution

lox:
This adds support passing in an IAM role in `AWS_ASSUME_IAM_ROLE` that will be assumed in the `environment` hook. The EC2 metadata service gets firewalled off at that point, so the job that runs has the permissions of the assumed role.

The motivation here is to make a secure way to give a given pipeline a set of AWS permissions.


lox:
Unfortunately this doesn't work if there are multiple agents on a host.

lox:
Closing as this was flawed unfortunately. Eventually!

PR #355, opened 2017-10-30
Run jobs in an ECS cluster

lox:
This introduces an ECS cluster and builds off the ECS optimized-AMI. The agents still run on the host, but all jobs are spawned as ECS tasks on the same host so that we can leverage the IAM Role per Task infrastructure that AWS provides for ECS.


toolmantim:
What are your feels so far?

lox:
Not good @toolmantim. The ECS AMI is really difficult to work with and ECS is really difficult to run ephemeral tasks with streaming output. I think I'm going to give up and do per-job IAM roles some other way.

toolmantim:
Ah, cool cool. Good to leave some notes here if anyone ponders about going down the same path in the future!

mbj:
Also I found ECS (the AWS engine) to be really slow in scheduling tasks. It can take several seconds till an eligible node gets its task dispatched. For projects with lots of concurrent tasks this can be quite an overhead.

gugahoi:
Have you considered other cluster orchestration solutions?

PR #356, opened 2017-11-17
Enable, behind a parameter, pre-emptive scaling

patrobinson:
## Context

In some scenarios we like to scale up pre-emptively, to always have a spare agent available no matter what. This ensures we are never waiting on an agent to boot (2-3minutes) while the site is down and we're trying to push through a fix.

## Changes

Create two new alarms, enabled behind a parameter, that scale up if there is less than one idle agent or down if there is more than n where n is the number of Agents Per Instance.

We also disable the normal alarms if this flag is set.

### DIFF
--- a/templates/autoscale.yml
+++ b/templates/autoscale.yml
@@ -39,13 +39,13 @@ Resources:
     Type: AWS::CloudWatch::Alarm
     Condition: PreemptiveScaling
     Properties:
-      AlarmDescription: Scale-up if IdleAgentCount > 1 for N minute
+      AlarmDescription: Scale-up if IdleAgentCount > X for Y minute
       MetricName: IdleAgentCount
       Namespace: Buildkite
       Statistic: Maximum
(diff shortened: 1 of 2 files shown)


lox:
I wonder if we could change the auto-scale parameters to be something like `MinimumAgentsAvailable` and then just have the one parameter?

patrobinson:
Pasting comments from slack here (since it won't keep history) https://buildkitechat.slack.com/archives/C19BG77H9/p1516167037000239

> Having the down scaling tied to active jobs == 0 means one developer working late/remotely could keep a large number of idle agents around by just repushing branches every 10 minutes or so

IMO I think the IdleAgentCount is a better metric to scale off than UnfinishedBuilds, since it tracks better if the agents are "busy" or not. If you're not using Preemptive scaling though this may not be the best metric to use.

lox:
I'm going to close this until we loop back on autoscaling metrics.

PR #359, opened 2017-12-14
Add a BuildkiteAgentTags parameter for extra agent tags

### LINKED ISSUE #137: Add custom agent parameter customisation
Love the set up. One thing I was hoping to do was add some extra parameters to the cloud formation template to pass though to the individual buildkite agents, for consistency with our existing agents.

eg. we user an `os=osx` and `os=linux` agent parameter.

We could deduce this from the existing parameters you create for us, but it'd be great to tweak them as well.

I've tweaked the generated CF JSON to get this work, but I'm not sure where to edit the original templates to get this working. I assume [around here in the buildkite-elastic.yml](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/buildkite-elastic.yml#L400)

I'm happy to have a crack if there's an easy way to test I haven't broken anything... on condition that I get to add _devops_ to my LinkedIn ;)

lox:
Closes #137.

### DIFF
--- a/packer/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/conf/bin/bk-install-elastic-stack.sh
@@ -31,9 +31,9 @@ region = $AWS_REGION
 EOF
 
 if [[ "${DOCKER_USERNS_REMAP:-false}" == "true" ]] ; then
-  echo "Enabling docker userns-remap"
-  cp /etc/sysconfig/docker.userns-remap /etc/sysconfig/docker
-  service docker restart
+	echo "Enabling docker userns-remap"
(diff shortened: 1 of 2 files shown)


sgleadow:
Turns out I originally raised #137, forgot about that, but still something I would use with our current set up!

This looks like it enables a single string field. If you have multiple tags, are they just comma delimited? eg.

`os=linux, team=stew`

lox:
Yup, this implementation allows for multiple tags split by a string. I will make that clearer.

gugahoi:
s/string/comma ?

Issue #311, opened 2017-07-10
Availability zones limited to 2


tspacek:
If I specify three AZs, it only uses two of them.

Steps:
- Create a new stack
- Specify `ap-southeast-2a, ap-southeast-2b, ap-southeast-2c` as the AZs. Leave VPCs, Subnets etc blank.

Expected:
- The auto scaler's AZs to be `ap-southeast-2a, ap-southeast-2b, ap-southeast-2c`

What happens:
- The auto scaler's AZs are `ap-southeast-2a, ap-southeast-2b`

It looks like the Cloudformation template only creates two subnets.

N. Virginia has 5 AZs

tspacek:
Cross ref: https://github.com/buildkite/elastic-ci-stack-for-aws/pull/157

I think a documentation change would do, to make clear in the Cloudformation params that it only takes two.

lox:
Sorry for the late response! Yup, I think a docs change is the best way to move forward.

lox:
This is fixed in #360.

PR #360, opened 2017-12-17
VPC rewrite - Private, and all available subnets

lox:
As part of spotfleets, I want to have access to all possible availability zones. Given the number of them differs across regions, this requires some cloudformation contortions. 

I wrote this as part of a spike on an ECS/Spotfleet stack. I'm planning to bring over some significant pieces from it which will make up v3 of this stack. 

There is now a `SubnetConfiguration` option that let's you choose either the cheaper public subnet options, or private subnets with redundant NAT instances. There is a `SSHBastionHost` parameter for launching an SSH Bastion for accessing the private subnets. 

The other change is that `AvailabilityZones` is a multi-select that you choose the AZ's you'd like to use. This makes it easier to only use specific AZ's and also supports Classic EC2 accounts where some AZ's aren't available but are still listed. 

Related: #202 #311 #157


lox:
Might be relevant to your interests, and would love feedback @holmesjr @tspacek @pda @Vektah @gugahoi @SomeoneWeird

holmesjr:
I'd suggest you do need multiple public subnets, because if the one AZ goes down with your NAT gateway in it, the agents in the other AZs lose their connection to the internet.

lox:
Yeah, good point @holmesjr. My main concern was cost, it's pretty painful having a default cost of $32.85 * subnet for the stack. It rules out using the stack for personal development in a lot of cases. 

I think probably what I will go for is to add a parameter like `UsePrivateVPCSubnets` with a note that it incurs a cost. Bummer, I hate configurable complexity :( 

It's infuriating that NAT gateways don't come in smaller sizes, we could run a DIY nat gateway for $5/month.

toolmantim:
Neat! What's the process for doing this: 

> This will effectively make SSH'ing into instances hard, but if you really wanted to you create a bastion SSH host in the public subnet.

I think a lot of beginner peeps might need help accomplishing that when debugging build problems.

SomeoneWeird:
Instead of using AWS's NAT gateways, can we run our own NAT instances (which could second as bastion hosts?)

lox:
We could add a `RunSSHBastionHost` option pretty easily @toolmantim.

lox:
> Instead of using AWS's NAT gateways, can we run our own NAT instances (which could second as bastion hosts?)

Yeah, it's tempting, but I reckon in the end that public subnets will be fine for most people that want cheap and then paying for proper NAT and a dedicated bastion host makes sense for people who run production clusters.

lox:
```
  NetworkConfiguration:
    Type: String
    Description: Whether to configure your subnets as public (no extra cost) or private with redundant NAT instances ($58/month)
    AllowedValues:
      - public
      - private-with-nat
      - private-with-nat-and-ssh-bastion
    Default: "public"
```

templates/buildkite-elastic.yml
```diff
@@ -127,14 +128,23 @@ Parameters:
...
     Default: ""
 
   AvailabilityZones:
     Type: CommaDelimitedList
     Description: Optional - Comma separated list of AZs that subnets are created in (if Subnets parameter is not specified)
     Default: ""
 
+  NetworkConfiguration:
+    Type: String
+    Description: Whether to configure your subnets as public, so your instances get public IP addresses (no extra cost) or private with redundant NAT instances ($58/month)
```
SomeoneWeird: I'm presuming this $58 is probably dependant on which region you launch into?

pda:
Nice, looking good.

templates/buildkite-elastic.yml
lox: Yeah, it's probably misleadingly specific. "Pricing starts at $0.045 per NAT gateway hour plus data processing and data transfer charges."

lox:
This needs some docs, but otherwise I'm fairly happy with it.

templates/vpc.yml
```diff
@@ -1,88 +1,472 @@
...
+  RegionHas3Zones: !Or [
+    !Equals [ !Ref "AWS::Region", "us-east-2"],
+    !Equals [ !Ref "AWS::Region", "us-west-2"],
+    !Equals [ !Ref "AWS::Region", "eu-west-1"],
+    !Equals [ !Ref "AWS::Region", "eu-central-1"],
+    !Equals [ !Ref "AWS::Region", "ap-southeast-2"]
+  ]
+
+  HasNatGateway1: !And [ {Condition: CreateNatGateways }, { Condition: CreateVpcResources } ]
+  HasNatGateway2: !And [ {Condition: CreateNatGateways }, { Condition: CreateVpcResources } ]
```
patrobinson: Nice balance of cost and availability

templates/vpc.yml
```diff
@@ -1,88 +1,472 @@
...
-  Routes:
+  PublicSubnet4:
+    Type: AWS::EC2::Subnet
+    Condition: HasPublicSubnet4
+    Properties:
+      VpcId: !Ref Vpc
+      AvailabilityZone: !If [
+        UseSpecifiedAvailabilityZones,
+        !Select [ 3, !Ref AvailabilityZones ],
+        !Select [ 3, !GetAZs "" ]
```
patrobinson: Internally we discourage the use of GetAZ call. Previously us-east had an AZ removed (it was C for us, but likely different for others). Because it was in the middle of the list, the calls to GetAZ changed and wanted to re-create our VPC, Redis instances etc.

patrobinson: Nope!

lox: In our case, you can at least provide the AZ's if you need to. I'm inclined to keep supporting an empty value that causes this lookup unless there is a better option. I'll do some investigating.

lox: 20171220-0rqont2ed18sny3

With a decent description it would probably be workable.

lox: The difficulty of this approach is that it's not possible to figure out a length of a list, or to check if an offset exists, so we'd have to also have an `AvailabilityZonesCount` or similar param. Realistically, we probably need to have this anyway, otherwise folks will get errors if they provide less AZ's than their region should theoretically allow.

lox: How about this?

lox:
This is where I am at:

image

Any feels @toolmantim?

lox:
This is a bit cleaner: 

image

PR #362, opened 2018-01-02
Bump the lambda metrics timeout to 5m

lox:
We've been seeing some larger deployments having issues with metrics lamba's timing out due to the new rate limiting exponential backoff. This should at least reduce the dogpile effect whilst we figure out the root cause.

### DIFF
--- a/templates/metrics.yml
+++ b/templates/metrics.yml
@@ -50,7 +50,7 @@ Resources:
         Fn::GetAtt:
         - LambdaExecutionRole
         - Arn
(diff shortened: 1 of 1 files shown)


toolmantim:
Should we test this in a modified cloudformation JSON file for the affected person, before pushing it to master?

lox:
Yeah, quite possibly. On reflection, the issue will already be solved for folks that can upgrade their stacks, so this isn't a great fix for the problem.

PR #371, opened 2018-01-10
Make CloudFormation generate unique role names

excessivedemon:
At the moment, deploying the same stack (with the same details such as stack name) on multiple regions would cause an error: `IAM Role already exists`.

By not setting the IAM Role's RoleName, we can have CloudFormation generate a unique RoleName, allowing the same stack (with the same details such as stack name) to be deployed on multiple regions.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -325,7 +325,6 @@ Resources:
   IAMRole:
     Type: AWS::IAM::Role
     Properties:
-      RoleName: { "Fn::Sub": "${AWS::StackName}-Role" }
       ManagedPolicyArns:
         - "Fn::If":
           - UseManagedPolicyARN


lox:
Thanks for this, and sorry you ran into this bug! I believe the issue is that to be unique the rolename needs the region:

```
RoleName: { "Fn::Sub": "${AWS::StackName}-${AWS::Region}-Role" }
```

The reason we give the role a predictable name is so that you can reference it in things like Terraform or add policies to it.

excessivedemon:
@lox instead of using a predictable role name (and hard coding the values when referencing it outside the CloudFormation stack), a better approach would be to have a randomly generated name and export the value as a stack output (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-exports.html) then, anything outside CloudFormation (example: bash scripts) can reference it via AWS CLI or API (to add policies / attach the role / etc...)

Another disadvantage of setting a `RoleName` is that If you specify a value, you cannot perform updates that require replacement of the resource. This could potentially cause issues in future updates.

lox:
Yup, I hear ya. That really only helps other cloudformation stacks though, for non-cloudformation points of integration you'd need to update them when the role name changes. 

Having RoleName is something I regret. I think that we'll likely remove it in the next major version in favour of a Parameter for specifying a policy to give instances.

lox:
Closing in favour of https://github.com/buildkite/elastic-ci-stack-for-aws/pull/395.

PR #376, opened 2018-02-06
Bump docker to 18.01.0-ce

lox:


lox:
Nope, 18.01.0-ce doesn't exist. 🤷🏼‍♂️

PR #379, opened 2018-02-09
Use 172.16.0.0/12 for our VPC to avoid clashes with ClassicLink

lox:
See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/vpc-classiclink.html#classiclink-routing


SomeoneWeird:
Just fwiw this is a breaking change for us as other security groups have these subnets whitelisted.

lox:
Yeah @SomeoneWeird, that was why I stuck to the original `10.0.0.0/16` originally.

PR #390, opened 2018-03-02
Allow docker version to be switched

lox:
Currently we pin the docker and docker-compose version to the stack, which means if docker is broken, the stack is broken.

This allows for different versions to be selected and installed on boot.


lox:
The downsides to this are basically "more complexity". I might have overcomplicated the version selection too. Beyond that, it defers installation of docker to after boot, rather than during AMI build, which makes for slower boot times, but moves us closer to being able to use vanilla AMIs. 🤷🏼‍♂️

toolmantim:
May as well ditch “latest-stable” and make it explicit, because it doesn’t dynamically figure out the latest or anything. And less complexity. And just make the default choice the latest non-testing versions?

lox:
🤔 My initial thinking was that it would create upgrading difficulties if you tried to just use the previous value, but I think you are right, it would only do that if you selected a specific docker version, in which case it would be the expected behaviour.

lox:
On some reflection, I think having a single known working docker baked into each version is better. Our design goal is as few as possible knobs and dials. Closing this.

PR #394, opened 2018-03-15
Don't blow up when no plugins are enabled

haines:
If there are no plugins enabled, `$PLUGINS_ENABLED` is an empty array [here](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v2.3.5/packer/conf/bin/bk-install-elastic-stack.sh#L49), which causes the install script to fail because it is an unbound variable and `set -u` is in effect.

I've changed it to convert the empty array into an empty string.


lox:
Oooops! Thanks @haines!

haines:
No problem!

PR #395, opened 2018-03-16
Fix cfn stack instance profile name

chandanadesilva:
The current version of the stack names the IAM Role in the Instance Profile with ${AWS::StackName}-Role.

Theis results in an IAM Role with a role name like 'the-build-stack-xyzqwrpqr-Role' when the stack is generated by a process which does not fix the stack name (as happens if this template is used as a nested stack). This role name changes every time the stack is re-deployed or updated.
Having this unpredictable role name is problematic if you want the build agent to assume a role, as the 'Principal' of an IAM policy needs to be fixed.
My change allows the user the option to specify a role name if needed.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -36,6 +36,7 @@ Metadata:
         - RootVolumeSize
         - AssociatePublicIpAddress
         - ManagedPolicyARN
+        - InstanceRoleName
 
       - Label:
           default: Auto-scaling Configuration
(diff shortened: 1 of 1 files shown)


lox:
Thanks! Are you able to fix this up so that it's just your commit?

chandanadesilva:
I have changed the base to stable 2-3

lox:
`master` was correct, I'd probably just do the following:

```bash
git reset --hard origin/master
git cherry-pick d647bec
git push --force origin/fix-cfn-stack-instance-profile-name
```

chandanadesilva:
ok, done, thanks for the advice

lox:
I think this is a good change, thanks for this!

lox:
Next release! Likely today.

chandanadesilva:
Wonderful!, thanks

PR #398, opened 2018-04-03
Add support for billing tags

tduffield:
Adds two new configuration values (`BuildkiteBillingTagName` and `BuildkiteBillingTagValue`) that can be used to apply a Cost Allocation Tag to instances created by the ASG.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -246,6 +246,16 @@ Parameters:
       - false
     Default: "true"
 
+  BuildkiteBillingTagName:
+    Type: String
+    Description: The name of the Cost Allocation Tag used for billing purposes
+    Default: "aws:createdBy"
+
+  BuildkiteBillingTagValue:
+    Type: String
+    Description: The value of the Cost Allocation Tag used for billing purposes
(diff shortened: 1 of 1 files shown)


lox:
I wonder if these need to be configurable, vs just adding the `aws:createdBy` with something to indicate elastic stack.

Ideally we'd like to keep parameters to a minimum, as they make it harder for people to understand what needs configuring. 

If you wanted arbitrary tags, presumably you could just use the Additional Tags section when you are creating the cloudformation stack?

tduffield:
They need to be configurable because people may use a different tag other than `aws:createdBy` to manage cost centers. For example, my team uses `X-Application`. 

As for the Additional Tags section, those are buildkite tags and we need very specific AWS tags to use in Cost Center.

tduffield:
@lox have you had an opportunity to think on this any more?

lox:
Yup, let's merge it in. Thanks for this and thanks for your patience, have been at AWS Summit.

lox:
Hmmmm. The build is failing with `Your account is not allowed to create tag aws:createdBy`

tduffield:
@lox where do I specify additional AWS tags?

PR #399, opened 2018-04-04
Installs the latest agent version on boot vs build

lox:
Previously we installed the latest stable and beta at build time, which speeds up boot time, but means that the versions installed depend on the most recent stack update. 

This switches to installing then on instance boot, which ensures that security and stability updates make it to build instances regardless of the age of the stack. 

This also drops support for the v2 line of agents which required a custom `bootstrap.sh`.

### DIFF
--- a/packer/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/conf/bin/bk-install-elastic-stack.sh
@@ -60,10 +60,11 @@ EOF
 download_agent() {
   local release="$1"
   local channel="$2"
+  local version="${BUILDKITE_AGENT_VERSION:-latest}"
 
-  echo "Downloading buildkite-agent ${release}..."
+  echo "Downloading buildkite-agent ${release}/${version}..."
   sudo curl -Lsf -o "/usr/bin/buildkite-agent" \
-    "https://download.buildkite.com/agent/${channel}/latest/buildkite-agent-linux-amd64"
+    "https://download.buildkite.com/agent/${channel}/${version}/buildkite-agent-linux-amd64"
   sudo chmod +x "/usr/bin/buildkite-agent"
   "buildkite-agent" --version
 }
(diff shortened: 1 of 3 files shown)


toolmantim:
I like auto-updates being the default! I wonder if we need a way to specify the version, in case we ship an agent bug that affects people and they need to roll back—it’s be impossible with this setup right?

lox:
I added a `BuildkiteAgentVersion` parameter that defaults to `latest`, this would allow selecting a specific version. It's free text, which means if you use a value that doesn't exist, your stack will fail. Thoughts @toolmantim?

toolmantim:
Awesome! That'll help a lot.

Can we add an example ("do you add the v on the start?") and maybe even an [AllowedPattern](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties) with something like `^latest|v3.*$`? That'd also prevent people trying v2 agents, when we no longer have the bootstrap.sh installed.

Maybe this is worth a major version bump in the stack? Given it also deprecates support for v2 agents?

andreimc:
This would be great to be merged in for v2.3.6 cause our agents our having some issues on beta.39!!! Would be great to get a release with a new AMI 👍

lox:
We'll get something out asap @andreimc!

lox:
@toolmantim 

> Can we add an example ("do you add the v on the start?") and maybe even an AllowedPattern with something like ^latest|v3.*$? That'd also prevent people trying v2 agents, when we no longer have the bootstrap.sh installed.

Done

> Maybe this is worth a major version bump in the stack? Given it also deprecates support for v2 agents?

It doesn't just deprecate them, it removes them. The new stack requires Agent v3, so yeah I reckon a major version bump for sure.

toolmantim:
Cool cool! One last thing I noticed, should we remove `BuildkiteAgentRelease`?

lox:
Closing in favour of #400.

PR #401, opened 2018-04-10
Support building elastic stack in other regions

mikeknox:
I wanted to standup my own pipeline for elastic-ci-stack-for-aws, and to do that we needed to:
* build in regions other than us-east-1
* build using buckets other than Buildkite's

To do this:
* added or extended support for specifying parameters as pipeline environment variables
* make `buildkite-ami.json` a generated artefact


lox:
Thanks @mikeknox! Mind rebasing this onto master so that it's a clean merge?

.buildkite/steps/packer.sh
```diff
@@ -17,14 +20,14 @@ echo "Packer image hash is $packer_hash"
...
   make build-ami | tee "$packer_file"
-  aws s3 cp "${packer_file}" "s3://${BUILDKITE_AWS_STACK_BUCKET}/${packer_file}"
+  aws s3 cp "${packer_file}" "s3://${BUILDKITE_STACK_BUCKET}/${packer_file}"
 else
   echo "Skipping packer build, no changes"
 fi
 
-image_id=$(grep -Eo "us-east-1: (ami-.+)$" "$packer_file" | awk '{print $2}')
-echo "AMI for us-east-1 is $image_id"
+image_id=$(grep -Eo "${AWS_DEFAULT_REGION}: (ami-.+)$" "$packer_file" | awk '{print $2}')
```
lox: Should this be `$AWS_REGION`?

mikeknox: @lox - I've pushed a rebase now.

`${AWS_DEFAULT_REGION}` vs `${AWS_REGION}`. Both are available at this point, and `${AWS_REGION}` is probably more appropriate so I'll test and update.

.buildkite/steps/publish.sh
```diff
@@ -134,7 +134,7 @@ generate_mappings() {
   local s3_mappings_cache
 
   image_id=$(buildkite-agent meta-data get image_id)
-  s3_mappings_cache="s3://${BUILDKITE_AWS_STACK_BUCKET}/mappings-${image_id}-${BUILDKITE_BRANCH}.yml"
+  s3_mappings_cache="s3://${BUILDKITE_STACK_BUCKET}/mappings-${image_id}-${BUILDKITE_BRANCH}.yml"
```
lox: Any reason for changing this variable name?

lox: Ahhh that's right, there are two different buckets

mikeknox: `$BUILDKITE_AWS_STACK_BUCKET` was used here, and `$BUILDKITE_STACK_BUCKET` in the Makefile.
I only used `$BUILDKITE_STACK_BUCKET` because it was shorter.

lox:
Unfortunately I think the scope of this change is probably going to be too big to accept without having our test suites run against it. 

@mikeknox would you mind if I copy-paste some bits of this change into smaller branches?

lox:
I included a lot of the region agnostic parts of this PR in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/401. Going to close this one out for now.

PR #402, opened 2018-04-15
Turn off cost allocation tags by default

lox:
The addition of cost allocation tags in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/398 broke our tests because it needs to be activated per-account:

https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/activate-built-in-tags.html

I've added a `EnableCostAllocationTags` that defaults to `false`, and also renamed the parameters `CostAllocationTagName` and `CostAllocationTagValue`.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -36,9 +36,6 @@ Metadata:
         - RootVolumeSize
         - AssociatePublicIpAddress
         - ManagedPolicyARN
-        - EnableCostAllocationTags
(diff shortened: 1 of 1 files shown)


lox:
Feels @tduffield?

tduffield:
@lox looks good to me!

lox:
@toolmantim yeah, I was umming and ahhing over names. I'll roll with that.

lox:
Done @toolmantim

PR #405, opened 2018-04-17
Use new Metrics API, drop requirement for org-slug and api-token 🎉

lox:
Using the new private elastic stack metrics API that @sj26 created we can drop the requirement for a personal access token, allowing just an agent registration token. 

Hoping this will be a big win for teams that have to manage lots of personal access tokens just for elastic stacks.

### DIFF
--- a/.buildkite/steps/test.sh
+++ b/.buildkite/steps/test.sh
@@ -13,16 +13,8 @@ echo "Using AMI $image_id"
 cat << EOF > config.json
 [
   {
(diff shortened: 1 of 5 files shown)


toolmantim:
Huzzah!!

sj26:
Yep, looks about right to me. Can we ship this to our own stack as a test? Or just :shipit:?

lox:
It's been running on our elastic stacks all today @sj26. Moving to master and doing some more testing.

sj26:
:clap:

PR #406, opened 2018-04-18
Allow buildkite step to set AWS_REGION and not get overwritten by cfn-env

nemuca:
The cfn-env at the moment will always export AWS_REGION to be the region that the agent is deployed into, which means you cannot override this in the buildkite pipeline or step.  This comit changes the behaviour so that if the environment variable isn't already set by buildkite pipeline nor step, then this will set it to the environment the agent was deployed into.


lox:
The problem with this is that it will affect any call to the aws-cli, which breaks things that are expecting the region to be that of the elastic stack. 

Do you have any examples of why you'd want to change the global AWS_REGION value for a step?

nemuca:
That was the intention, that if you set the AWS_REGION in the buildkite pipeline.yml either via the global env or the various step's env, then you intend those aws cli or sdk commands to run in the context of that AWS_REGION.  

We don't have any public examples, but our use case is that we have the build agents in one region and use these to build and deploy the same AWS assets to many regions.  We have a series of steps where the only different is setting env:AWS_REGION for that step to the destination region.

With this change, if you don't set the AWS_REGION in the pipeline, then the pre-existing behaviour occurs - where it sets AWS_REGION to the region of the agent.

lox:
I wonder if a better approach for elastic stack might be to set a variable that defines the region the stack is in and then specifically use `--region` in code that is region specific.

That would leave `AWS_REGION` to overridable 🤔

lox:
E.g something like `BUILDKITE_STACK_REGION`.

lox:
Btw, what tools are you using that rely on `AWS_REGION` vs `AWS_DEFAULT_REGION`?

nemuca:
Primarily the cli, ruby sdk, and node sdk's.  Aside from these there are numerous existing templates and scripts using the AWS_REGION env var.

Now that you mention it, why not also allow users to set AWS_DEFAULT_REGION if they want?

Keeping in mind that this change is only changing the cfn-env environment bootstrapping which from what I can gather it is only triggered when doing a build step, and therefore should only apply to that specific build process and flow to child processes and not parent processes.

Is a change to the cfn-env going to affect the overall env vars used by other processes not related to the build step being executed?

lox:
> Now that you mention it, why not also allow users to set AWS_DEFAULT_REGION if they want?

Unfortunately there doesn't seem to be clear consensus on how `AWS_REGION` and `AWS_DEFAULT_REGION` differ, otherwise that would be a really good answer :(

toolmantim:
> I wonder if a better approach for elastic stack might be to set a variable that defines the region the stack is in and then specifically use --region in code that is region specific.
> something like `BUILDKITE_STACK_REGION`

This sounds very sensible! And we can default `AWS_REGION` to `BUILDKITE_STACK_REGION` if it isn't set in the pipeline/plugins/hooks

toolmantim:
(and if we have any plugins that are relying on `AWS_REGION` being set by the stack, we could add support for them to use `BUILDKITE_STACK_REGION` if no `AWS_REGION` is set)

lox:
I've had a go at this in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/408, I think this should solve your problem @nemuca?

nemuca:
I've had a look a #408 and see you're removing ambiguity are which region the stack the agent is in, as it will always be known and not overridable, which leaves AWS_REGION and AWS_DEFAULT_REGION at the control of the user.

The changes look like they achieve both outcomes, but I'm not sure of the cost to backward compatibility, specifically for anyone who previously relied on AWS_REGION and AWS_DEFAULT_REGION being set by the agent when their build step runs.  Can you still support backward compatibility by setting these env vars in cfn-env to the value from BUILDKITE_STACK_REGION if not set by the build step.  E.g.

if [[ -z "\${AWS_DEFAULT_REGION-}" ]]; then
  export AWS_DEFAULT_REGION="${BUILDKITE_STACK_REGION}"
fi
if [[ -z "\${AWS_REGION-}" ]]; then
  export AWS_REGION="${BUILDKITE_STACK_REGION}"
fi

That way for those who rely on the build agent setting the value automatically to the agents region, it still works, and for those who set the value themselves in the pipeline, it uses that region.

lox:
Yeah, I think that makes sense @nemuca.

lox:
See https://github.com/buildkite/elastic-ci-stack-for-aws/pull/408/commits/25fa430315990854ce5f6976e039378bc88896be @nemuca

nemuca:
Yep, that looks perfect

lox:
Closing in favour of #408.

PR #408, opened 2018-04-19
Replace AWS_REGION with BUILDKITE_STACK_REGION

lox:
Currently the stack depends on having `AWS_REGION` and `AWS_DEFAULT_REGION` set, but this breaks builds that want to override these values for their own purposes. This PR unsets those values at a stack level and instead uses `BUILDKITE_STACK_REGION`. 

See #406 for more context.

### DIFF
--- a/packer/conf/buildkite-agent/hooks/environment
+++ b/packer/conf/buildkite-agent/hooks/environment
@@ -4,6 +4,11 @@ set -eu -o pipefail
 # shellcheck source=/dev/null
 source ~/cfn-env
 
+# We set BUILDKITE_STACK_REGION in the cfn-env file, so if no value is set by the job
+# then we use that to set the two environment variables that aws tools depend on
+export AWS_DEFAULT_REGION="${AWS_DEFAULT_REGION:-$BUILDKITE_STACK_REGION}"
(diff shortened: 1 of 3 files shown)


packer/conf/buildkite-agent/hooks/environment
```diff
@@ -4,6 +4,11 @@ set -eu -o pipefail
 # shellcheck source=/dev/null
 source ~/cfn-env
 
+# We set BUILDKITE_STACK_REGION in the cfn-env file, so if no value is set by the job
+# then we use that to set the two environment variables that aws tools depend on
+export AWS_DEFAULT_REGION="${AWS_DEFAULT_REGION:-$BUILDKITE_STACK_REGION}"
+export AWS_REGION="${AWS_REGION:-$BUILDKITE_STACK_REGION}"
```
toolmantim: Would this mean that if someone changed their step config's env vars, the secrets/ecr code below this would be affected, and not just their command?

toolmantim: Should we perhaps store the original value of those `AWS_` envs (in case they’re specified), then set it to the `BUILDKITE_STACK_REGION` value (this original code), then restore the `AWS_` envs to their original value after we’ve run the plugins?

lox:
Closing this for now.

PR #415, opened 2018-05-17
Copy docker cron scripts to hourly not daily

lox:
The packer folder is `cron.hourly`, but we copy it into `cron.daily`.

This reconciles that down to hourly.

### DIFF
--- a/packer/scripts/install-docker.sh
+++ b/packer/scripts/install-docker.sh
@@ -37,9 +37,9 @@ sudo chmod +x /usr/bin/docker-compose
 docker-compose --version
 
 echo "Adding docker cron tasks..."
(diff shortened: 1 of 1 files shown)


lox:
I reckon I'm actually going to go the other way on this and make it daily. The cleanup as part of the `environment` hook should cover this, the cron is just to catch any it doesn't.

arturopie:
@lox it seems the `environment` hook only clean up images, but not containers and networks. Can you add those clean ups to the hook as well? I'm referring to theses: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/9fa2db4df055a7ef3630205056e1172a4ca5b46e/packer/conf/docker/cron.hourly/docker-gc#L13

PR #418, opened 2018-05-17
Use the docker directory to check for disk space

arturopie:
Some people might use a RAM drive for docker to get better performance. By specifying the docker dir, we make sure we are checking for disk space in the right filesystem.


arturopie:
@lox what's the status of this PR. It seems like a very simple change. Is there something I'm missing? Anything you want me to fix?

lox:
Apologies, I was thinking it over! Let's do it. Did you end up figuring out how to reliably override those env per-build?

arturopie:
Are you referring to the `DISK_MIN_INODES` and `DOCKER_PRUNE_UNTIL` env variables? Those can be set in the `pipeline.yml` file

PR #419, opened 2018-05-18
Support all availability zones in a region: Take 3

lox:
This allows an elastic stack to run across up to 6 availability zones. The advantage to this is access to spot pricing in a wider range of AZ's. 

This also allows for configuration of those subnets as public (like is the case currently in master), or private. If they are private, then an SSH Bastion host with a public ip can also be created to access them.

Configuration for these different configurations is via the new `SubnetConfiguration` parameter, which gives the following options:

* 2 public subnets with public IP's
* 3 public subnets with public IP's
* 4 public subnets with public IP's
* 5 public subnets with public IP's
* 6 public subnets with public IP's
* 2 private subnets + 2 public subnets with NAT Gateways for internet access
* 3 private subnets + 2 public subnets with NAT Gateways for internet access
* 4 private subnets + 2 public subnets with NAT Gateways for internet access
* 5 private subnets + 2 public subnets with NAT Gateways for internet access
* 6 private subnets + 2 public subnets with NAT Gateways for internet access
* Use existing VPC/Subnets

We've tried to make this work a few times before, but ran into issues getting the upgrade experience to work smoothly. I think I've solved that. 

One remaining concern is whether people will figure out that you need to enter availability zones that match the `SubnetConfiguration` you enter. There is no way to figure them out automatically. I think it's probably worth the slight increase in friction of creating a stack for the major improvement in utility and pricing.

Previous attempts #360

### DIFF
--- a/templates/vpc.yml
+++ b/templates/vpc.yml
@@ -101,7 +101,7 @@ Resources:
     Type: AWS::EC2::VPCGatewayAttachment
     Condition: CreateVpcResources
     Properties:
-      GatewayId: !Ref Gateway
+      InternetGatewayId: !Ref Gateway
       VpcId: !Ref Vpc
 
   # Retaining old names for Subnet0 and Subnet1 to avoid subnet clashes


templates/vpc.yml
```diff
@@ -1,87 +1,531 @@
...
+    !Equals [ !Ref SubnetConfiguration, "4 public subnets with public IP's" ],
+    !Equals [ !Ref SubnetConfiguration, "4 private subnets + 2 public subnets with NAT Gateways for internet access" ]
+  ]
+  HasFiveAZs: !Or [
+    !Equals [ !Ref SubnetConfiguration, "5 public subnets with public IP's" ],
+    !Equals [ !Ref SubnetConfiguration, "5 private subnets + 2 public subnets with NAT Gateways for internet access" ]
+  ]
+  HasSixAZs: !Or [
+    !Equals [ !Ref SubnetConfiguration, "6 public subnets with public IP's" ],
+    !Equals [ !Ref SubnetConfiguration, "7 private subnets + 2 public subnets with NAT Gateways for internet access" ]
```
patrobinson: Should be 6 private subnets?

templates/buildkite-elastic.yml
```diff
@@ -598,8 +586,47 @@ Resources:
   AgentAutoScaleGroup:
     Type: AWS::AutoScaling::AutoScalingGroup
     Properties:
-      VPCZoneIdentifier:
-        { "Fn::If": [ "CreateVpcResources", [ { Ref: Subnet0 }, { Ref: Subnet1 } ], { Ref: Subnets } ] }
+      VPCZoneIdentifier: !If [
+        CreateVpcResources,
+        !If [
```
zsims: This can be simplified to:

```
!If
  - CreateVpcResources
  - !Split
    - ','
    # This eliminates `AWS::NoValue` entries
    - !Join 
      - ','
      - !Ref PrivateSubnet1
      - !Ref PrivateSubnet2
      - !Ref PublicSubnet1
      - !Ref PublicSubnet2
      - !If [HasPrivateSubnet3, !Ref PrivateSubnet3, !Ref 'AWS::NoValue']
      - !If [HasPrivateSubnet4, !Ref PrivateSubnet4, !Ref 'AWS::NoValue']
      - !If [HasPrivateSubnet5, !Ref PrivateSubnet5, !Ref 'AWS::NoValue']
      - !If [HasPrivateSubnet6, !Ref PrivateSubnet6, !Ref 'AWS::NoValue']
      - !If [HasPublicSubnet3, !Ref PublicSubnet3, !Ref 'AWS::NoValue']
      - !If [HasPublicSubnet4, !Ref PublicSubnet4, !Ref 'AWS::NoValue']
      - !If [HasPublicSubnet5, !Ref PublicSubnet5, !Ref 'AWS::NoValue']
      - !If [HasPublicSubnet6, !Ref PublicSubnet6, !Ref 'AWS::NoValue']
```

zsims:
Nice! We always prefer to create our own VPC (for security reasons). Any chance this could be split out into a sub-stack? E.g. in the existing `aws-stack.yml`:

```yaml
...

Resources:
  ElasticCiVpc:
    Condition: CreateVpcResources
    Type: AWS::CloudFormation::Stack
    Properties:
      TemplateURL: https://s3.amazonaws.com/buildkite-aws-stack/latest/aws-stack-vpc.json
      Parameters:
        SubnetConfiguration: !Ref SubnetConfiguration
```

It'd then make the base template much simpler, and allow us (consumers that *always* want our own VPC) to easily read and security vet the template.

PR #421, opened 2018-05-18
Set InstanceRoleName as stack template output

dblandin:
We allow Buildkite agents to assume other IAM roles for certain deployment/operational actions. The role to be assumed needs to include the role ARN within its assume role policy.

It would be great to thread a dependency between the stack and the assume policy document via the stack output.

We use the existing `AutoScalingGroupName` stack output to create/update scheduled actions on the ASG and would like to use the role name to aid in other resources creation.

### DIFF
--- a/templates/outputs.yml
+++ b/templates/outputs.yml
@@ -2,3 +2,5 @@
 Outputs:
   AutoScalingGroupName:
     Value: { Ref: AgentAutoScaleGroup }
+  InstanceRoleName:
+    Value: { Ref: InstanceRoleName }


lox:
Seems sensible, thanks @dblandin!

dblandin:
@lox Thanks!! :100:

arturopie:
Will this output return only the value of the parameter `InstanceRoleName`?
We don't set the `InstanceRoleName` parameter and let CloudFormation create the role for us. In our case, the `InstanceRoleName` output is empty.

dblandin:
@arturopie Hmmm, I might have extracted the value in a suboptimal way.

I'm not too familar with the CloudFormation output syntax but it seems like we might want to reach into the created role and grab it's `RoleName` property instead.

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/1bfa09d9fd6d35761ae0e5fad9b8081ff041a6dc/templates/buildkite-elastic.yml#L366-L369

dblandin:
@arturopie Something like this? https://github.com/buildkite/elastic-ci-stack-for-aws/pull/428

Still need to check if that syntax is valid. Looks like the template is using a mix between the yaml and json function invocation styles.

arturopie:
@dblandin something like that. I'm not sure about the syntax neither. it will need to be tested

PR #424, opened 2018-05-24
Convert LaunchConfiguration to a LaunchTemplate

lox:
This positions us to use the same config for an ASG as a SpotFleet. See https://aws.amazon.com/about-aws/whats-new/2017/11/introducing-launch-templates-for-amazon-ec2-instances/ for some more context.


lox:
Argh, unfortunately I can't find a way to specify a spot price for ASG's with LaunchTemplates.

PR #428, opened 2018-05-25
Set InstanceRoleName output from IAMRole resource

dblandin:
If the `InstanceRoleName` parameter (optional) is unset, the CloudFormation stack generates a name for you based on the stack name.

This commit updates the output to extract the name from the resource directly which accounts for both cases, and should be more reliable going forward.

Follow up to https://github.com/buildkite/elastic-ci-stack-for-aws/pull/421. Addresses https://github.com/buildkite/elastic-ci-stack-for-aws/pull/421#issuecomment-392137428

### DIFF
--- a/templates/outputs.yml
+++ b/templates/outputs.yml
@@ -3,4 +3,4 @@ Outputs:
   AutoScalingGroupName:
     Value: { Ref: AgentAutoScaleGroup }
   InstanceRoleName:
-    Value: { Ref: InstanceRoleName }
+    Value: !GetAtt IAMRole.RoleName


templates/outputs.yml
```diff
@@ -3,4 +3,4 @@ Outputs:
   AutoScalingGroupName:
     Value: { Ref: AgentAutoScaleGroup }
   InstanceRoleName:
-    Value: { Ref: InstanceRoleName }
+    Value: { GetAtt: IAMRole.RoleName }
```
dblandin: The CloudFormation docs show a slightly different function invocation format:

```yaml
!GetRef AgentAutoScaleGroup
!GetAtt IAMRole.RoleName
```

source: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/outputs-section-structure.html

lox:
Ugh, apologies, I should have caught this in code review of #421 😓

templates/outputs.yml
lox: Yup, should be either:

`{ "Fn::GetAtt" : [ "IAMRole", "RoleName" ] }`

Or:

`!GetAtt IAMRole.RoleName`

dblandin:
Updated! The `!GetAtt` syntax is pretty cool :+1:

lox:
Yup, been meaning to convert all of our templates to use it. We convert them to JSON as part of the build process. I only recently figured out how to support the `!Blah` syntax (it's a bit of YAML hack).

lox:
Thanks @dblandin!

PR #429, opened 2018-05-26
#420: move cron jobs to run hourly

### LINKED ISSUE #420: Clean up docker containers and network during global environment hook
Sometimes our jobs fail when trying to create a new docker network. 

E.g.:

image

This would be mitigated if we run a clean up for containers and network at the beginning of the job, when we do the disk clean up.

arturopie:
Fixes #420


lox:
Thanks! Sorry, was on a plane to Iceland 🛫

PR #433, opened 2018-06-07
Add debugging to s3 bucket deletion

lox:
I can't figure out why s3 bucket deletion doesn't work in our cleanup process. This adds some debugging.


lox:
I still have no idea. The cleanup script runs locally, but doesn't clean up resources in prod.

PR #434, opened 2018-06-07
Unfortunately IAMRole.RoleName doesn't exist, so output ARN instead

lox:
This is a follow on from https://github.com/buildkite/elastic-ci-stack-for-aws/pull/428.

@dblandin unfortunately IAMRole.RoleName doesn't exist, so we'll have to settle for the ARN.

### DIFF
--- a/templates/outputs.yml
+++ b/templates/outputs.yml
@@ -2,5 +2,5 @@
 Outputs:
   AutoScalingGroupName:
     Value: { Ref: AgentAutoScaleGroup }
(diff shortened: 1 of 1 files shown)


lox:
I'm going to merge this as the master build is broken currently.

holmesjr:
If you want the role name, just !Ref it.

lox:
> If you want the role name, just !Ref it.

How did I not think of this. 😓Thanks @holmesjr

PR #439, opened 2018-06-22
Fixes permissions issues accessing docker socket

lox:
This makes sure that the gid inside containers is mapped to the docker group on the host, so that if the docker socket is mounted in it can be accessed.

See https://www.jujens.eu/posts/en/2017/Jul/02/docker-userns-remap/#id2 for more context.


lox:
Sneaky merge, master is broken without this.

PR #440, opened 2018-06-22
432: use exit instead of return in bk-check-disk-space.sh script

### LINKED ISSUE #432: bk-check-disk-space.sh Error: return: can only `return' from a function or sourced script
When `bk-dhcek-disk-space.sh` is ran as part of the `environment` hook. See https://github.com/buildkite/elastic-ci-stack-for-aws/blob/684ca4115ad4b15e38693aeaff42a296e80f6f96/packer/conf/buildkite-agent/hooks/environment#L20 it errors with the following message:

image

arturopie:
Fixes #432


lox:
Thanks @arturopie! 🙏🏻

PR #444, opened 2018-07-07
Make KeyName optional

### LINKED ISSUE #413: Is the key / keyname needed. Would you consider not making the keyname mandatory
At first glance EC2 key pair does not look critical to the stack. Could it be done away with as a mandatory option?

Rationale
We aspire to and advocate no ssh keys in production. CI/CD systems given the capability it normally has in the AWS account is definitely considered to be a production system in our world.

Secondly we find having no access to the instances avoids any unrepeatable configuration tom foolery and promotes automation over clickops.

Thanks

zsims:
This closes #413, by making the `KeyName` parameter optional. The parameter type has to be changed to `String` as the `AWS::EC2::KeyPair::KeyName` typed parameter means a valid key name must be specified.

If no `KeyName` is specified, then the SSH ingress rule will not be created.

Some tests:

# With a new stack
```bash
$ aws cloudformation create-stack --stack-name optional-ssh-zac --template-body file://build/aws-stack.json --parameters "ParameterKey=BuildkiteAgentToken,ParameterValue=secret" --capabilities CAPABILITY_NAMED_IAM
# ... 
CREATE_COMPLETE
```

# Against an *existing* 3.2.1 stack
```bash
$ aws cloudformation create-change-set --stack-name buildkite --change-set-name optional-ssh-key-test --template-body file://build/aws-stack.json --parameters "ParameterKey=BuildkiteAgentToken,ParameterValue=secret" --capabilities CAPABILITY_NAMED_IAM
# ...
$ aws cloudformation describe-change-set --stack-name buildkite --change-set-name optional-ssh-key-test --query 'Changes[].ResourceChange.{LogicalResourceId:LogicalResourceId,Replacement:Replacement,Action:Action}'
[
    {
        "LogicalResourceId": "AgentAutoScaleGroup",
        "Replacement": "Conditional",
        "Action": "Modify"
    },
    {
        "LogicalResourceId": "AgentLaunchConfiguration",
        "Replacement": "True",
        "Action": "Modify"
    },
    {
        "LogicalResourceId": "AgentLifecycleHook",
        "Replacement": "Conditional",
        "Action": "Modify"
    },
    {
        "LogicalResourceId": "AgentScaleDownPolicy",
        "Replacement": "Conditional",
        "Action": "Modify"
    },
    {
        "LogicalResourceId": "AgentScaleUpPolicy",
        "Replacement": "Conditional",
        "Action": "Modify"
    },
    {
        "LogicalResourceId": "AgentUtilizationAlarmHigh",
        "Replacement": "False",
        "Action": "Modify"
    },
    {
        "LogicalResourceId": "AgentUtilizationAlarmLow",
        "Replacement": "False",
        "Action": "Modify"
    },
    {
        "LogicalResourceId": "SecurityGroup",
        "Replacement": "True",
        "Action": "Modify"
    },
    {
        "LogicalResourceId": "SecurityGroupSshIngress",
        "Replacement": null,
        "Action": "Add"
    }
]
```


lox:
This is looking great, thanks for this @zsims. My only feedback is that the conditional SSH ingress needs to also take into account `AuthorizedUsersUrl` which lets you provide a url for downloading `authorized_keys` from.

zsims:
Thanks for the review @lox, have updated the PR with your suggestion

templates/buildkite-elastic.yml
```diff
@@ -332,6 +332,16 @@ Conditions:
...
 
+    HasKeyName:
+      "Fn::Not": [ "Fn::Equals": [ { Ref: KeyName }, "" ] ]
+
+    EnableSshIngress:
+      "Fn::And":
+        - { Condition: HasKeyName }
+        - { Condition : CreateSecurityGroup }
+        # Enable ingress if authorized users (keys) can be specified another way
+        - "Fn::Not": [ "Fn::Equals": [ { Ref: AuthorizedUsersUrl }, "" ] ]
```
lox: This probably needs to be in an `Fn::Or`, or am I booleaning wrong?

zsims: Well that's embarrassing, good catch! I've pushed up a fix

PR #448, opened 2018-07-11
Reference security group by id rather than name

zsims:
This handles default vs non default VPC reference rules per
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-ingress.html#cfn-ec2-security-group-ingress-groupid

This follows on from #444 which made the rule conditional.


lox:
Seems legit!

PR #453, opened 2018-07-17
Set correct variable to pass to upstream ecr plugin

bshelton229:
I believe the variable exported after reading `AWS_ECR_LOGIN_REGISTRY_IDS` isn't correct to configure the ecr login plugin. We're currently getting around this by explicitly setting `BUILDKITE_PLUGIN_ECR_ACCOUNT_IDS_0` in our stack, but thought it was worth trying to get fixed to configure this the documented way.

https://github.com/buildkite-plugins/ecr-buildkite-plugin/blob/3068b26de2eb43b8ae6ebf081b1f0626d018d8c0/hooks/pre-command#L58

Thanks!


lox:
Thanks, appreciated!

PR #462, opened 2018-08-01
Update to latest Amazon Linux 2 LTS

lox:


holmesjr:
Good idea. I noticed the agents complain about a lot of yum updates with the older AMI.

lox:
We've been seeing some strange network performance issues with the current stack version. Hoping this fixes it.

jeremiahsnapp:
@lox I noticed that the new stack no longer exports `GODEBUG=netdns=cgo` from the PR linked below. Do you think this could be related to the issues you are seeing?

https://github.com/buildkite/elastic-ci-stack-for-aws/pull/233

PR #469, opened 2018-08-15
Add secrets manager hooks

lox:
This adds a param `EnableSecretsManagerPlugin` that brings in the experimental https://github.com/buildkite/elastic-ci-stack-secrets-manager-hooks.

### DIFF
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -451,6 +451,7 @@ Resources:
             Action:
               - secretsmanager:ListSecrets
               - secretsmanager:GetSecretValue
+            Resource: "*"
       Roles:
         - { Ref: IAMRole }
 
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -447,6 +447,10 @@ Resources:
               - sns:Unsubscribe
               - sns:Subscribe
             Resource: "*"
+          - Effect: Allow
+            Action:
+              - secretsmanager:ListSecrets
+              - secretsmanager:GetSecretValue
       Roles:
         - { Ref: IAMRole }
 
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -449,9 +449,17 @@ Resources:
             Resource: "*"
           - Effect: Allow
             Action:
-              - secretsmanager:ListSecrets
+              - secretsmanager:List*
+              - secretsmanager:Describe*
+            Resource: !Sub "arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:buildkite/*"
+          - Effect: Allow
+            Action:
               - secretsmanager:GetSecretValue
-            Resource: "*"
+            Resource: !Sub "arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:buildkite/*"
+            Condition:
+              ForAnyValue:StringEquals:
+                "secretsmanager:VersionStage": "AWSCURRENT"
+
       Roles:
         - { Ref: IAMRole }
 
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -567,6 +567,7 @@ Resources:
               SECRETS_PLUGIN_ENABLED=${EnableSecretsPlugin} \
               ECR_PLUGIN_ENABLED=${EnableECRPlugin} \
               DOCKER_LOGIN_PLUGIN_ENABLED=${EnableDockerLoginPlugin} \
+              SM_SECRETS_HOOKS_ENABLED=${EnableSecretsManagerPlugin} \
               DOCKER_USERNS_REMAP=${EnableDockerUserNamespaceRemap} \
               AWS_REGION=${AWS::Region} \
                 /usr/local/bin/bk-install-elastic-stack.sh
--- a/templates/buildkite-elastic.yml
+++ b/templates/buildkite-elastic.yml
@@ -67,6 +67,7 @@ Metadata:
           default: Plugin Configuration
         Parameters:
         - EnableSecretsPlugin
+        - EnableSecretsManagerPlugin
         - EnableECRPlugin
         - EnableDockerLoginPlugin
 
@@ -237,6 +238,14 @@ Parameters:
       - true
       - false
     Default: "true"
+
+  EnableSecretsManagerPlugin:
+    Type: String
+    Description: Enables secrets-manager plugin for all pipelines
+    AllowedValues:
+      - true
+      - false
+    Default: "false"
 
   EnableECRPlugin:
     Type: String
@@ -664,7 +673,7 @@ Resources:
   SecurityGroupSshIngress:
     Condition: EnableSshIngress
     Type: AWS::EC2::SecurityGroupIngress
-    Properties: 
+    Properties:
       GroupId: !GetAtt SecurityGroup.GroupId
       IpProtocol: tcp
       FromPort: 22
--- a/.buildkite/steps/test.sh
+++ b/.buildkite/steps/test.sh
@@ -64,6 +64,10 @@ cat << EOF > config.json
   {
     "ParameterKey": "EnableDockerUserNamespaceRemap",
     "ParameterValue": "true"
+  },
+  {
+    "ParameterKey": "EnableSecretsManagerPlugin",
+    "ParameterValue": "true"
   }
 ]
 EOF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,3 +9,6 @@
 [submodule "plugins/docker-login"]
 	path = plugins/docker-login
 	url = https://github.com/buildkite-plugins/docker-login-buildkite-plugin.git
+[submodule "plugins/secrets-manager"]
+	path = plugins/secrets-manager
+	url = https://github.com/buildkite/elastic-ci-stack-secrets-manager-hooks.git
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Features:
 - Configurable spot instance bid price
 - Configurable auto-scaling based on build activity
 - Docker and Docker Compose support
-- Per-pipeline S3 secret storage (with SSE encryption support)
+- AWS Secrets Manager integration
 - Docker Registry push/pull support
 - CloudWatch logs for system and buildkite agent events
 - CloudWatch metrics from the Buildkite API
@@ -68,40 +68,7 @@ aws cloudformation create-stack \
 
 ## Build Secrets
 
-The stack will have created an S3 bucket for you (or used the one you provided as the `SecretsBucket` parameter). This will be where the agent will fetch your SSH private keys for source control, and environment hooks to provide other secrets to your builds.
-
-The following s3 objects are downloaded and processed:
-
-* `/env` - An [agent environment hook](https://buildkite.com/docs/agent/hooks)
-* `/private_ssh_key` - A private key that is added to ssh-agent for your builds
-* `/git-credentials` - A [git-credentials](https://git-scm.com/docs/git-credential-store#_storage_format) file for git over https
-* `/{pipeline-slug}/env` - An [agent environment hook](https://buildkite.com/docs/agent/hooks), specific to a pipeline
-* `/{pipeline-slug}/private_ssh_key` - A private key that is added to ssh-agent for your builds, specific to the pipeline
-* `/{pipeline-slug}/git-credentials` - A [git-credentials](https://git-scm.com/docs/git-credential-store#_storage_format) file for git over https, specific to a pipeline
-
-These files are encrypted using [Amazon's KMS Service](https://aws.amazon.com/kms/). See the [Security](#security) section for more details.
-
-Here's an example that shows how to generate a private SSH key, and upload it with KMS encryption to an S3 bucket:
-
-```bash
-# generate a deploy key for your project
-ssh-keygen -t rsa -b 4096 -f id_rsa_buildkite
-pbcopy < id_rsa_buildkite.pub # paste this into your github deploy key
-
-aws s3 cp --acl private --sse aws:kms id_rsa_buildkite "s3://${SecretsBucket}/private_ssh_key"
-```
-
-If you want to set secrets that your build can access, create a file that sets environment variables and upload it:
-
-```bash
-echo "export MY_ENV_VAR=something secret" > myenv
-aws s3 cp --acl private --sse aws:kms myenv "s3://${SecretsBucket}/env"
-rm myenv
-```
-
-**Note: Currently only using the default KMS key for s3 can be used, follow [#235](https://github.com/buildkite/elastic-ci-stack-for-aws/issues/235) for progress on using specific KMS keys**
-
-If you really want to store your secrets unencrypted, you can disable it entirely with `BUILDKITE_USE_KMS=false`.
+Secrets are stored in Amazon Secrets Manager. See [Secrets Manager Documentation](plugins/secrets-manager/README.md) for more details and examples.
 
 ## What’s On Each Machine?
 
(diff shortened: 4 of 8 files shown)


lox:
Thoughts on how to bring this in @toolmantim? The idea is it would replace the s3 secrets stuff, but I'm unsure on how to approach deprecating the old stuff.

toolmantim:
This is so great, much better than the S3 method.

I wonder if the S3 method is still best for general hook scripts support (maybe even adding more than just `environment` hooks as well) — and SM is for secrets only?

Also the other outstanding thing I was wondering was, how well would a end-user developer (or semi-experienced dev) go with setting a secret for their pipeline, if all we have is https://github.com/buildkite/elastic-ci-stack-secrets-manager-hooks? The readme at the moment covers a lot, and the prefix makes it tricky to know what to type exactly (vs a bucket, which is kind of just sitting there and has a web UI).

I think this method adds more onus on us to make it easy to use.

toolmantim:
`gitcredential helper` 😍

lox:
Yeah, it's tricky @toolmantim. Generally speaking, I think the developer experience for this is radically better, because you can set individual ENV values. The current setup you need to download an env file and modify it and re-upload it, with very specific SSE settings. 

Given users are several steps disconnected from the stack they are running on, I'm not sure how we give them exactly the correct syntax 🤔Same deal with the current s3 bucket deal though. Any ideas? 

I was thinking this morning that it would be neat to have the agent read it's entire config and hooks from secrets manager when it starts up. Then you'd have a central place to configure it.

templates/buildkite-elastic.yml
```diff
@@ -438,6 +447,11 @@ Resources:
               - sns:Unsubscribe
               - sns:Subscribe
             Resource: "*"
+          - Effect: Allow
+            Action:
+              - secretsmanager:ListSecrets
+              - secretsmanager:GetSecretValue
+            Resource: "*"
```
toolmantim: Can we limit this by prefix?

toolmantim:
> The current setup you need to download an env file and modify it and re-upload it, with very specific SSE settings.

Yep totally. I'm more talking about the experience of figuring out how to do it for the very first time, for the person creating their very first stack. Perhaps someone who set up the elastic CI stack for the web UI.

The ongoing "maintain an env file" experience of S3 is definitely crap, it's just very accessible. But this is totally the better way, I think we might just need to help first-timers a bit more with docs.

Perhaps our SM examples should just assume the default stack name / default secrets prefix, e.g. remove all the `export secrets_manager_prefix="elastic-ci-stack"` lines from the readme.

Something like…

> ## Configuring Git Credentials
>
> Here's an example for how you'd configure a global git credential for the entire stack, using a [GitHub personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/):
>
> ```
> aws secretsmanager create-secret \
>   --name "buildkite-elastic-ci-stack/global/git-credentials" \
>   --secret-string "https://<username>:<access-token>@github.com"
> ```
>
> You can also override it a per-pipeline basis:
>
> ```
> aws secretsmanager create-secret \
>   --name "buildkite-elastic-ci-stack/pipeline/<pipeline-slug>/git-credentials" \
>   --secret-string "https://<username>:<access-token>@github.com"
> ```

Maybe we could just work to improve the readme together? Or should it go into elastic CI stack readme? 🤔

> I was thinking this morning that it would be neat to have the agent read it's entire config and hooks from secrets manager when it starts up. Then you'd have a central place to configure it.

Nice! The agent token could be read from SM. Do people use SM as a general config store?

lox:
Perhaps we don't need that key prefix. I've been trying to figure out how teams will delegate access to things, perhaps we go with something like:

* `/buildkite/{org_slug}/ssh-private-key`
* `/buildkite/{org_slug}/git-credentials`
* `/buildkite/{org_slug}/env/{env_name}`
* `/buildkite/{org_slug}/pipeline/{pipeline_slug}/ssh-private-key`
* `/buildkite/{org_slug}/pipeline/{pipeline_slug}/git-credentials`
* `/buildkite/{org_slug}/pipeline/{pipeline_slug}/env/{env_name}`

We could even add per-team support:

* `/buildkite/{org_slug}/team/{team_slug}/ssh-private-key`
* `/buildkite/{org_slug}/team/{team_slug}/git-credentials`
* `/buildkite/{org_slug}/team/{team_slug}/env/{env_name}`
* `/buildkite/{org_slug}/team/{team_slug}/pipeline/{pipeline_slug}/ssh-private-key`
* `/buildkite/{org_slug}/team/{team_slug}/pipeline/{pipeline_slug}/git-credentials`
* `/buildkite/{org_slug}/team/{team_slug}pipeline/{pipeline_slug}/env/{env_name}`

That would mean you structured your secrets by buildkite constructs. Would make it much easier to have "copy this command" things.

lox:
We can do prefix matching on the secrets manager ARN's, so we would default to `/buildkite/*` but you could have a stack that only had access to particular org, or a set of pipelines 🤔

toolmantim:
This seems really nice:

* `/buildkite/{org_slug}/ssh-private-key`
* `/buildkite/{org_slug}/git-credentials`
* `/buildkite/{org_slug}/env/{env_name}`
* `/buildkite/{org_slug}/pipeline/{pipeline_slug}/ssh-private-key`
* `/buildkite/{org_slug}/pipeline/{pipeline_slug}/git-credentials`
* `/buildkite/{org_slug}/pipeline/{pipeline_slug}/env/{env_name}`

You could also add:

* `/buildkite/agent-access-token`

toolmantim:
Which would make it:

> ## Configuring Git Credentials
>
> Here's an example for how you'd configure a global git credential for the entire stack, using a [GitHub personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/):
>
> ```
> aws secretsmanager create-secret \
>   --name "/buildkite/<org-slug>/git-credentials" \
>   --secret-string "https://<username>:<access-token>@github.com"
> ```
>
> You can also override it a per-pipeline basis:
>
> ```
> aws secretsmanager create-secret \
>   --name "/buildkite/<org-slug>/pipeline/<pipeline-slug>/git-credentials" \
>   --secret-string "https://<username>:<access-token>@github.com"
> ```

I wonder if we should drop the initial `/` though?

> ## Configuring Git Credentials
>
> Here's an example for how you'd configure a global git credential for the entire stack, using a [GitHub personal access token](https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/):
>
> ```
> aws secretsmanager create-secret \
>   --name "buildkite/<org-slug>/git-credentials" \
>   --secret-string "https://<username>:<access-token>@github.com"
> ```
>
> You can also override it a per-pipeline basis:
>
> ```
> aws secretsmanager create-secret \
>   --name "buildkite/<org-slug>/pipeline/<pipeline-slug>/git-credentials" \
>   --secret-string "https://<username>:<access-token>@github.com"
> ```

I think that seems nicer?

lox:
> I wonder if we should drop the initial / though?

Yeah, that makes sense. I was following the convention that parameter store uses, but I don't think that is a thing with Secrets Manager.

It's really hard to know how to piece the things together Systems Manager Parameter Store is the other thing, which can store secrets too. There is a comparison [here](https://medium.com/@nimboya/aws-secrets-manager-vs-hashicorp-vault-vs-aws-parameter-store-bcbf60b0c0d1). The advantage the Secret Manager has is that it has a minimal web UI and it also supports key rotation (or at least makes it possible to write a lambda to do it).

lox:
Cool. I reckon that is much nicer now.

toolmantim:
Nice, I reckon so too!

I think adding access token support is probably really smart too… any security conscious person probably raises an eyebrow when pasting a token into a CloudFormation form field.

But if you start to push all the other bits of config into SM, you just gotta remember that each key costs (which is easier to swallow for a secret value)

lox:
> /buildkite/agent-access-token

I was thinking that would be cool to actually have in the agent. It gets a bit blurrier how you map the many different types of agents you have onto which registration token.

lox:
I was thinking of using https://aws.amazon.com/blogs/mt/integrating-aws-cloudformation-with-aws-systems-manager-parameter-store/ for the agent registration token in the cloudformation template to start with.

lox:
Now the gnarly bit is how we replace the s3 stuff with the new secrets manager stuff, and how we document both in the Elastic Stack Readme.

toolmantim:
Switch this to being the default, and have an option to enable "Legacy S3 Secrets Support" (which we drop in the future)?

toolmantim:
(and that option can default to being on for now?)

lox:
Yeah, that is the way I was leaning too. Then do we just link to https://github.com/buildkite/elastic-ci-stack-secrets-manager-hooks in the docs, or do we copy the docs into the main README?

templates/buildkite-elastic.yml
```diff
@@ -449,9 +449,17 @@ Resources:
...
             Action:
-              - secretsmanager:ListSecrets
+              - secretsmanager:List*
+              - secretsmanager:Describe*
+            Resource: !Sub "arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:buildkite/*"
+          - Effect: Allow
+            Action:
               - secretsmanager:GetSecretValue
-            Resource: "*"
+            Resource: !Sub "arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:buildkite/*"
```
toolmantim: Nice!

templates/buildkite-elastic.yml
```diff
@@ -449,9 +449,17 @@ Resources:
...
+            Condition:
+              ForAnyValue:StringEquals:
+                "secretsmanager:VersionStage": "AWSCURRENT"
```
toolmantim: Nice!

toolmantim:
Looks like the `Configuration` section still needs to go: https://github.com/buildkite/elastic-ci-stack-secrets-manager-hooks#configuration

And then I think just link to that readme?

What did you think of my GitHub example for the `gitcredentials-helper`?

lox:
> What did you think of my GitHub example for the gitcredentials-helper?

I thought it was good. Lemme wrangle it in there.

toolmantim:
With the example, I figured if it was a global cred, then it wouldn't be limited to a GitHub org/repo because the access tokens aren't scoped like that either.

Related, should we make it clear that we don't offer any isolation of the secrets? In that, any pipeline's build can read secrets from anything in `buildkite/*`?

lox:
README is updated now @toolmantim

toolmantim:
> I've been pondering whether we could actually do better than this. We could accept a list of pipelines that the stack would work for as a CloudFormation param?

Interesting! Could make sense until agent queues can be locked down on the backend.

lox:
Hmmmmm, maybe we should have used `buildkite/{org_slug}/queue/{queue_name}` as the prefix! 😅Then you could guarantee a stack could only access a give queue's secrets.

lox:
Now that I think about it, it's kind of sketchy that different stacks can all access the entire `buildkite/*` secrets namespace. Makes it hard to create a `deploy` stack and a `default` stack with secrets and access control.

toolmantim:
Oh yeah! 🤔 I guess using the queue name parameter in the permissions scope (and keyname) is probably the only viable way to do it?

lox:
We could use tags? 🤔

```
# Create a secret available on all pipelines in a given queue
aws secretsmanager create-secret \
  --name "/buildkite/<org-slug>/git-credentials" \
  --secret-string "https://<username>:<access-token>@github.com" \
  --tags "Key=queue,Value=default"
```

lox:
In the keyname is probably simpler. It would mean we'd need to bring back the org-slug in the CloudFormation template too, but that isn't a huge deal.

toolmantim:
Oh interesting idea about tags! You could have the `default` queue not have the IAM permission to require the tag, but any other queue require the secret have the matching queue tag. So for the standard case, it's still simple.

lox:
Although, the keys have to be unique, so you'd have collisions if you wanted different values for different key contexts.

lox:
E.g have one `buildkite/<org-slug>/git-credentials` for your `default` queue and another value for your `deploy` queue.

lox:
Unless we allowed `/buildkite/<org-slug>/git-credentials/1` etc

toolmantim:
Yeah okay, I reckon you're right that just adding the queue name is probably the best way to do it. And those simple examples can just include `default` in them.

lox:
I'm going to close this, it needs some more thought.

lox:
How it will co-exist with the s3 secrets and how it fits into the release of 4.x vs perhaps going into the new ECS stack.

lox:
Basically it was just too many variables for my brain to cope with 😅

toolmantim:
Entirely fair!

lox:
The other option would to do away with conventions for paths and just let folks provide an env var in their pipeline that specifies a Secrets Manager key to look up the ssh key or git-credentials for checkout. 

Something like `SSH_SECRETS_MANAGER_KEY` or `GIT_CREDENTIALS_SECRETS_MANAGER_KEY`. Then you would just need to make sure that your stack had access to what ever key that pointed to.

lox:
Would end up being closer in spirit to https://github.com/buildkite/docker-ssh-env-config.

toolmantim:
Do you mean, removing the pipeline slug part… so something like `buildkite/{queue_name}/{key}` with perhaps some magical env syntax? Like:

```
steps:
  - command: deploy.sh
    env:
      - SOME_SECRET="@some-secret-key"
```

and then that gets substituted via the SSM hook with the contents of `buildkite/{queue_name}/some-secret-key`, and can show a nice error if it doesn't exist (perhaps even showing an aws cli command example to set the key?

toolmantim:
(I stole that syntax from https://zeit.co/docs/features/env-and-secrets btw)

toolmantim:
(more docs here: https://zeit.co/docs/getting-started/secrets)

lox:
The hard problem is getting ssh or https credentials for the initial checkout. That is 99% of the problem I'm trying to solve, the rest could be in a plugin with similar syntax to the above.

toolmantim:
What's the problem with the original convention approach, but with only these two that are automatically searched for git creds:

* `buildkite/{queue_name}/{pipeline_slug}/ssh-private-key` (auto searched for via git checkout)
* `buildkite/{queue_name}/{pipeline_slug}/git-credentials` (auto searched for via git checkout)

If you want to use your own custom secrets, you use:

* `buildkite/{queue_name}/{pipeline_slug}/{key}`

```shell
aws secretsmanager create-secret \
  --name "buildkite/default/npm-token-ci-user" \
  --secret-string "..."
```

Then, if you wanted to expose that to every step in your pipeline:

```yml
env:
  - NPM_TOKEN="@npm-token-ci-user"
steps:
  - label: "🛠"
    command: test.sh
```

or just a single step:

```yml
steps:
  - label: "🛠"
    command: test.sh
    env:
      - NPM_TOKEN="@npm-token-ci-user"
```

or if you wanted to pass into a container:

```yml
steps:
  - label: "🛠"
    plugins:
      docker-compose#v2.5.1:
        run: tests
        env:
          - NPM_TOKEN
    env:
      - NPM_TOKEN="@npm-token-ci-user"
```

lox:
@toolmantim yup, I think just implementing those two keys would be a pragmatic compromise. I'll do that. Thanks for talking it through.

toolmantim:
👍🏼

I just updated the custom secret example, to remove the `pipeline_slug` part — I didnt want to give people the illusion that a secret is only visible to a pipeline, security wise.

But it made we wonder about the other guy creds ones.

How do we make it clear, in the API itself, that the secrets are visible to all pipelines? 🤔

toolmantim:
Actually, I think the API we’ve got above is fine, as far as making that clear. We probably just need it littered through the readme and what-not too, that secrets are visible to jobs on that queue. And if you want to lock that to teams etc, you need a custom env hook (although we could add team-lockdown support as a stack param)

lox:
@toolmantim my intention was to implement just the two automatically searched paths:

* buildkite/{queue_name}/{pipeline_slug}/ssh-private-key (auto searched for via git checkout)
* buildkite/{queue_name}/{pipeline_slug}/git-credentials (auto searched for via git checkout)

I think the other stuff is better handled holistically with a plugin or a top-level feature of the agent / platform.

lox:
The reason I had the `{org_slug}` in there is that folks quite often have multiple orgs backed by the one AWS account. I think it's probably worth keeping in there. Thoughts @toolmantim?

PR #476, opened 2018-09-04
Added Timeout as parameter so it can be passed from stack parameters.

RexChenjq:
When using the template, timeout value for autoscaling group->creation policy can now be customised. 
Default value is PT5M which sometimes blocks the deployment as the job it is running is too heavy and 5mins is not able to finish the ASG setup.


rex-myob:
@lox  Yep good idea.

I have renamed the parameter to InstanceCreationTimeout and fixed the syntax of !Ref by using { Ref } to maintain the consistency of the style.

Thanks!

lox:
Thanks very much!

PR #477, opened 2018-09-04
Add journald cloudwatch logs

### LINKED ISSUE #459: Missing some logs on 4.0.0-rc1
Instances built against the 4.0.0-rc1 version of the stack no longer manage to ship a couple of their logs into Cloudwatch (e.g. `/buildkite/docker-daemon`)

The failure seems to be in the local logging config on the instances, because in `/var/log/awslogs.log` we see:

```
2018-07-22 23:11:09,783 - cwlogs.push.stream - WARNING - 3618 - Thread-1 - No file is found with given path '/var/log/lifecycled'.
2018-07-22 23:11:14,784 - cwlogs.push.stream - WARNING - 3618 - Thread-1 - No file is found with given path '/var/log/docker'.
```

(And yep, those paths are actually missing.)

If I had to guess, this could be caused by the move to journald (systemd) in Amazon Linux 2. (`sudo journalctl -u docker.service` on an instance does show the Docker daemon logs)

lox:
Since we swapped to Amazon Linux 2 (and systemd) lots of our logs go via `systemd`, which isn't supported by `awslogs`. This adds @sj26's fork of https://github.com/sj26/journald-cloudwatch-logs to copy the systemd journal to cloudwatch logs.

Closes #459

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,7 +273,7 @@ Provide us with logs from Cloudwatch Logs:
 
 ```
 /buildkite/elastic-stack/{instance-id}
-/buildkite/docker-daemon/{instance-id}
+/buildkite/systemd/{instance-id}
 ```
 
 Alternately, drop by `#aws-stack` and `#aws` channels in [Buildkite Community Slack](https://chat.buildkite.com/) and ask your question!
--- a/packer/buildkite-ami.json
+++ b/packer/buildkite-ami.json
@@ -26,6 +26,10 @@
     },
     {
       "type": "shell",
+      "script": "scripts/install-journald-cloudwatch-logs.sh"
+    },
+    {
+      "type": "shell",
       "script": "scripts/install-utils.sh"
     },
     {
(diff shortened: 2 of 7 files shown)


lox:
@sj26 @toolmantim it pains me that this change means that where previously we had different log streams for agents, lifecycled and the system and now we have a single one.

It's a considerable drop in usability IMO, but I can't figure out any non-epic way to split them out into files from systemd or fragment them into streams with https://github.com/sj26/journald-cloudwatch-logs.

sj26:
I'd always wanted to modify the journald-to-cloudwatch thing to allow creating multiple log streams, where each stream can have distinct journal queries feeding it. Would that be better?

lox:
I thought you might suggest that @sj26. Such nerd sniping 😭

lox:
I had a bit of a look, I think that will be a pretty major undertaking @sj26 as I think you'd need to add some sort of structure to the config file to declare different log groups. It would solve my problem though.

lox:
I might merge this for now, but will consider this before the final 4.x release.

sj26:
:+1:

PR #483, opened 2018-09-11
Make templates modular and use nested stacks

lox:
Previously the stack used some node code to merge the `templates/` into a single template in json and yaml form. 

To prepare for a future where folks can mix and match parts of the stack, this breaks the stack into smaller stacks, and then a main stack that composes those smaller stacks.

The stacks are:

* **Agent stack**.  This includes the autoscaling group that runs the agent instances.
* **VPC**. The basic networking stack. This is the one that is most commonly replaced
* **Metrics**. The agent metrics lambda. Having this modular makes it easier to run a single metrics stack for many elastic stacks.

Finally there is `aws-stack`, which is the main stack that references the above nested stacks. 

The other objective in this refactor was to make it much easier to create your own stacks for testing and development. The `Makefile` is majorly cleaned up and should be easier to use.


toolmantim:
Sounds like a good approach. What's the different in the UX for setting up and updating stacks, from an end user POV? Does it look/operate differently in CloudFormation?

lox:
@toolmantim it's pretty similar in terms of UX. You just see nested stacks in the UI:

image

toolmantim:
Okay cool, that's good! If you bug fixed one of the stacks, you'd re-release the main one and they update that?

lox:
Closing for now, going to focus on simplifying the existing stack first.

PR #493, opened 2018-09-17
Add docs for using the bootstrap script

toolmantim:
Quite a few times on support I've wanted to link people to the docs telling them how to customise their agent (e.g. setting a Docker daemon option, or installing an extra CLI tool).

This adds a docs section showing people how to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,7 +196,26 @@ To debug an agent first find the instance id from the agent in Buildkite, head t
 
 # Customizing Instances with a Bootstrap Script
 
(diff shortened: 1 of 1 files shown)


lox:
Good idea, been meaning to do this for a while. 

If they wanted to add an IAM policy, the could create one like:

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "s3:GetObject",
      ],
      "Resource": ["arn:aws:s3:::my_bucket_name/my_bootstrap.sh"]
    }
  ]
}
```

They would need to attach it to the Role of the Elastic Stack, or make it a managed policy.

toolmantim:
If we assume the stack doesn’t exist yet, how would you do it?

Create a IAM policy with that, and then specify the policy’s ARN in `ManagedPolicyARN` yeah?

Probably need a section “Granting instances IAM permissions” or some such, and we could link to there.

toolmantim:
Okay, this is ready to go

PR #494, opened 2018-09-27
Support timestamp-lines

### LINKED ISSUE #485: Include `timestamp-lines` option in template
I'd be great to allow setting the `timestamp-lines` agent option in the cloudformation template (false to be BC), to allow having a cluster with timestamped output

raylu:
This fixes #485.
See: https://github.com/buildkite/agent/pull/501 and https://buildkite.com/docs/agent/v3/configuration


lox:
Thanks @raylu, lots of thumbs up! 😀

arturopie:
@lox why isn't `BuildkiteAgentTimestampLines` enabled by default. Is it too noisy?

raylu:
I didn't want to change the default behavior in case anyone upgrades, but maybe it's a good idea.

PR #495, opened 2018-09-30
Tunable knob for ASG Cooldown period

prateek:


lox:
Thanks @prateek, can you tell us a bit more about your use case? Interested to hear a bit more about what problem you are trying to solve that needs the cooldown periods customized!

prateek:
tl;dr - the 300 hardcoded value was not in line with my expectation of elasticity of the stack. 

@lox sure thing! The use-case is standard CI - my team has one mega (github.com/m3db/m3) repo and some smaller repos (github.com/m3db/*) – we're in the process of consolidating in a monorepo right now. Developers in the team rely on the for quick PR feedback, so push to GH/CI frequently. 

The reason for the PR might completely be misconfiguration (please tell me if it is).

Here are the knobs I had when I was testing:
- ScaleUpAdjustment: 6
- ScaleDownAdjustment: -3
- ScaleDownPeriod: 120

Along with that, each of my CI builds has 6 jobs. So what would end up happening was I'd push some code, look at the build and go incorporate feedback. By the time I pushed again, the stack had hit the cooldown period once and had triggered an instance kill. At this point, it would wait for the cooldown period before spinning up more instances.

lox:
After some thought, if I have to choose between a config option with a warning on it and no control for folks that need it, I'll choose the config option. Let's get this merged in and documented and I'll do some testing. Thanks for the perspective @ecv and apologies it's taken me so long @prateek.

templates/aws-stack.yml
```diff
@@ -185,6 +186,12 @@ Parameters:
     Default: -1
     MaxValue: 0
 
+  ScaleCooldownPeriod:
+    Description: Minimum number of seconds to wait between any two scaling group changes
```
toolmantim:
```suggestion
    Description: Minimum number of seconds to wait between scaling events. Must be greater than the time needed for an instance to boot.
```

lox: Yup, that works!

dhalperi:
@lox, 27 days ago
> 
> After some thought, if I have to choose between a config option with a warning on it and no control for folks that need it, I'll choose the config option. Let's get this merged in and documented and I'll do some testing. Thanks for the perspective @ecv and apologies it's taken me so long @prateek.

Any reason this is not merged yet?

lox:
A happy new year to you and your family too @dhalperi! We’ll get this out soon.

PR #500, opened 2018-11-05
Cleanup agent params

lox:
This cleans up how a few of the agent parameters in the stack are presented and moves some of the optional ones to an `Advanced Buildkite Configuration` heading.


toolmantim:
Good idea. I wonder if it should be "Agent Configuration" and "Advanced Agent Configuration (Optional)"

PR #502, opened 2018-11-28
update make command for packer

themrmax:


lox:
Thanks and good catch!

PR #508, opened 2018-12-19
Ensure region is set before starting awslogsd

dgarbus:
We recently upgraded our stacks to 4.0.1 and noticed that all our logs are being sent to us-east-1, despite the stacks being created in us-west-2. In a similar vein to #507, we should ensure that the awslogsd service is not started until the region is properly set via the `bk-install-elastic-stack.sh` script.

It looks like this was previously added in #225 but was accidentally removed in 4.0 during the switch to Amazon Linux 2/systemd.


lox:
Argh, good catch @dgarbus!

Issue #452, opened 2018-07-16
Support multiple stacks for different buildkite orgs in the same AWS account 


lox:
Via a customer request:

> Had a question, We have two elastic CI stacks in the same AWS account, one for *** buildkite account and another for the **** buildkite account. Now it seems like both agents have their queue as `default` which is messing with their independent autoscaling...

> Would it be possible to prefix or suffix the metric name with the stack name or something similar so that each stack (i.e. metric lambda function) ships to the respective metric and assists in autoscaling.

PR #510, opened 2018-12-28
Add Buildkite Org to Cloudwatch Metrics as a Dimension to support multiple orgs per AWS account

lox:
We removed `BuildkiteOrgSlug` when we introduced the new Agent Metrics API. The idea was to reduce the number of required parameters, but we also removed the piece of data that was needed to support multiple Buildkite Orgs per AWS account, as otherwise the metrics are disambiguated only by queue. 

This adds back `BuildkiteOrgSlug` with a default of `` and then adds an `Org` dimension to the metrics we publish to Cloudwatch.

Closes https://github.com/buildkite/elastic-ci-stack-for-aws/issues/452.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -679,6 +679,7 @@ Resources:
             BUILDKITE_AGENT_EXPERIMENTS="${BuildkiteAgentExperiments}" \
             BUILDKITE_AGENT_RELEASE="${BuildkiteAgentRelease}" \
             BUILDKITE_QUEUE="${BuildkiteQueue}" \
+            BUILDKITE_ORG_SLUG="${BuildkiteOrgSlug}" \
             BUILDKITE_ELASTIC_BOOTSTRAP_SCRIPT="${BootstrapScriptUrl}" \
             BUILDKITE_AUTHORIZED_USERS_URL="${AuthorizedUsersUrl}" \
             BUILDKITE_ECR_POLICY=${ECRAccessPolicy} \
@@ -756,6 +757,9 @@ Resources:
         - Key: BuildkiteAgentRelease
           Value: !Ref BuildkiteAgentRelease
           PropagateAtLaunch: true
+        - Key: BuildkiteOrgSlug
+          Value: !Ref BuildkiteOrgSlug
+          PropagateAtLaunch: true
         - Key: BuildkiteQueue
           Value: !Ref BuildkiteQueue
           PropagateAtLaunch: true
@@ -765,7 +769,6 @@ Resources:
             Value: !Ref CostAllocationTagValue
             PropagateAtLaunch: true
           - !Ref "AWS::NoValue"
-
     CreationPolicy:
       ResourceSignal:
         Timeout: !Ref InstanceCreationTimeout
@@ -827,6 +830,8 @@ Resources:
       Dimensions:
         - Name: Queue
           Value: !Ref BuildkiteQueue
+        - Name: Org
+          Value: !Ref BuildkiteOrgSlug
       ComparisonOperator: GreaterThanThreshold
 
   AgentUtilizationAlarmLow:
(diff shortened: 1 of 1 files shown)


templates/aws-stack.yml
```diff
@@ -93,6 +94,11 @@ Parameters:
       - edge
     Default: "stable"
 
+  BuildkiteOrgSlug:
+    Description: Buildkite organization slug
```
toolmantim: This might be unclear as to why it's needed, or what it does? It doesn't need to actually match your Buildkite org slug does it, it just scopes the metrics? It'd be so nice if we could just extract it from the agent token somehow 🤔

toolmantim: Would it be better to have it blank, and optional, rather than the string "none"?

```suggestion
    Description: Optional Buildkite organization slug (e.g. "my-org"). This only needs to be set when running stacks for different Buildkite organisations in a single AWS account.
```

lox: I couldn't think of an easy way to set up the conditionals everywhere else to support an empty one. I guess we could just publish an `Org=` dimension, but that seemed weirder.

toolmantim: Where do people see the dimension?

lox:
@toolmantim so despite adding the new dimension to the metric, I realized that we still need to filter on that dimension, which means we need to specify it in the stack still.

toolmantim:
Okie dokie. We should update the org slug param description though. It's unclear what the "org slug" is, and that it doesn't actually even need to match the org slug (it's just a name spacing/scoping thing… it won't error if it's wrong, or the org name changes).

I think also we can pre-fill CloudFormation parameters from URL values now? We should update the in-app links so we can pre-fill the parameters for people.

Should I roll back the backend change?

lox:
> Okie dokie. We should update the org slug param description though. It's unclear what the "org slug" is, and that it doesn't actually even need to match the org slug (it's just a name spacing/scoping thing… it won't error if it's wrong, or the org name changes).

It does need to match the org-slug! That's the metric that will be set in the backend. This change still relies on getting the org-slug from the metrics API change you made. 

> I think also we can pre-fill CloudFormation parameters from URL values now? We should update the in-app links so we can pre-fill the parameters for people.

🤷🏼‍♂️

> Should I roll back the backend change?

Nah, let's roll with it. I think it will be a good building block.

toolmantim:
> It does need to match the org-slug! That's the metric that will be set in the backend. This change still relies on getting the org-slug from the metrics API change you made.

Oh, I thought it was passed through to the Lambda from the stack! Nevermind then.

lox:
> Oh, I thought it was passed through to the Lambda from the stack! Nevermind then.

It was originally, but I figured we'd just leave it with the metrics populating it. You do have a point that it means you need to get the org-slug correct, but that's probably a good idea anyway?

toolmantim:
I'd always err on "make it simpler for the users, even if it less-simple to implement" — but I don't know just how hard it is to require only the agent token.

One other question: what will happen for people that set up one stack, then need to set up a second one. Do they go and update the existing stack, add the org slug to scope it, apply (and hopefully the metrics work okay?)… then they create the second stack? Or do they just create the second stack (and have to remember to set the org scope?)

lox:
Darned if I can make it work anyway. 

> One other question: what will happen for people that set up one stack, then need to set up a second one. Do they go and update the existing stack, add the org slug to scope it, apply (and hopefully the metrics work okay?)… then they create the second stack? Or do they just create the second stack (and have to remember to set the org scope?)

They would need to update the first stack. The nice thing about always collecting the org in the metrics regardless of whether a stack filters on it is that the data is collected so that this update would work instantly.

lox:
@toolmantim this is working now, I'm pretty keen to merge it in as various folks have asked about it. Feels?

PR #334, opened 2017-08-27
Add configuration option to enable EBS volume encryption.

ineffyble:
Adds support for #318. So far not tested due to other issues.

Note that if set to true, the instance types available for use [will be limited](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSEncryption.html#EBSEncryption_supported_instances).

PR #337, opened 2017-09-02
Support ebs encryption

lox:
Merged #334 in here so I can run CI against it. Thanks @ineffyble!

PR #513, opened 2019-01-14
feat(encrypted-volume): add support for encrypted volumes. Fixes #318

### LINKED ISSUE #318: Support encrypted EBS volumes

mefellows:
I note that the PR in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/334 for related issue https://github.com/buildkite/elastic-ci-stack-for-aws/issues/318 is not in master and consequently hasn't been released.

I'm not sure if it disappeared as part of the amalgamation into a single template or because it didn't meet functional expectations.

I've reproduced the PR here in the hope we can get it in.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -41,6 +41,7 @@ Metadata:
         - AuthorizedUsersUrl
         - BootstrapScriptUrl
         - RootVolumeSize
+        - EncryptRootVolume
         - ManagedPolicyARN
         - InstanceRoleName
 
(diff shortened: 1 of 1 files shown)


mefellows:
Bugger! Do you know what it was, we could probably pick it up and try and fix it (it's causing us a config compliance issue).

ineffyble:
Thank you for picking this up after I dropped the ball on it.

lox:
Ah-hah, here is the context: https://github.com/buildkite/elastic-ci-stack-for-aws/pull/337

PR #517, opened 2019-01-18
Support encrypted EBS root volumes

### LINKED ISSUE #318: Support encrypted EBS volumes

lox:
Closes #318.


lox:
Argh. That was not quite what I intended to do.

PR #518, opened 2019-01-18
Support encrypted EBS root volumes (take 2)

### LINKED ISSUE #318: Support encrypted EBS volumes

lox:
Accidentally messed this up in #517, trying again. 

Closes #318.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -41,6 +41,7 @@ Metadata:
         - AuthorizedUsersUrl
         - BootstrapScriptUrl
         - RootVolumeSize
+        - EncryptRootVolume
         - ManagedPolicyARN
         - InstanceRoleName
 
@@ -220,6 +221,14 @@ Parameters:
     Type: Number
     Default: 250
     MinValue: 10
+
+  EncryptRootVolume:
+    Description: Encrypt instance root EBS volumes (will require a supported instance type)
+    Type: String
+    AllowedValues:
+      - true
+      - false
+    Default: "false"
 
   SecurityGroupId:
     Type: String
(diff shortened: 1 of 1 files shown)


lox:
I think perhaps we might need to provide a set of encrypted AMIs if we want this to work. I don't know how keen I am to do that.

lox:
> The process of creating an encrypted EBS boot volume begins with an existing AMI (either Linux or Windows). If you own the AMI, or if it is both public and free you can use it directly. Otherwise, you will need to launch the AMI, create an image from it, and then use that image to create the encrypted EBS boot volume (this applies, for example, to Windows AMIs). The resulting encrypted AMI will be private; you cannot share it with another AWS account.

Ah, right. Nope, we can't even provide them, they are per-account. For this to work, folks would need to copy the latest AMI into their account and encrypt it.

mefellows:
My thinking for this feature is that it's targeted for those customers who would likely harden their own AMIs for compliance/security reasons - either based off a public AMI or re-created from your [template](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/buildkite-ami.json). 

Does this sound reasonable or do you see obvious alternatives that we should pursue instead?

ineffyble:
@lox If my understanding is correct, in that circumstance the user just copies & encrypts the AMI, and provides it as the existing `ImageID`.

The encrypted flag does not need to be set, because the AMI/snapshot is encrypted, so the volume is too.

So it might just be a case of mentioning that in the docs, with no code change neccessary.

mefellows:
Thanks, this all actually makes a lot of sense, and we'll give it a go in the couple of days. Appreciate your responsiveness 👍

lox:
No problems!

PR #521, opened 2019-01-25
Prevent systemd from killing agent process group

### LINKED ISSUE #520: Agent doesn't terminate gracefully on scale-in
The elastic stack is supposed to wait for running jobs to finish before scale-in, if it can. Since the 4.x series, it appears that agents are being terminated as soon as the shutdown hook comes in.

lox:
The agent handles the termination of the bootstrap. This was preventing the agent from gracefully terminating.

Closes #520.


dhalperi:
🎊

PR #522, opened 2019-01-25
Expose AgentLifecycleTopic for programatic scaling

tduffield:
Expose the `AgentLifecycleTopic` name so that users can use utilities like Terraform to easily create custom integrations to supplement the Cloudwatch Alarms already provided by the ElasticStack.

Specifically, we want to be able to write custom alarms that can enforce additional scaling events outside of just the default alarms. 

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -332,6 +332,9 @@ Outputs:
   InstanceRoleName:
     Value: !Ref IAMRole
 
+  AgentLifecycleTopic:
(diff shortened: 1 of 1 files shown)


tduffield:
An example of this that we're doing internally is we use the CloudFormation Terraform resource to manage all our stacks, and we create custom alarms that more efficiently scale down our ASGs based on the specific needs of each of the queues.

lox:
Works for me!

PR #523, opened 2019-01-25
Add new "TerminateInstanceAfterJob" configuration

tduffield:
This new setting will allow you to stop (and terminate) an instance after it has completed a job.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Features:
 - [What Type of Builds Does This Support?](#what-type-of-builds-does-this-support)
 - [Multiple Instances of the Stack](#multiple-instances-of-the-stack)
 - [Autoscaling](#autoscaling)
+- [Terminating the instance after job is complete](#terminating-the-instance-after-job-is-complete)
 - [Docker Registry Support](#docker-registry-support)
 - [Versions](#versions)
 - [Updating Your Stack](#updating-your-stack)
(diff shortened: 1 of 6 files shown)


lox:
Interesting! What’s the use case for this? My primary concern would be the 4 minute boot time of instances.

Might ECS be a better fit?

tduffield:
FWIW this is a pattern we've been using successfully for about 6 months internally. We're just getting around to contributing it back upstream :)

lox:
Makes sense! What would you think of instead of an "instance agent mode" this was just a boolean option along the lines of "terminate instance after job", similar to `buildkite-agent start --disconnect-after-job`. I think that might avoid a lot of the cognitive overhead that modes bring.

tduffield:
Sure. Works for me. I'll work up something tomorrow.

tduffield:
@lox okay, updated the PR with your suggestions.

lox:
This is looking good, I really like the README description. One last request, could we use a drop-in unit for overriding adding just the `ExecStopPost` lines?

**/etc/systemd/system/buildkite-agent@.service.d/10-power-off-stop.conf**

```
[Service]
ExecStopPost=/usr/local/bin/mark-asg-unhealthy
ExecStopPost=/bin/sudo poweroff
```

Keeps merge conflicts to a minimum.

tduffield:
Alrighty. With that suggestion I was able to keep most of the changes into the boot script, keeping the AMI setup mostly the same. That's a good pattern overall I think. Thanks for that!

tduffield:
I also made the disconnect timeout configurable, and documented it in the README.

README.md
```diff
@@ -141,6 +141,19 @@ This means you can scale down to zero when idle, which means you can use larger
...
 Metrics are collected with a Lambda function, polling every minute.
 
+## Terminating the instance after job is complete
+
+You may set `BuildkiteTerminateInstanceAfterJob` to `true` to force the instance to terminate after it completes a job. Setting this value to `true` tells the stack to apply the following configurations:
+
+1. Sets `disconnect-after-job` and `disconnect-after-job-timeout` in the `buildkite-agent.cfg` file.
+2. Adds `ExecStopPost` steps to the agent's systemd service to mark the instance as unhealthy and to stop the instance.
+
+While not enforced, it is highly recommended if you also set to set your `AgentsPerInstance` value to `1`.
```
jeremiahsnapp: @tduffield looks like this needs some cleaning up. Was it supposed to be this?

While not enforced, it is highly recommended you also set the `AgentsPerInstance` value to `1`.

tduffield: Thanks!

packer/conf/buildkite-agent/scripts/mark-asg-unhealthy
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -euo pipefail
+
+instance_id=$(curl -fsSL http://169.254.169.254/latest/meta-data/instance-id)
+region=$(curl -fsSL http://169.254.169.254/latest/meta-data/placement/availability-zone | head -c -1)
+
+aws autoscaling set-instance-health --instance-id "$instance_id" --region "$region" --health-status "Unhealthy"
```
tekumara: Would love something very similar to this. Our use-case is running some data crunching on very large instances types, and to minimise costs we would like the instances to terminate a soon as the job completes.

One difference in our use-case is that we don't want the terminated instance to be replaced immediately with a fresh one (unless there is a scheduled job).

Would you be willing to accommodate this in your PR?

I think the change would be, instead of marking the instance unhealthy, using [terminate-instance-in-auto-scaling-group](https://docs.aws.amazon.com/cli/latest/reference/autoscaling/terminate-instance-in-auto-scaling-group.html) with the `--should-decrement-desired-capacity` flag, which could be configurable.

Alternatively happy to submit this change myself as a new PR.

tduffield: @tekumara okay, I pushed up a potential implementation in 724d56c. Let me know if you think this would work for you.

tekumara: Fantastic, yes I think that will do the trick! Thankyou 🙏

PR #527, opened 2019-01-30
Allow users to specify additional sudo permissions

tduffield:
Specify a comma separated list into `BuildkiteAdditionalSudoPermissions`

This allows users to easily extend which commands buildkite-agent can run using sudo. The setting can also be set to `ALL` as an escape hatch for those who want to run a super-privileged agent (though this is not recommended).

### DIFF
--- a/packer/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/conf/bin/bk-install-elastic-stack.sh
@@ -70,6 +70,11 @@ if [[ "${BUILDKITE_AGENT_RELEASE}" == "edge" ]] ; then
 	buildkite-agent-edge --version
 fi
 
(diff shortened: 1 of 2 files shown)


tduffield:
@lox so I wanted to know what you thought of this PR? This, along with the terminate PR, allowed us to create a 2x2 matrix of queues that pretty much met all of our needs: 

* default
* default-privieleged (sudo access, no docker namespacing)
* single-use (terminates after 1 job)
* single-use-privileged (sudo access, no docker namespacing, terminates after 1 job). 

If we can get this merged in the upstream, we'd be able to use the default AMI's for 100% of our workload! The only thing left would be to get our windows AMI committed upstream (which we are currently working towards).

lox:
I'm ok with it! Thanks for getting this into shape and contributing back!

PR #529, opened 2019-02-07
Add an experimental lambda scaler

lox:
This adds a lambda that handles the metrics polling and setting the `DesiredCount` on the ASG. 

This polls the Buildkite API more frequently and bypasses the AWS autoscaling mechanisms. My current tests are showing instances booting and picking up work in about 60 seconds, which is cool. 

Still TODO is cooldown of some sort, probably just for scaling in.

### DIFF
--- a/.buildkite/steps/test.sh
+++ b/.buildkite/steps/test.sh
@@ -64,6 +64,10 @@ cat << EOF > config.json
   {
     "ParameterKey": "EnableDockerUserNamespaceRemap",
     "ParameterValue": "true"
+  },
+  {
+    "ParameterKey": "EnableExperimentalLambdaBasedAutoscaling",
+    "ParameterValue": "true"
   }
 ]
 EOF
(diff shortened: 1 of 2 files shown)


tduffield:
I'm not super familiar with Golang, but am I reading this right...that we'd be deciding scaling based on a min/max of agents, not instances? Because if that's the case, that is great!

lox:
Correct @tduffield!

lox:
@toolmantim would love a review! I think this should have pretty minimal impact on the default autoscaling, and it lets us experiment with the new scaler.

toolmantim:
Awesome! What’s the best way to review this you think @lox? Spin up a stack with the YML from the build?

lox:
🤔

Yeah, I think it might be. Alternately, we could merge it and then experiment with our internal runner stacks. Perhaps we need a way to easily update just the version of the scaler that is running.

toolmantim:
I won’t have a chance to spin anything up until next week sorry. But I can do, if you need!

toolmantim:
If you merge, then at least the stack has everything hooked up, and we can update the scaler Lambda on a live stack manually?

ecv:
This reduced my spin-up wait from just-under three minutes to twenty-four (24) seconds flat. 

Psyched for the cooldown option; mad at Amazon for EC2 ASG being so doggone slow; all praise to @lox

PR #533, opened 2019-02-12
Make sure we reload the systemd unit files

tduffield:
This is a bug affected the new terminate functionality in 4.1.0. 

If we do not reload the unit files, the new power-off.conf file gets ignored.


tduffield:
Hold off on merging this PR. There are a few more tweaks required.

tduffield:
@lox okay, I've tested these changes locally and can confirm the auto-termination is now working as expected. Apologies for not catching this sooner.

lox:
No problems, thanks for the PR!

PR #534, opened 2019-02-13
Allow users to configure the root block device

tduffield:
If someone is not using the default AMI, it's possible their AMI might use a different root block device name. `/dev/sda1` for example. While we're at it, also allow them to configure the volume type. 


jeremiahsnapp:
Hi @lox, do you have any objection to merging this? I'd like to get rid of our forked version.

lox:
Seems entirely reasonable! Sorry, I've been travelling.

jeremiahsnapp:
Thanks a lot @lox!

PR #535, opened 2019-02-13
Add helpers to Makefile for building packer image

tduffield:
This adds some simple helpers to the Makefile to make building custom AMIs for dev a little bit easier. It also ensures that we're building against the latest Amazon 2 AMI by default.


lox:
👌🏻

PR #539, opened 2019-02-26
Set the priority of the agent to its instance integer

### LINKED ISSUE #540: [feature request] Encourage "breadth first" scheduling
We've occasionally observed issues where a particular AWS instance running multiple buildkite-agents has many jobs running while other instances have very few jobs running.  

It might be nice to encourage jobs to be spread breadth-first across instances.  One way to do this may be to modify the service template: 

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/conf/buildkite-agent/systemd/buildkite-agent%40.service

to set `BUILDKITE_AGENT_PRIORITY` (documented here https://buildkite.com/docs/agent/v3/prioritization).  If that variable works as advertised, then we could start each additional agent on a given instance with a lower (or higher) priority.

tduffield:
This will ensure that we optimize for leveraging the instances that are available to us, rather than unintentionally packing all jobs onto a single instance.

Fixes #540 

### DIFF
--- a/packer/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/conf/bin/bk-install-elastic-stack.sh
@@ -101,6 +101,7 @@ hooks-path=/etc/buildkite-agent/hooks
 build-path=/var/lib/buildkite-agent/builds
 plugins-path=/var/lib/buildkite-agent/plugins
 experiment="${BUILDKITE_AGENT_EXPERIMENTS}"
+priority=%n
 EOF
 
 if [[ "${BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB:-false}" == "true" ]] ; then


tduffield:
@lox we're running into a significant amount of grief in our pipelines that this setting would alleviate. How do we feel about this? Would love to be able to get this merged if possible.

lox:
Sorry for the delay, I should ACK these initially and let you know I'm thinking them through. Basically what this accomplishes is a "spread" scheduling mode. I've been wondering if a) whether this should just be done server side and b) what impact this has on autoscaling where you have multiple agents per node. It makes it harder to scale instances in when jobs aren't packed to a node. 

That said, after thinking those things, I'm inclined to just ship this and see how it goes. Thanks @tduffield!

lox:
Btw, I'd love to hear a bit more about what problems you are seeing. Presumably jobs scheduling on the one host when there are plenty of spare instances?

tduffield:
@lox that's the gist of it yes. Our issue is textbook resource contention. Some steps take very few resources. Others take much more. I don't know the mechanics of the scheduling, but we always seemed to have all the resource heavy jobs scheduled on a single instance. 

In an ideal world, we could define our resource requirements and buildkite would take that into account when scheduling jobs. However, the `priority` setting was our best best to at least _try_ and spread out the load.

PR #541, opened 2019-02-26
Write cloudwatch metrics from lambda scaler

lox:
This updates the https://github.com/buildkite/buildkite-agent-scaler version to the latest which includes writing `JobScheduledCount` metrics to Cloudwatch Metrics.


lox:
A huge issue we discovered with the scaler is that it bypasses lifecycle hooks entirely. It's kind of taken the wind out of my sails a bit.

lox:
@tduffield so this now uses the lambda for scale-up and the new `--disconnect-after-idle-timeout` feature in the agent, paired with the "terminate and decrement desired count on the asg". Seems to work nicely!

lox:
Found a bug here where desired count isn't being decremented when instances scale down.

PR #542, opened 2019-02-26
adding support for permission boundaries into the stack

tnh:
Thoughts on this?

This adds the concept of permission boundaries

https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_boundaries.html

There are two key parts to permission boundaries. The boundary ARN that sits around the role. Generally you also want to namespace the role itself too. 

If I'm going to have buildkite create stuff in IAM using a ManagedPolicy, I might also want to accompany that with a boundary too.

eg https://youtu.be/YQsK4MtsELU?t=2273

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -49,7 +49,9 @@ Metadata:
         - RootVolumeName
         - RootVolumeType
         - ManagedPolicyARN
+        - ManagedPolicyPermissionBoundryARN
         - InstanceRoleName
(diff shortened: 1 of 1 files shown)


lox:
Interesting idea! I hadn't heard of permission boundaries before. Could you explain a scenario where this would be useful? 🤔

lox:
Going to close this one as we haven't heard back! Happy to revisit if there is a strong usecase.

PR #545, opened 2019-03-11
Restart agent on SIGPIPE (journald restart)

lox:
We had an issue where the agent failed to restart on SIGPIPE, which turns out to be a thing when systemd and journald restart. 

Similar to https://github.com/influxdata/telegraf/pull/2716.


lox:
RestartForceExitStatus= option was not added until systemd 215, but from what I can see the version is 219 in the latest Amazon Linux 2. Fingers crossed!

PR #547, opened 2019-03-16
Bump agent to 3.9.1

tekumara:
This is particularly because I'm interested in https://github.com/buildkite/agent/pull/931


jeffora:
This would be great because currently `skip-checkout: true` in the latest docker-compose plugin causes agents prior to `3.9.0` to break (fixed in https://github.com/buildkite/agent/pull/909)

lox:
Thanks folks!

Issue #436, opened 2018-06-15
Faster scale up from 0


pauldraper:
Right now, it seems takes several minutes for the stack to even decide to launch the first EC2 instance.

Having some slack makes sense, but if there are literally 0 instances....it should start one immediately.

lox:
Thanks @pauldraper, we REALLY want this too. It's amazingly difficult to get a general solution in place unfortunately, it's something I've been working on for years (depressingly). Stepped scaling (see https://github.com/buildkite/elastic-ci-stack-for-aws/issues/93) is hard to have custom min/max sizes for and percentage based scaling is really difficult to handle stacks that scale down to zero (which is really important for cost-saving). 

The plan currently is to use a custom Lambda to do the scaling. Keep an eye on https://github.com/buildkite/elastic-ci-stack-for-aws/pull/426.

pauldraper:
@lox , interesting. Will do!

lox:
Fixed via #554.

Issue #322, opened 2017-08-10
Scale up/down based on utilization instead of UnfinishedJobsCount


joshma:
(I left a slack message, but wanted to surface here as well.)

Can we consider switching the scale up/down metric based on utilization, e.g. percentage or count of idle agents? The main issue with `UnfinishedJobsCount == 0` is that it only triggers if agents complete **all jobs**, so if our cluster is always running 1 job, then our agents never scale down.

On the other hand, scale up happens if `ScheduledJobsCount > 0`, so during the course of a day, load picks up, jobs are scheduled and not processed quickly enough, and we hit the max cluster size. Jobs get completed, and the queue is empty at times, but instances are often running at least one job, so the cluster never scales down.

If the answer is "add your own alarms", that's fine too, it just didn't seem kosher given we're managing the rest via CloudFormation.

lox:
The difficult part is that there are often more agents then there are hosts, so idle agents needs to be divided by agents per host to be useful. 

I'd absolutely be interested in tweaking this so it's more useful, the other option I've been thinking about is the new Target tracking for Autoscaling: https://aws.amazon.com/blogs/aws/new-target-tracking-policies-for-ec2-auto-scaling/. 

We could possibly add in a scaling rule for CPU utilization by host too?

joshma:
Hm interesting - looks like target tracking is limited to ALB requests, CPU, and network in/out. CPU utilization by host (through target tracking or old school) could work well for us. I'm assuming most folks' jobs are CPU-bound? It might be weird if there's a cluster e.g. waiting for cloudformation templates to apply and that causes scaledown.

Is it actually true that we'd need to divide agents by host? Haven't thought about this too carefully, but terminating an idle host would lower the total number of agents, regardless of how many agents are on it, thus decreasing the % of idle agents (for a fixed number of active agents).

Is there a way for us to test this? I'm finding that cloudwatch doesn't let us do an alarm on `X / Y`, could the lambda function be updated to emit `BusyAgentCount / TotalAgentCount`?

joshma:
For what it's worth, I'm currently using the target tracking on CPU - set it to 50% arbitrarily:

![dd](https://user-images.githubusercontent.com/64739/30242780-ab14a8a4-9551-11e7-8537-eedea9f408af.png)

[...]

PR #554, opened 2019-03-26
Remove alarm based autoscaling

### LINKED ISSUE #93: Stepped scaling support
For our application our main app requires 27 agents, but other pipelines only require 1 or 2. The only way we can make the stack responsive at the start of the day is to make the scaling policy +27 -27 scale out/in, but it's a waste of $ to do that for just a couple of scheduled jobs.

It'd be much better if we could use stepped scaling, and have it scale in/out in smaller increments

lox:
This removes the old alarm-based autoscaling in favour of lambda-based scale-out and per-host based idle scale-in.

It's radically faster. I'm seeing 50-75s from the time a job is queued to the time an agent is available on a new host! 

This will likely need to be released in v5.0.0. 

Closes #93, #322 and #436.

### DIFF
--- a/.buildkite/steps/test.sh
+++ b/.buildkite/steps/test.sh
@@ -64,10 +64,6 @@ cat << EOF > config.json
   {
     "ParameterKey": "EnableDockerUserNamespaceRemap",
     "ParameterValue": "true"
-  },
-  {
-    "ParameterKey": "EnableExperimentalLambdaBasedAutoscaling",
-    "ParameterValue": "true"
   }
 ]
 EOF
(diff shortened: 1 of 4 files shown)


lox:
@tekumara @tduffield this change makes `BuildkiteTerminateInstanceAfterJobDecreaseDesiredCapacity` the default behaviour if autoscaling is enabled or `BuildkiteTerminateInstanceAfterJob` is set. Any objections? Given the behaviour of the new autoscaler, if an instance is needed then `DesiredCapacity` will just get incremented anyway.

tekumara:
@lox 👏nice! No objections from me. Just out of curiosity, what makes it so much faster?

lox:
@tekumara a few things. The old metrics lambda polled every 60 seconds which was combined with a 120 second delay before AWS Autoscaling alarms fired. Beyond that, the new lambda approach sets the number of desired instances right away vs slowly scaling up to what you need which suits CI based workflows better.

PR #560, opened 2019-04-05
Install pigz for parallel decompression in docker pull

lox:
Docker added some support for slightly improving `docker pull` performance if `unpigz` is available, see the notes here: https://github.com/moby/moby/issues/37957#issuecomment-426633293


toolmantim:
🎉

PR #571, opened 2019-04-29
Use the new dag experiment for better parallel ordering

lox:
🤫


tduffield:
<whisper>Does DAG support an array of strings or just strings? :wink: </whisper>

lox:
This seems to save considerable time.

PR #575, opened 2019-05-05
Remove AWS Autoscaling in favor of buildkite-agent-scaler

lox:
Remove AWS autoscaling in favour of lambda scaler. 

This will the default in 5.0.0.


jeremiahsnapp:
@lox doesn't this change allow us to completely remove lifecycled?

lox:
It's nice to keep it so that we can gracefully handle spot instance termination, I thought.

PR #576, opened 2019-05-05
Read a Parameter Store key for Agent Token

lox:
Replaces `BuildkiteAgentToken` with `BuildkiteAgentTokenParameterStoreKey` 🤔

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -99,11 +99,11 @@ Parameters:
       - edge
     Default: "stable"
 
-  BuildkiteAgentToken:
-    Description: Buildkite agent registration token
-    Type: String
-    NoEcho: true
+  BuildkiteAgentTokenParameterStoreKey:
+    Description: The SSM Parameter Store path to the Agent Token
+    Type: AWS::SSM::Parameter::Name
     MinLength: 1
+    Default: "/buildkite/agent_token"
 
   BuildkiteAgentTags:
     Description: Additional tags seperated by commas to provide to the agent. E.g os=linux,llamas=always


lucas044:
I know this is implemented in the `buildkite-agent-scaler` but what is the rationale for using SSM over Secretsmanager?

lox:
That is a very good question @lucas044 and one I've gone back and forth on. I've had this discussion with a lot of folks at different sized tech companies and the consensus seems to be that SSM Parameter Store is far and away more common and widely adopted than SM. 

I tend to agree, SM seems like a thing AWS built because they missed the boat on pricing Parameter Store well. At 40c a secret per month, it ends up getting pricey for small teams, especially if you want secret-per-env.

The final straw for me is that you can access SM from Parameter Store: https://docs.aws.amazon.com/systems-manager/latest/userguide/integration-ps-secretsmanager.html

Seems like Parameter Store is the best of both worlds given that?

Can you think of any compelling reasons to force folks onto Secrets Manager?

lox:
FWIW, I don't think I will proceed with this PR, but still want to find a way to get the key into Lambda without passing it via plain-text env.

lucas044:
Probably should have stipulated that I was not married to either option. Was just curious as to how the decision was being made as I am in the same boat with going back and forth to figure out what the best coarse of action is.

Sounds like all the things you've mentioned are rational and make complete sense.

I have a feeling that AWS will eventually make Secretsmanager the default tool seeing as all key/values within the secrets manager are "secrets" and will all be encrypted. Where SSM `can` store plain text if required. But with its current maturity I would definitely agree with this decision.

Thanks for the prompt response though!

lox:
Yeah, it's currently an extremely head-scratching situation for tooling providers. If I were AWS, I'd massively reduce the pricing on SM and bump up the hidden rate limits on Parameter Store to annoy folks enough to move over. 🤷🏼‍♂️

lucas044:
Sorry just a few more things to note, what are your thoughts on some of these items and whether it can or would affect some clients (disregarding the obvious pricing model from AWS):

- SM supports cross region access (probably not relevant?)
- SM abstracts KMS, so SSM you will still have to manage access at the key level

lox:
* SM supports cross region access (probably not relevant?)

🤷🏼‍♂️ Does that mean Parameter Store is region specific?

* SM abstracts KMS, so SSM you will still have to manage access at the key level

I'd have to look into that one. Abstracting KMS isn't high up on my list of value props probably. Is that something people want?

jufemaiz:
I’m pretty sure it is @lox

lucas044:
- Does that mean Parmater Store is region specific?

Yes it does. So if you wanted a BK stack to span across multiple regions, you'd have to figure out a way to share the CMK/decrypt and re-encrypt etc...can get ugly

- Is that something people want?
I find at larger enterprises this is the direction they are wanting to head. Remove any of those overheads of complexities.

I believe SM can handle better in a large scale environment, where SSM is better in a smaller scale where they don't have a large secrets base. From my experience anyway.

lox:
All good points! Given Parameter Store can access Secrets Manager secrets though via the `/aws/reference/secretsmanager/CFCreds1` paths, does that mitigate those concerns?

lox:
Closing in favour of https://github.com/buildkite/elastic-ci-stack-for-aws/pull/601

PR #577, opened 2019-05-08
Use an SSM Parameter for BuildkiteAgentToken

lox:
Another take on https://github.com/buildkite/elastic-ci-stack-for-aws/pull/576

Feels @lucas044?

### DIFF
--- a/.buildkite/steps/launch.sh
+++ b/.buildkite/steps/launch.sh
@@ -5,6 +5,7 @@ set -eu
 os="${1:-linux}"
 stack_name="buildkite-aws-stack-test-${os}-${BUILDKITE_BUILD_NUMBER}"
 stack_queue_name="testqueue-${os}-${BUILDKITE_BUILD_NUMBER}"
+ssm_secret_key="/buildkite/aws-stack-test/${stack_name}"
 
 # download parfait binary
 wget -N https://github.com/lox/parfait/releases/download/v1.1.3/parfait_linux_amd64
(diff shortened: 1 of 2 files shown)


holmesjr:
I know you've had this discussion, but can you consider using secrets manager?

I ask because a reference to SSM in this way uses a plain-text string. Resolving a secrets manager value allows for KMS:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html#dynamic-references-secretsmanager

lox:
By the same token, we could use ssm secure if we used Dynamic References @holmesjr.

I lean more towards this approach as it's backwards compatible and the UX is slightly easier for first-time users, but it would be nice to have a way to use encrypted storage.

lox:
Perhaps the approach in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/576 was better after all. 😬

lucas044:
I have no real opinion about either option at this point.
Weighted to SM over SSM but I'm more weighted towards not having my secret exposed on the lambda environment variable more :joy:

lox:
I am back to being paralysed by indecision 😅

lox:
Yeah, going back to #576, at least that way we have options.

lox:
Re-considering.

holmesjr:
> By the same token, we could use ssm secure if we used Dynamic References @holmesjr.
> 

The docs point out the difference:

> The secretsmanager dynamic reference can be used in all resource properties. Using the secretsmanager dynamic reference guarantees that neither Secrets Manager nor CloudFormation logs or persists any resolved secret value.

vs

> Currently, secure string parameters can only be used for resource properties that support the ssm-secure dynamic reference pattern. (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html#template-parameters-dynamic-patterns-resources)

lox:
Gotcha @holmesjr. There doesn't seem to be any way of dynamically creating those dynamic references though, for instance from a path in a parameter template.

lox:
I think it's just too much of a bummer that SecureString isn't supported.

PR #583, opened 2019-05-10
Add python3 and future lib to allow prepping for Python2 EOL

GreyKn:
With Python2 aiming to EOL on Jan 1, 2020 (https://github.com/python/devguide/pull/344), we need to start prepping for Python3 conversion.

The [future module](https://python-future.org/) allows uplifting Python2 code to be mostly compatible with Python3, allowing for incremental conversions.


lox:
Yeah, I guess this probably makes sense!

PR #584, opened 2019-05-14
Fix cloudformation ImageId regression

jeremiahsnapp:
The following commit removed the ability to use the
value set by the ImageId parameter.

https://github.com/buildkite/elastic-ci-stack-for-aws/commit/7d2ec14d1dc87ae32b98e7b75905e3d1fe6c8d51

Signed-off-by: Jeremiah Snapp <jeremiah@chef.io>


lox:
Good find @jeremiahsnapp!

PR #585, opened 2019-05-14
Adds AllowedPattern to ImageId

kyledecot:
Adds `AllowedPattern` regex to `ImageId`. I mistakenly passed the image name and had to wait for the stack to rollback to see that I had passed the wrong value. This would have given me an error more quickly.


lox:
Makes sense, I wonder if this still will allow the empty value though?

PR #586, opened 2019-05-17
Decrease capacity after terminating instance

jeremiahsnapp:
* Bumps buildkite-agent to version 3.11.5.
* Makes terminate-instance script always decrease capacity.
* Removes `BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB_TIMEOUT` parameter since this is best handled by `ScaleInIdlePeriod`.
* Using the new scaler by default allows the bk-install-elastic-stack scripts to be refactored to have a cleaner way of defining `buildkite-agent.cfg` and also allows the buildkite-agent service to run `post stop` commands by default.
* Updates README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,16 +144,9 @@ Metrics are collected with a Lambda function, polling every minute.
 
 ## Terminating the instance after job is complete
 
-You may set `BuildkiteTerminateInstanceAfterJob` to `true` to force the instance to terminate after it completes a job. Setting this value to `true` tells the stack to apply the following configurations:
-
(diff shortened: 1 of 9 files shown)


jeremiahsnapp:
@lox I think this should fix the broken decrease capacity issue we talked about as well as clean up the way some things are managed since we know we're using the new scaler. Please make sure I didn't mess anything up though. 😄

lox:
Thanks @jeremiahsnapp! I had a lot of these fixes on my list to do! 🙇🏼‍♂️

PR #589, opened 2019-05-22
Use a LaunchTemplate instead of a LaunchConfiguration

lox:
Apparently using LaunchTemplates solves the tag propagation problem that has resulted in the emails folks have been getting about IAM Authentication and Cross-stack Requests. 

It also opens up some nice possibilities for the easier introduction of Spot Fleet down the track. 

Tried this before in #424, but it wasn't possible then because there wasn't a way to specify a spot bid, but that now is possible in [InstanceMarketOptions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-instancemarketoptions.html).

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -693,102 +693,134 @@ Resources:
       Roles:
         - !Ref IAMRole
 
-  AgentLaunchConfiguration:
-    Type: AWS::AutoScaling::LaunchConfiguration
+  AgentLaunchTemplate:
+    Type: "AWS::EC2::LaunchTemplate"
     Properties:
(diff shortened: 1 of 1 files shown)


patrobinson:
❤️ This makes is possible to do "spot fleets" without spot fleets, we just need to be able to specify a [mixed instances policy](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-mixedinstancespolicy), which gives us the capability to launch n instance types in m availability zones.

PR #595, opened 2019-06-03
Add default InstanceCreationTimeout value for windows instances

jeremiahsnapp:
In my testing the stack's Windows Server 2019 instances take approximately 6 minutes to fully come online so we should make sure the `InstanceCreationTimeout` value is sufficiently large enough to accommodate the longer startup time.


lox:
Thanks @jeremiahsnapp!

PR #598, opened 2019-06-14
Open up RDP port on Windows instances

tduffield:
When creating the default security group, make sure we open up RDP/3389 so that people can connect to the instance since (as far as I can tell) we do not install a SSH server on the Windows AMI.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -432,6 +432,15 @@ Conditions:
           - { Condition: HasKeyName }
           - !Not [ !Equals [ !Ref AuthorizedUsersUrl, "" ] ]
 
+    EnableRdpIngress:
+      !And
+        - { Condition : CreateSecurityGroup }
+        - { Condition : UseWindowsAgents }
+        # Enable ingress if a key can be specified another way
+        - !Or
+          - { Condition: HasKeyName }
+          - !Not [ !Equals [ !Ref AuthorizedUsersUrl, "" ] ]
+
     # Whether or not there's any managed polices to attach
     HasManagedPolicies:
       !Or [ { Condition: UseManagedPolicyARN }, { Condition: UseECR } ]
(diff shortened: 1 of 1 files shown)


lox:
I'm slightly nervous about making the default public IP's and exposing RDP, I trust it considerably less than OpenSSH. 

Thoughts @tduffield?

tduffield:
@lox I mean...so do I. But it seems weird to have the "default" behavior for linux to be "you can access it" and the default behavior for windows to be "you cannot access it." I wonder if the "right" answer is to not allow _any_ access to all. It's not technically needed by any part of the native buildkite system (afaik), and is really up to the discretion of the infra maintainers.

lox:
>  But it seems weird to have the "default" behavior for linux to be "you can access it" and the default behavior for windows to be "you cannot access it." 

Totally. But I have 20 years of experience with SSH and trust it where as I have about 2 with Windows RDP and it makes me nervous 😅

> I wonder if the "right" answer is to not allow any access to all.

That is the direction I'd like to go in. What if we got Amazon System Manager set up on the instances and then folks could use that and something like https://aws.amazon.com/blogs/mt/smart-rdp-and-ssh-remediation-with-aws-systems-manager-automation-api-actions/ to setup SSH or RDP access for ad-hoc debugging if they required?

I guess we'd still need to solve the security group firewall issues.

lox:
How about we add a flag for customizing the CIDR of SSH/RDP and perhaps a parameter that is required to turn on SSH/RDP access which defaults to off?

lucas044:
SSM feedback:
I'm currently using SSM for SSH/RDP (only via powershell) access and disabled the security group for SSH access. This is now the defacto.

You definitely lose some functionality in a shell or a session manager, but the ease of authentication via AWS rather than a SSH key or password being thrown around swings this entire thing in the favor of SSM. Also no need for ports to be opened, so can trash the entire security group too :)

lox:
Nice @lucas044, is that in your terraform port?

lucas044:
Yeah correct!

tduffield:
I like the SSM stuff. If people want SSH/RDP they can provide their own security groups easily enough.

tduffield:
Closing this PR and moving the discussion over to https://github.com/buildkite/elastic-ci-stack-for-aws/issues/620.

PR #600, opened 2019-06-17
Run the bootstrap script AFTER the buildkite-agent user is created

tduffield:
This allows end-users to reference the buildkite-agent user in their bootscript to handle things like setting file/folder permissions.

My exact use case is that I install additional standalone binaries and folders that I need the buildkite-agent user to have access to even if they are not in the Administrators group. 


lox:
Makes sense!

PR #602, opened 2019-07-03
Add correct cost tags to S3

hawkowl:
This PR:

- Adds the cost allocation tags to S3 buckets


tduffield:
Whether or not the monitoring is enabled should be an option. We (personally) want it on :)

lox:
Thanks for this @hawkowl, would be great to see this as two PR's, the first to fix the S3 cost tags and then an optional param for disabling instance detailed monitoring? Would be happy to have the latter default to off.

hawkowl:
@lox Hi, this one is now just the correction of the S3 cost tags.

lox:
Many thanks!

PR #603, opened 2019-07-18
Allow encryption for EBS volumes

underscorebrody:
This PR:
- Adds a parameter to enable encryption on the EBS volumes created for the buildkite agents
- Adds a parameter to set the KMS Key ID to be used for encrypting the EBS volumes

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -263,7 +263,7 @@ Parameters:
     ]
 
   RootVolumeKmsKeyId:
-    Description: Optional - ARN of the AWS Key Management Service CMK used for encryption.\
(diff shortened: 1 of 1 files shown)


lox:
Historical context: https://github.com/buildkite/elastic-ci-stack-for-aws/pull/518

I dimmly recall reading that the constraint about Public AMI's and encryption had changed recently. Did this work for you @underscorebrody?

underscorebrody:
Yup you're right, this approach isn't going to work. May try to come at this another way, but closing for now.

ckornacker:
This is working now (for Linux AMIs anyway). I'm using it to harden my Buildkite agents and was just about to open PR before I found this one. Are you OK to re-open this PR instead?

PR #606, opened 2019-07-25
bump docker version to 19.03.0

dreyks:
is there a way I can test this locally?


lox:
> is there a way I can test this locally?

I'm afraid not yet, we should get public builds working soon 😓

PR #607, opened 2019-07-25
Update ArtifactBucketPolicy to match docs

gough:
This pull request updates the policy to match what is suggested in the documentation (linked below). 

In our testing, the default policy also did not allow uploading of artifacts despite what the policy seemed to allow. However, this could be due to another issue in our configuration.

https://buildkite.com/docs/agent/v3/cli-artifact#using-your-own-private-aws-s3-bucket

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -702,9 +702,14 @@ Resources:
         Statement:
           - Effect: Allow
             Action:
-              - s3:Put*
-              - s3:List*
-              - s3:Get*
+              - s3:GetObject
+              - s3:GetObjectAcl
+              - s3:GetObjectVersion
+              - s3:GetObjectVersionAcl
+              - s3:ListBucket
+              - s3:PutObject
+              - s3:PutObjectAcl
+              - s3:PutObjectVersionAcl
             Resource:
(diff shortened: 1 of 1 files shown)


kevinprince:
It may also be wise to update the documentation to take into account that AWS accounts can now have global S3 public access restrictions which cause a 403 regardless of the IAM permissions and you need to turn on the private ACL.

lox:
Thanks for this @gough! 

Worth noting, I don't think we actually set `BUILDKITE_ARTIFACT_UPLOAD_DESTINATION` anywhere in the elastic stack 🤔

lox:
Let's get this merged and then look at the follow on changes.

jufemaiz:
Can confirm this @lox.

Oddly I'm not seeing my `.buildkite/hooks/environment` script running anywhere on the CI stack o.O

jufemaiz:
Actually digging into this more, the end point (for example, bucket) and the path are not separate for BK.

It would be nice for an agent environment hook to set an ENV, such as `BUILDKITE_ARTIFACT_S3_BUCKET`, for use by the repositories to create the `BUILDKITE_ARTIFACT_UPLOAD_DESTINATION`

PR #616, opened 2019-08-06
Add cloudformation output exports

### LINKED ISSUE #615: Export more stack outputs
I have a few usecases where I want to refer to resources created by this stack in other places or stacks however the parameters I want are not exported.

I'd like to add extra outputs for these if you are willing?

jradtilbrook:
This resolves #615

I've only added the VPC and export names for the existing ones so they can be referenced in other places. Let me know if this is okay or any changes.

To clarify the exact use-case; we would like to run 2 of these stacks with different sized instances but want to share a VPC and role created by one, with the other without having to create that role externally and pass it into both.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -376,19 +376,32 @@ Parameters:
     Default: "false"
 
 Outputs:
+  VpcId:
+    Value: !Ref Vpc
+    Export:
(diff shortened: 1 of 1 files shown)


lox:
Seems legit! Thanks for your time and thought as always @jradtilbrook!

jradtilbrook:
No worries! Thanks for accepting it

PR #618, opened 2019-08-07
FOR DISCUSSION: Copy and encrypt buildkite agent AMI

underscorebrody:
These changes update the aws stack to allow copying the buildkite agent AMI using a lambda and encrypting it on the consumer's AWS account using a provided KMS key.

This PR includes the JS code (in `temp-ami-copy`) for a lambda that would likely need to be made into its own repository (according to existing patterns) if you decide to integrate this change into the stack. The `cfn-response.js` module is provided by Amazon and is being included in the lambda per their suggestion to help handle responses https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html

With this change, the priority for which AMI to use will be:
1. AMI from `ImageId` parameter if provided
2. Copied/encrypted version of default AMI if a `RootVolumeKmsKeyId` parameter is provided
3. Default AMI

As part of the copying process, the lambda will also create a grant on the KMS key for `AWSServiceRoleForEC2Spot` to allow spot instances to decrypt the encrypted volume.

The permission set on the `CopyImageLambdaExecutionRole` resource should be the least access privilege required to execute the AMI copy/encrypt process (based on much trial and error).

Very much open to feedback/changes on this, but wanted to open up the PR and see what people think.

### DIFF
--- a/temp-ami-copy/ami-copy/cfn-response.js
+++ b/temp-ami-copy/ami-copy/cfn-response.js
@@ -1,0 +1,54 @@
+/* Copyright 2015 Amazon Web Services, Inc. or its affiliates. All Rights Reserved.
+   This file is licensed to you under the AWS Customer Agreement (the "License").
+   You may not use this file except in compliance with the License.
+   A copy of the License is located at http://aws.amazon.com/agreement/ .
+   This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied.
+   See the License for the specific language governing permissions and limitations under the License. */
+
+exports.SUCCESS = "SUCCESS";
+exports.FAILED = "FAILED";
+
+exports.send = function (event, context, responseStatus, responseData, physicalResourceId, noEcho) {
+
+    var responseBody = JSON.stringify({
+        Status: responseStatus,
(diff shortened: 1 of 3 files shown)


lox:
I spent some time thinking on this yesterday. My concern is it's just a heap of code for what it does.

I wonder if there is any otherway we can do this 🤔

Perhaps I'll reach out to AWS.

underscorebrody:
I definitely agree, I'd love to do this in a simpler, cleaner way. Based on the way that KMS keys and volume encryption in EC2 works (or at least my understanding of it), the only other option is to do all this manually each time we want to bump our CF stack 😕. Would be very interested to hear what AWS has to say.

keithduncan:
Hi @underscorebrody I’m going to close this to not leave you hanging any longer.

Unfortunately this doesn’t appear to be a common enough use case that we’d look to adopt and support in the mainline Elastic CI Stack. We’d instead ask that you manage your encrypted EBS backed AMIs separately and then plug them using the `ImageId` stack parameter or `ImageIdParameter` AWS SSM Parameter Store parameter.

I’m sorry to turn down this contribution, thank you again for taking the time to write this implementation and open this pull request 🙇🏻‍♂️

PR #623, opened 2019-08-26
Allow customising parameters related to disk cleanup

### LINKED ISSUE #465: Provide a way of customizing `DOCKER_PRUNE_UNTIL`, `DISK_MIN_AVAILABLE`, and `DISK_MIN_INODES` for the cron job
Right now, there is no way of customizing those values for the hourly docker-gc cron job. The defaults are not good enough for us since our docker images are quite big and the disk gets fill up very quick. We also run a RAM drive for the runners, which requires different clean up params from the builders.

I imagine and approach could be to add those variables to the cloud formation template parameters, and generate the cron job on provision time, rather than image building time. Another option is to generate a file on provision time that the cron job would source when it runs. Thoughts?

lox:
Allows the min disk threshold before docker cleanup as a parameter. Can be a unit or a percentage.

Closes #465.


packer/linux/conf/bin/bk-check-disk-space.sh
```diff
@@ -1,25 +1,57 @@
...
+# min inodes must be a number, default to 250,000
+
+# Converts human-readable units like 1.43K and 120.3M to bytes
+dehumanize() {
+  awk '/[0-9][bB]?$/ {printf "%u\n", $1*1024}
+       /[tT][bB]?$/  {printf "%u\n", $1*(1024*1024*1024)}
+       /[gG][bB]?$/  {printf "%u\n", $1*(1024*1024)}
+       /[mM][bB]?$/  {printf "%u\n", $1*(1024)}
+       /[kK][bB]?$/  {printf "%u\n", $1*1}' <<< "$1"
+}
```
plasticine: lol, excellent 👌

packer/linux/conf/bin/bk-check-disk-space.sh
```diff
@@ -1,25 +1,57 @@
...
+
+min_available=${1:-5G}
```
plasticine: 5gig seems much more reasonable as a default here I think

packer/linux/conf/buildkite-agent/hooks/environment
```diff
@@ -17,13 +22,13 @@ if ! docker ps ; then
 fi
 
 echo "Checking disk space"
-if ! /usr/local/bin/bk-check-disk-space.sh ; then
+if ! /usr/local/bin/bk-check-disk-space.sh "${DISK_MIN_AVAILABLE:-5G}" ; then
```
plasticine: D’you need this default in two places?

packer/linux/conf/buildkite-agent/hooks/environment
```diff
@@ -17,13 +22,13 @@ if ! docker ps ; then
...
 
   echo "Cleaning up docker resources older than ${DOCKER_PRUNE_UNTIL:-4h}"
   docker image prune --all --force --filter "until=${DOCKER_PRUNE_UNTIL:-4h}"
 
   echo "Checking disk space again"
-  if ! /usr/local/bin/bk-check-disk-space.sh ; then
+  if ! /usr/local/bin/bk-check-disk-space.sh "${DISK_MIN_AVAILABLE:-5G}"; then
```
plasticine: Ditto

packer/linux/conf/buildkite-agent/hooks/environment
plasticine: Worth paraming inodes also here?

packer/linux/conf/bin/bk-check-disk-space.sh
sj26: Are these the right way around?

```
$ dehumanize 50b
51200
$ dehumanize 50kb
50
```

lox: Fixed @sj26

sj26: 🙌

templates/aws-stack.yml
```diff
@@ -375,6 +375,16 @@ Parameters:
       - "false"
     Default: "false"
 
+  MinimumDiskAvailableBeforeCleanup:
+    Type: String
+    Description: Either a percentage (%) or absolute unit (B, MB, GB) of disk below which disk cleanup is run
```
sj26: What an _absolute unit_.

PR #624, opened 2019-08-28
Fix lambda role

jradtilbrook:
Trying to use the CloudFormation template from the HEAD of master yields this error: `Error: ROLLBACK_COMPLETE: ["The following resource(s) failed to create: [AutoscalingLambdaExecutionRole]. . Rollback requested by user." "Value of property Policies must be a list of objects"]` (Note we are running the stack through terraform but I don't think that is necessarily related).

This PR fixes that issue.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -960,7 +960,7 @@ Resources:
                 - Effect: Allow
                   Action:
                     - ssm:GetParameter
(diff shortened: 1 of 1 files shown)


lox:
Thanks @jradtilbrook

jradtilbrook:
Well, seeing as how I introduced it in a broken state, that is quite alright! :tada: 
I've now tested it from the stack built on master and it's working for me now with KMS encryption and SSM parameter (from the lambda side). There's another issue now though that I'll raise in a separate issue

PR #625, opened 2019-08-29
Get SSM parameter as raw text

jradtilbrook:
This is a carry on from #624 to fix the newly introduced support for SSM parameter for the agent token.

Without the change, the get-parameter command returns the value as it appeared in the default json output which includes quotes.
Eg: `"token is here"`.

This was then double quoted by the script and when the EC2 instance boots up it sends the token to the API as `""token is here""` which is invalid.

With this change the quotes are removed from the get-parameter command so there is only one level of quotes


jradtilbrook:
@lox What do you think to this fix?

PR #626, opened 2019-09-02
Auto Scaling group with Mixed Instances

mohit-aws:
This enables combination of Instance types and pricing models for running CI jobs. It also adds the ability to control the mix of On-Demand and Spot instances.


lox:
This seems like a great idea! How would you recommend folks configure these? Is there some sensible defaults we can set?

mohit-aws:
@lox , I am making a few more changes to the template with better default configuration and also working on some internal approval to post my patch. I am going to delete this PR and then create a new one soon. Thanks for looking in to this.

PR #628, opened 2019-09-06
Ensure random windows password satisfies policy

jeremiahsnapp:
Every now and then we still see Windows instances fail to finish the `bk-install-elastic-stack.ps1` script because it doesn't generate a random password that satisfies the system's password policy even though the loop retries ten times. The other problem with that loop code is that it runs `exit 1` if the loop fails to generate a good password but apparently that `exit 1` doesn't trigger the `trap on_error` like you would expect in a bash script so the powershell script simply exits without marking the system unhealthy. That leaves the auto scaler in a confused state because an ec2 instance is running without a running buildkite agent.

After reviewing the Windows password policy we came up with the code in this PR that simply ensures that the generated password will always satisfy the password policy requirements. This allows us to get rid of the loop, the `try-catch` code and the `exit 1`.

### DIFF
--- a/packer/windows/conf/bin/bk-install-elastic-stack.ps1
+++ b/packer/windows/conf/bin/bk-install-elastic-stack.ps1
@@ -119,32 +119,29 @@ Set-PSDebug -Trace 0
 
 Write-Output "Creating buildkite-agent user account in Administrators group"
 
+$lowerChars = [char[]](97..122)  # a-z
+$upperChars = [char[]](65..90)   # A-Z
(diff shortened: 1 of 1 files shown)


jeremiahsnapp:
@lox we tested this code in our own AMIs and it's working well. We'd like to see it merged here so we can get the fix in your AMIs since we use them for some of our queues. Thanks!

lox:
Looks good! Thanks for the fix @jeremiahsnapp!

PR #633, opened 2019-09-12
Ensure powershell scripts fail when executable fails

jeremiahsnapp:
We have to explicitly test the exit status after running executables because powershell doesn't automatically fail when an executable's exit status is non-zero.

We've deployed these changes to our own Windows Buildkite packer template and it really increases the confidence that the powershell scripts haven't silently failed.

### DIFF
--- a/packer/windows/conf/bin/bk-fetch.ps1
+++ b/packer/windows/conf/bin/bk-fetch.ps1
@@ -5,6 +5,7 @@ $ErrorActionPreference = "Stop"
 
 If ($From -Like "s3://*") {
   aws s3 cp $From $To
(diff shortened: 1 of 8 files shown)


lox:
Ugh, so gross 😢

Are you sure there isn't an alternative?

tduffield:
Other than not use Powershell? Not that we've found. We got this guidance from a "Windows expert" (for whatever that's worth). But, for context, think of Powershell as a programming language, not a scripting one. You gotta tell Go when to exit, and with what error code.

tduffield:
At least here we don't have to handle `nil` :trollface:

lox:
Cool, let's merge this in. Is what it is!

PR #635, opened 2019-09-22
Auto Scaling group with Mixed Instances

mohit-aws:
This will enable the customers to launch and automatically scale a fleet of on-Demand Instances and Spot Instances within a single ASG to run the CI jobs. With this one or more instance types can also be specified which will override the default instance type specified in the launch template.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -56,6 +56,9 @@ Metadata:
         Parameters:
         - MinSize
         - MaxSize
+        - SpotAllocationStrategy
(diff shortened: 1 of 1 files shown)


lox:
@mohit-aws could I trouble you to rebase this?

My main questions here is how will users customize the instance types?

yob:
Thanks for this @mohit-aws. We've just merged mixed instance support in #710 - hopefully it meets your needs. If you get a chance to try it, we'd love your feedback.

PR #636, opened 2019-09-24
Replaced Linux fixed AMI source with source AMI filter

cawilson:
Replaced the hard-coded AMI source when building the Linux instance (using Packer) to use a filter query instead. This way the latest AMI is always sourced and the value doesn't need to be as actively maintained (matches the Packer builder for the Windows image, which uses a filter query).

Doing this also reduces the need to apply updates at build time.


lox:
Thanks @cawilson! Appreciate it! 🙇🏼‍♂️

PR #638, opened 2019-10-01
Power off immediately if cloud-init fails

### LINKED ISSUE #572: AWS EC2 instance never comes up - ASG reports 1 instance, but stick in pending
This sure looks like an AWS bug not a BK bug.

We have new autoscaler-based deployment in us-west-1. Our ASG goes all the way down to 0 instances, up to 15. Usually things work great; in the last few days we've had VMs that don't come up for 20+ minutes.

When there's only 1 step queued for an instance, this can be frustrating - that blocked step hangs seemingly "forever". BK does not try to scale up, EC2 does not actually have the instance.

The group is big enough that new VMs are spun up in both AZs multiple times per day, which rules out basic config errors like one subnet not being networked correctly. It really seems like an AWS issue.

dbaggerman:
This appears to fix #572 for us.

We see occasional failures from `cloud-init` failing to communicate with the EC2 metadata endpoint. These appear in systemd as a failure of the `cloud-final` service. By adding `OnFailure=poweroff.target`, the instance will power itself off when this occurs and allow the ASG to replace it instead of hanging around in a hung state forever.

### DIFF
--- a/packer/linux/conf/buildkite-agent/systemd/cloud-final.service.d/10-power-off-on-failure.conf
+++ b/packer/linux/conf/buildkite-agent/systemd/cloud-final.service.d/10-power-off-on-failure.conf
@@ -1,0 +1,3 @@
+[Unit]
+OnFailure=poweroff.target
+OnFailureJobMode=replace-irreversibly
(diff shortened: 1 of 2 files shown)


lox:
Darn, sorry I missed this @dbaggerman!

lox:
This seems like a great idea.

PR #639, opened 2019-10-03
Allow min disk before clean to be customized

### LINKED ISSUE #465: Provide a way of customizing `DOCKER_PRUNE_UNTIL`, `DISK_MIN_AVAILABLE`, and `DISK_MIN_INODES` for the cron job

arturopie:
Allows the min disk threshold before docker cleanup as an CloudFormation parameter. It can be a unit (e.g. 120.3M, 1.5GB)

Thanks @lox for starting the work on this. I'll try to finish it with this PR.

I'm also adding unit tests. If you are ok with me adding unit tests with shunit2, I can create another PR for setting up a Dockerfile for running all unit tests.

This is a draft since I'm going going through the comments from the original PR (https://github.com/buildkite/elastic-ci-stack-for-aws/pull/623)

Closes #465.


arturopie:
@lox I haven't manually tested these changes, and I don't have access to this repo's CI so I'm not sure about its coverage. How much manual testing does this PR need?

lox:
Really appreciate the help @arturopie! I'll review today.

lox:
@arturopie argh, of course `awk` maths is limited to 32-bit ints 😭

arturopie:
@lox @plasticine the CI is failing, but I have no visibility of it. Can you let me know what's the failure?

lox:
@arturopie it's a linting error, I'm working on fixes in #623, nearly there.

PR #641, opened 2019-10-27
Set up git credentials helper to use .git-credentials file as source

YuxuanLin:
Now if I upload git credentials to S3 bucket Git via https doesn't work directly. Setting git credential helper to pick up ~/.git-credentials is necessary and doable via env hooks but it's better to set up this step at an early stage and won't hurt anybody.

### DIFF
--- a/packer/linux/scripts/install-git-lfs.sh
+++ b/packer/linux/scripts/install-git-lfs.sh
@@ -9,3 +9,6 @@ tar -xvzf git-lfs.tgz -C git-lfs
 sudo chmod 755 git-lfs/git-lfs
 sudo ./git-lfs/install.sh
 rm -rf git-lfs.tgz git-lfs/
+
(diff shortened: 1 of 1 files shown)


lox:
Hey @YuxuanLin 👋🏻 You should be able to use `git-credentials` with https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks (which are included in the elastic stack)!

Check out https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/blob/master/hooks/environment#L83-L99

YuxuanLin:
Thanks @lox . Safe to close now.

Btw if the git-credentials file only contains one line uri without a new empty line below, it won't work silently because of this line https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/blob/687edd8bea54c62f1c35505d809209388827e176/git-credential-s3-secrets#L78

PR #642, opened 2019-10-30
Docker builder prune

sj26:
Also prune docker builder caches to avoid:

image-3


lox:
I’ll fix this build error tomorrow @sj26

sj26:
Oh, I did a bunch of cleanup in aws but forgot to retry the build sorry!

PR #643, opened 2019-11-01
Prune the docker build cache too

plasticine:
Does what it says. At the moment we just let the build cache grow forever.


plasticine:
Closing in favour of #642

PR #646, opened 2019-11-21
Increase the threshold for disk cleanup to 5GB free for 4.3

huonw:
e97921cc74e8a69fccbe98ebcc6ee0f50a9ccbcf on `master` increased the `DISK_MIN_AVAILABLE` threshold to 5GB a few months ago, but it doesn't seem to be in the 4.3.5 release that happened since then.

The 1GB threshold is causing us problems (https://github.com/buildkite/elastic-ci-stack-for-aws/issues/465#issuecomment-555846619) so this PR cherrypicks that existing commit onto the 4.3 series branch, as a temporary work-around while waiting for the stalled #623 and/or #639.

### DIFF
--- a/packer/conf/bin/bk-check-disk-space.sh
+++ b/packer/conf/bin/bk-check-disk-space.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


huonw:
Hi @lox, could I ask you to take a look at this PR? We're currently using a custom AMI built from this PR, instead of the 4.3.5 release.

Issue #246, opened 2017-02-10
Alternative to cloudformation


KevinGrandon:
Our security policy prevents us from using cloudformation to deploy an elastic stack due to strict iam resourcing control.

We're currently using Terraform and would love to have an alternative to cloudformation to setup the elastic stack.

lox:
Thanks for the issue @KevinGrandon, Terraform is absolutely something we are considering. Will discuss this some more and update shortly.

Would it be possible to provide more detail on CloudFormation? Does "strict iam resourcing control" rule out ANY CloudFormation, or are there things about how we are using that are problematic? Would it be possible for us to address these security concerns and let you move forward with CloudFormation? Many folks have been wrapping it in a Terraform resource: https://www.terraform.io/docs/providers/aws/r/cloudformation_stack.html

lox:
We could for instance lock down the permissions to only work in the provided vpc / subnets:

https://aws.amazon.com/blogs/security/how-to-help-lock-down-a-users-amazon-ec2-capabilities-to-a-single-vpc/

lox:
@toolmantim did you have any other insight into this one?

SomeoneWeird:
@KevinGrandon how does Cloudformation differ to Terraform?

lox:
Do you mean in so far as it's security implications?

toolmantim:
I don't have any more insight I'm afraid!

I know @pvdvreede is ingesting this via Terraform. Have you had any luck?

SomeoneWeird:
@lox Yeah. You should just be able to create an IAM user with the necessary permissions and then execute CF as that user. I'm not seeing what else terraform would do, seeing as it would obviously use a subset of IAM

pvdvreede:
We are using Terraform as a wrapper around the Cloudformation. In Terraform we create a IAM policy and then pass that policy into the Cloudformation. However, the Cloudformation still creates the instance profile, role and then attaches this extra policy. 

Would passing an IAM Role into the Cloudformation so that it only creates the instance profile, and connects it to the passed in Role help with your Security restrictions, @SomeoneWeird ?

SomeoneWeird:
@pvdvreede I think you mean @KevinGrandon there :)

lox:
I'd guess the issue here is the broad permissions that would be required for the stack in it's present form. It needs IAM creation for one, which is basically root access. We could look at locking this down, but I'd want to know if @KevinGrandon has other requirements beyond that.

[...]

PR #651, opened 2019-11-27
yml templates/aws-stack: Add Mixed Instance support to the ASG

aws5295:
Playing off of https://github.com/buildkite/elastic-ci-stack-for-aws/pull/635
This is one approach for adding mixed instance support while allowing the instance types to be configurable.

The obvious drawback, is that this only supports 1-4 instance types.
However, it lays down an example for how users can modify the template for more instance type support.

[AWS announcement on Mixed Instance types](https://aws.amazon.com/blogs/aws/new-ec2-auto-scaling-groups-with-multiple-instance-types-purchase-options/)

[Detailed description of the various parameters](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-purchase-options.html)

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -56,6 +56,13 @@ Metadata:
         Parameters:
         - MinSize
         - MaxSize
+        - SpotAllocationStrategy
+        - OnDemandBaseCapacity
+        - OnDemandPercentageAboveBaseCapacity
+        - InstanceType1
+        - InstanceType2
+        - InstanceType3
+        - InstanceType4
         - ScaleOutFactor
         - ScaleInIdlePeriod
         - ScaleOutForWaitingJobs
@@ -209,7 +216,7 @@ Parameters:
     MinLength: 1
 
   SpotPrice:
-    Description: Spot bid price to use for the instances. 0 means normal (non-spot) instances
+    Description: Maximum Spot bid price to use for the instances. 0 allows you to specify mixed instances (a combination of On-Demand and Spot Instances across multiple instances types)
     Type: String
     Default: 0
 
@@ -223,6 +230,44 @@ Parameters:
     Description: Minimum number of instances
     Type: Number
     Default: 0
+
+  InstanceType1:
+    Description: The primary instance type to use when requesting mixed instance types.
+    Type: String
+    Default: ""
+
+  InstanceType2:
+    Description: The secondary instance type to use when requesting mixed instances. Omit this parmameter to only request 1 instance type.
+    Type: String
+    Default: ""
+
+  InstanceType3:
+    Description: The tertiary instance type to use when requesting mixed instances. Omit this parmameter to only request 2 instance types.
+    Type: String
+    Default: ""
+
+  InstanceType4:
+    Description: The quaternary instance type to use when requesting mixed instances. Omit this parmameter to only request 3 instance types.
+    Type: String
+    Default: ""
+
+  SpotAllocationStrategy:
+    Description: Indicates how to allocate Spot capacity across Spot pools.
+    AllowedValues:
+    - capacity-optimized
+    - lowest-price
+    Type: String
+    Default: capacity-optimized
+
+  OnDemandBaseCapacity:
+    Description: Minimum number of instances in the ASG's initial capacity that must be fulfilled by On-Demand instances.
+    Type: Number
+    Default: 0
+
+  OnDemandPercentageAboveBaseCapacity:
+    Description: Percentage of On-Demand Instances for additional capacity beyond the optional On-Demand base amount.
+    Type: Number
+    Default: 40
 
   ScaleOutFactor:
     Description: A decimal factor to apply to scale out changes to speed up or slow down scale-out
@@ -437,6 +482,18 @@ Conditions:
 
     UseDefaultRootVolumeName:
       !Equals [ !Ref RootVolumeName, "" ]
+
+    UseInstanceType1:
+      !Not [ !Equals [ !Ref InstanceType1, ""] ]
+
+    UseInstanceType2:
+      !Not [ !Equals [ !Ref InstanceType2, ""] ]
+
+    UseInstanceType3:
+      !Not [ !Equals [ !Ref InstanceType3, ""] ]
+
+    UseInstanceType4:
+      !Not [ !Equals [ !Ref InstanceType4, ""] ]
 
     UseManagedPolicyARN:
       !Not [ !Equals [ !Join [ "", !Ref ManagedPolicyARN ], "" ] ]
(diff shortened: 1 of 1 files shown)


lox:
It pains me to add so many params, but I reckon this is probably a good compromise.

lox:
I'm actually tempted to remove the `SpotBid` param entirely.

kushmansingh:
Could the `InstanceType` parameter potentially be changed to the `CommaDelimitedList` type. I think that might be backwards compatible and the autoscalinggroup overrides could probably be created using `split` or some other cleverness

lox:
> Could the InstanceType parameter potentially be changed to the CommaDelimitedList type? I think that might be backwards compatible and the autoscalinggroup overrides could probably be created using split or some other cleverness

That would be ideal, but I don't believe it's possible. Would welcome someone else giving it a shot though!

kushmansingh:
I banged my head against this for a few hours and unfortunately there's no good way to do this with the Cloudformation functions 😞 . This is the closest I got but the output was a string and we need it to be an object mapping 
```yaml
            Overrides: !Split
              - ","
              - !Join
                - ",- InstanceType: "
                - !Split
                  - ","
                  - !Su
                    - ",${instanceTypes}"
                    - instanceTypes: !Join [",", !Ref InstanceType]
```

holmesjr:
This kind of thing could be a lot easier with CF generated by CDK. Would you be interested in a version of the BuildKite stack in CDK so you can evaluate it for future releases? I might be able to devote some time if so...

pda:
@holmesjr Yeah, CDK would open up a lot more flexibility.

I imagine it'll be tough to convince CDK to produce CloudFormation that could be safely applied as an update to an existing stack, without replacing lots of resources that don't really need replacing. Or, attempting to produce that CloudFormation from CDK would make the CDK more complex and less idiomatic.

It's worth thinking about, and perhaps tinkering, but no guarantees we'd switch to it, even if it was working.

We'd probably want to weight it up against e.g. Terraform in the context of thinking about how this choice fits with potentially creating equivalent stacks for other cloud providers / orchestration systems. 🤔

esalter:
This is also something we were looking into, so I'm glad to see someone else had the same problem. +1 from me.

I agree though that it would be better to not have so many params.  But it seems like that's our only option.

It's off topic, but I'd love to see a Terraform solution like @pda mentioned.  We actually manage everything in terraform, and this is the one part of our IaC that uses terraform to launch a CF stack, which just feels odd.

bshelton229:
We'd love to see this merged. We've been running into spot capacity issues taking our queues down until capacity returns. We're also a terraform shop and have stayed with this upstream cloudformation stack (applied via terraform in our shop) because of how great it is. We may need to bite off creating a terraform module to get mixed instance types in the ASG if it doesn't land in this stack in the near future. We would of course use the compiled resources (ami, lambda, etc) from this repo. Would there be interest in open sourcing a terraform solution? If terraform became a more supported option alongside cloudformation, would there be interest in the terraform living in this repo alongside everything else? Or, would a separate terraform module repo make more sense?

yob:
Thanks for the reminder @bshelton229.

We remain interested in improved support for utilising spot instances, and we're actively discussing options at the moment.

The parameter count is definitely a bummer. Partly for the overall impact on the ergonomics of using the stack, and partly because this takes us to 55 params - dangerously close to the cap of 60.

Personally, I haven't used a Mixed Instance ASG before. Could we do something like this?

1. Use a mixed instance ASG for all stacks (eg. drop the conditional on `MixedInstancesPolicy`)
2. Drop the `SpotPrice` param
3. Default `OnDemandPercentageAboveBaseCapacity` to 100%, so out of the box stacks will only use OnDemand instances
4. Document that teams can opt into using spot instances by reducing `OnDemandPercentageAboveBaseCapacity` to less 100% (possibly as low as 0%).
5. Drop `InstanceType1` in favour of the existing `InstanceType` param 

It's possible (likely?) I've suggested something that isn't practical. However, if it *did* work then the nett increase in params from this PR drops from +7 to +5.

The terraform questions are interesting and I don't know the answers offhand. There's some relevant discussion in #246 and it might be worth dropping a comment in there. I suspect our summary would be: we recognise terraform is super popular with our customers and would ideally like to support it, but as a small team we're nervous about maintaining feature parity in two parallel elastic stacks.

yob:
> Default OnDemandPercentageAboveBaseCapacity to 100%, so out of the box stacks will only use OnDemand instances

Purely on naming things: what about inverting this and calling the parameter `SpotInstanceMaxPercentage`. Default it to 0, and internally we set `OnDemandPercentageAboveBaseCapacity` to  `100 - SpotInstanceMaxPercentage`.

I reckon users installing the stack will be thinking of that question from the "how can I enable spot instances" angle, rather than "how can I reduce the number of OnDemand instances so spot instances fill the gap"

yob:
> Default it to 0, and internally we set OnDemandPercentageAboveBaseCapacity to 100 - SpotInstanceMaxPercentage.

Oh, silly me for assuming math functions are available in cloudformation :grimacing:

yob:
> Could the InstanceType parameter potentially be changed to the CommaDelimitedList type?

Here's the best I could come up with.

It treats the `InstanceType` param as a comma delimited value, but supports a fixed number of 1-4 types. It's a bit ugly internally (append extra `,` to the value to ensure there's always at least 4 items in the list, then re-split) but looks neat-ish from the users perspective.

```diff
diff --git a/templates/aws-stack.yml b/templates/aws-stack.yml
index 8296d75..21ebe7e 100644
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -39,9 +39,6 @@ Metadata:
         - ImageId
         - ImageIdParameter
         - InstanceType
-        - InstanceType2
-        - InstanceType3
-        - InstanceType4
         - AgentsPerInstance
         - KeyName
         - SecretsBucket
@@ -207,7 +204,7 @@ Parameters:
     Default: ""
 
   InstanceType:
-    Description: Instance type
+    Description: Instance type. Comma separated list with 1-4 instance types. The order is a prioritised preference for launching OnDemand instances, and an non-prioritised list of types to consider for Spot Instances (where used).
     Type: String
     Default: t3.large
     MinLength: 1
@@ -223,21 +220,6 @@ Parameters:
     Type: Number
     Default: 0
 
-  InstanceType2:
-    Description: The secondary instance type to use when requesting mixed instances. Omit this parmameter to only request 1 instance type.
-    Type: String
-    Default: ""
-
-  InstanceType3:
-    Description: The tertiary instance type to use when requesting mixed instances. Omit this parmameter to only request 2 instance types.
-    Type: String
-    Default: ""
-
-  InstanceType4:
-    Description: The quaternary instance type to use when requesting mixed instances. Omit this parmameter to only request 3 instance types.
-    Type: String
-    Default: ""
-
   OnDemandPercentage:
     Description: Percentage of total instances that should launch as OnDemand. Reduce this to use some Spot Instances when they're available.
     Type: Number
@@ -465,13 +447,13 @@ Conditions:
       !Equals [ !Ref RootVolumeName, "" ]
 
     UseInstanceType2:
-      !Not [ !Equals [ !Ref InstanceType2, ""] ]
+      !Not [ !Equals [ !Select [ "1", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ], ""] ]
 
     UseInstanceType3:
-      !Not [ !Equals [ !Ref InstanceType3, ""] ]
+      !Not [ !Equals [ !Select [ "2", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ], ""] ]
 
     UseInstanceType4:
-      !Not [ !Equals [ !Ref InstanceType4, ""] ]
+      !Not [ !Equals [ !Select [ "3", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ], ""] ]
 
     UseManagedPolicyARN:
       !Not [ !Equals [ !Join [ "", !Ref ManagedPolicyARN ], "" ] ]
@@ -826,7 +808,7 @@ Resources:
           KeyName: !If [ "HasKeyName", !Ref KeyName, !Ref 'AWS::NoValue' ]
           IamInstanceProfile:
             Arn: !GetAtt "IAMInstanceProfile.Arn"
-          InstanceType: !Ref InstanceType
+          InstanceType: !Select [ "0", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ]
           ImageId: !If [ HasImageId, !Ref ImageId, !If [ HasImageIdParameter, !GetAtt ImageIdParameterStack.Outputs.ImageId, !FindInMap [ AWSRegion2AMI, !Ref 'AWS::Region', !Ref InstanceOperatingSystem ] ] ]
           BlockDeviceMappings:
             - DeviceName: !If [ UseDefaultRootVolumeName, !If [ UseWindowsAgents, /dev/sda1, /dev/xvda ], !Ref RootVolumeName ]
@@ -940,18 +922,18 @@ Resources:
             LaunchTemplateId: !Ref AgentLaunchTemplate
             Version: !GetAtt "AgentLaunchTemplate.LatestVersionNumber"
           Overrides:
-          - InstanceType: !Ref InstanceType
+          - InstanceType: !Select [ "0", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ]
           - InstanceType: !If
             - UseInstanceType2
-            - !Ref InstanceType2
+            - !Select [ "1", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ]
             - !Ref "AWS::NoValue"
           - InstanceType: !If
             - UseInstanceType3
-            - !Ref InstanceType3
+            - !Select [ "2", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ]
             - !Ref "AWS::NoValue"
           - InstanceType: !If
             - UseInstanceType4
-            - !Ref InstanceType4
+            - !Select [ "3", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ]
             - !Ref "AWS::NoValue"
       MinSize: !Ref MinSize
       MaxSize: !Ref MaxSize
```

yob:
Thanks again for this, and apologies it took us so long to mull over  @aws5295 :heart: 

We've just merged #710, based on this PR. Hopefully it gets the balance right between support for mixed instance ASGs and conserving our remaining parameter count! Feedback is welcome.

Issue #652, opened 2019-11-28
bk-install-elastic-stack.sh gets errors during initialization


tiendan:
Hello,

We have an AWS Buildkite stack on `us-west-2` region (which uses the `ami-0662826fbca636496` AMI by default), and when we try to create a custom AMI the instances cannot start correctly. In EC2 console we are seeing:

`Client.InstanceInitiatedShutdown: Instance initiated shutdown`

As the state transition message, and when we checked the scripts our best guess was that it was getting an error during user data execution and stopping itself ([this script](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/linux/conf/bin/bk-install-elastic-stack.sh#L15)).

When we manually ran the user data script on a machine with custom AMI and another with default AMI, we got errors due to undefined variables such as `BUILDKITE_AGENT_TOKEN_PATH` and `BUILDKITE_SCALE_IN_IDLE_PERIOD`. Do you think these are basic bash syntax errors? Or is there something wrong with our stack definition? We recreated the stack using `https://s3.amazonaws.com/buildkite-aws-stack/latest/aws-stack.yml` just yesterday, so it's pretty up-to-date.

Thanks!

lox:
I'll have a look, but just so you are aware `https://s3.amazonaws.com/buildkite-aws-stack/latest/aws-stack.yml` refers to the latest released version, not the `master`, which is https://s3.amazonaws.com/buildkite-aws-stack/master/aws-stack.yml.

Perhaps you might be mixing and matching those two?

tiendan:
Ok, sorry I didn't realize the branch structure and as the commits for these two parameters were quite old (4 months and 7 months old), compared to the latest 4.3.5 AMI build (less than a month) I assumed they were part of the stable build too. Closing the issue & the PR, thanks!

PR #653, opened 2019-11-28
Fix for possibly unset variables

tiendan:
Should fix https://github.com/buildkite/elastic-ci-stack-for-aws/issues/652


lox:
These shouldn't be unset!

tiendan:
Ok, I see! Thanks to your comment on the issue, I could see the branch structure and this should fix the issue for us.

PR #654, opened 2019-12-05
bump docker and docker compose versions

dreyks:
Docker 19.03.5 + docker compose 1.25.0 (allowing to use BUILDKIT builder backend)


lox:
Awesome!

sj26:
Docker Compose 1.25.0 seems to have a critical regression which may break the docker-compose plugin's ability to use pre-built images:

https://github.com/docker/compose/pull/7052

:-(

1.25.1 is coming soon, though.

PR #655, opened 2019-12-06
Revert "bump docker and docker compose versions"

pda:
Partially reverts buildkite/elastic-ci-stack-for-aws#654

docker-compose 1.25.0 has a regression preventing https://github.com/buildkite-plugins/docker-compose-buildkite-plugin from using an image pre-built in a previous build step. This would cause much longer build times, and in some cases build failures if the later step is unable to build the image itself (e.g. if the Dockerfile is BuildKit-specific but the step building it doesn't support that).

See: https://github.com/docker/compose/pull/7039 and other issues linked from there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ If you really want to store your secrets unencrypted, you can disable it entirel
 
 * [Amazon Linux 2 LTS](https://aws.amazon.com/amazon-linux-2/)
 * [Buildkite Agent v3.16.0](https://buildkite.com/docs/agent)
-* [Docker 19.03.2](https://www.docker.com)
+* [Docker 19.03.5](https://www.docker.com)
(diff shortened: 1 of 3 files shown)


sj26:
Do we need to revert docker as well, could we just revert compose? Or are they tightly coupled?

pda:
Hmm good point. I don't  think they're coupled.

The other option is we could bump elastic-stack master to docker-compose 1.25.1-rc1,  but perhaps it's better to stay on stable 1.24.1

sj26:
Yeah I reckon leave docker bumped, but revert docker-compose until 1.25.1 proper is released.

pda:
This is now a partial-revert PR, just getting docker-compose back down to 1.24.1

sj26:
🙌

PR #659, opened 2019-12-22
POC: Add nvidia-docker

donbobka:
Just a Proof of Concept. 

Things to improve:
- Probably it's better to create separate AMI with nvidia-docker. 
- Make configurable default runtime (current PR makes `nvidia` as default runtime)


sj26:
Very interesting! But I'm sorry we can't accept this into the default stack. Please feel free to run this and create your own AMIs for use in your stacks. :-)

PR #664, opened 2020-02-05
README.md: Fix invalid IAM policy

KevinGrandon:
I received an invalid JSON error when using the example. Slightly better developer experience for folks using this snippet if we update it.


pda:
Thanks @KevinGrandon :+1:

PR #667, opened 2020-02-18
Support multiple build instance security groups

### LINKED ISSUE #562: Support for multiple Security Groups
I understand why `SecurityGroupId` is a `String` rather than a `CommaDelimitedList`, but I'd like to launch my stack with several SGs attached.

If this isn't on the radar, I'll probably get around to a PR at some point, but just wanted to check here first.

jdub:
Fixes #562. Doesn't break existing users of `SecurityGroupId`, but it could be deprecated and later removed.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -265,7 +265,7 @@ Parameters:
 
   SecurityGroupId:
     Type: String
-    Description: Optional - Security group id to assign to instances
+    Description: Optional - Comma separated list of security group ids to assign to instances
     Default: ""
 
   ImageId:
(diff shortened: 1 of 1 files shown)


templates/aws-stack.yml
```diff
@@ -268,6 +269,11 @@ Parameters:
     Description: Optional - Security group id to assign to instances
```
rich-zhang: should we put "DEPRECATED" here? Or is it possible that we might not deprecate it?

sj26:
Thanks for the contribution @jdub! This looks to me like we could use the existing parameter, and adjust the desciption to something like "multiple groups can be separated by commas" and it would have clean backwards compatibility?

jdub:
@sj26 Happy to make that change.

jdub:
@sj26 Ready to go.

jdub:
@sj26 pls re-review 😄

sj26:
Sorry for the delay @jdub, thinks have got a bit hairy. I'll poke someone on the team who can do another review and assist with the release process.

jdub:
@sj26 @pda Thanks!

PR #670, opened 2020-03-03
windows: pin awscli to v1; v2 breaks 'aws ecr get-login'

pda:
Windows AMI pinned to awscli v1

See https://github.com/buildkite-plugins/ecr-buildkite-plugin/issues/37

This patch pins `choco install awscli` to the currently-latest `1.18.11`; see https://chocolatey.org/packages/awscli/1.18.11#versionhistory (Chocolatey doesn't make it possible to pin to a major version.)

This should be reverted after the ECR plugin supports awscli v2.

### DIFF
--- a/.buildkite/steps/launch.sh
+++ b/.buildkite/steps/launch.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-# shellcheck disable=SC1117
 set -eu
 
 os="${1:-linux}"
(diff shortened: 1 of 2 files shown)


yob:
should we backport this to `stable-4-3` as well?

yob:
oh wait, there's no windows support in `stable-4-3`. Forget I said anything :D

pda:
Should be able to unpin now that https://github.com/buildkite-plugins/ecr-buildkite-plugin/pull/40 is available?

yob:
I think we're intentionally keeping awscli v1 for the upcoming v5.0.0 release, to avoid adding another breaking change into a release that's been baking for so long. We hope to bump to v2 after the v5 release though.

PR #683, opened 2020-04-24
Set content-type on YML Files

kyledecot:
I recently ran into an issue when attempting to download one of the CloudFormation YAML templates provided in the release via Terraform which is caused by not setting the `content-type` to `text/yaml` on the files stored in S3.

Currently the files return `application/octet-stream` which causes the [http datasource](https://www.terraform.io/docs/providers/http/data_source.html) in Terraform to report failure as it only allows `text/*` and `application/json`.

This PR sets the content type to `text/yaml` during the upload of the file which happens in a Buildkite step.


yob:
thanks!

Issue #597, opened 2019-06-10
Replace Lambda env with SSM (or Secrets Manager) for agent token


pauldraper:
Background: While probably not unanimous, there is some consensus that secrets should be kept in SSM parameter store (or Secrets Manager) rather than Lambda env vars.

Lambda functions then use the AWS SDK to request the secrets at runtime.

* https://hackernoon.com/you-should-use-ssm-parameter-store-over-lambda-env-variables-5197fc6ea45b
* https://medium.com/@furqanshaikh/securing-secrets-in-aws-lambda-93c12861371f
* https://dev.to/dvddpl/where-do-you-keep-credentials-for-your-lambda-functions-5dno
* https://www.sethvargo.com/secrets-in-serverless/
* https://www.stackery.io/blog/serverless-secrets/

tl;dr it provides finer-grained access to secrets and lessens likelihood of exposure.

---

Proposal: Have an template paremeter (with default, .e.g. /buildkite/agent-token) for an SSM path. The user is required to place the agent token at this path.

Side benefit: it will fix #422

lox:
This is something we've been discussing a lot recently @pauldraper. I completely agree, although the path to getting there isn't as clear.

The issue is primarily the user experience of creating a new Elastic Stack for the first time. It's incredibly easy to generate a token and create a new stack currently and we're reluctant to impact that. It also makes upgrading between the two approaches hard. 

If the specific issue is that we should be using SSM to pass the secret to lambda, perhaps we could treat that in isolation to entering the token into the template?

ianwremmel:
[`chamber`](https://github.com/segmentio/chamber) (with, in some cases, the assistance of [`aws-vault`](https://github.com/99designs/aws-vault)) makes it really easy to put secrets in a predictable location in SSM.

jradtilbrook:
FWIW I've found that the lambda already supports retrieving a token from SSM by setting the path/key in the environment variable `BUILDKITE_AGENT_TOKEN_SSM_KEY`. So I think the minimal change is to modify this cloudformation template to set that on the lambda.  
I've made a start in #601 and would appreciate review and comments :pray:

PR #685, opened 2020-05-11
Import Lambda based autoscaler from the Severless Application Repository

keithduncan:
I’ve replaced the inline Lambda autoscaler resources with a reference to the test serverless application repository entry I created to test this.

The changes I made there to require an SSM key have propagated up to here too where everything downstream now depends on and uses the SSM Parameter. This template either creates a plaintext one from the (now deprecated) given BuildkiteAgentToken, or passes through the given BuildkiteAgentTokenParameter.

Depends on https://github.com/buildkite/buildkite-agent-scaler/pull/28

cc @lox @pda @chloeruka 

Fixes https://github.com/buildkite/elastic-ci-stack-for-aws/issues/597


Makefile
```diff
@@ -115,15 +115,15 @@ create-stack: build/aws-stack.yml env-STACK_NAME
 		--stack-name $(STACK_NAME) \
 		--disable-rollback \
 		--template-body "file://$(PWD)/build/aws-stack.yml" \
-		--capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM \
+		--capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM CAPABILITY_AUTO_EXPAND \
```
keithduncan, line 118: This is required due to using the `Transform` declaration, itself required due to adding an `AWS::Serverless::Application` resource.

templates/aws-stack.yml
```diff
@@ -593,6 +578,14 @@ Resources:
       SubnetId: !Ref Subnet1
       RouteTableId: !Ref Routes
 
+  BuildkiteAgentTokenParameter:
+    Type: AWS::SSM::Parameter
+    Condition: CreateAgentTokenParameter
+    Properties:
+      Name: !Sub "/${AWS::StackName}/buildkite/agent-token"
```
keithduncan, line 585: I might rename this `/${AS::StackName}/buildkite-agent/token`, thoughts?

templates/aws-stack.yml
```diff
@@ -949,10 +941,8 @@ Resources:
         Statement:
           - Effect: Allow
             Principal:
-              Service:
-                - lambda.amazonaws.com
-            Action:
-              - sts:AssumeRole
+              Service: lambda.amazonaws.com
```
keithduncan, line 944: These are minor style changes that helped me read it, I’m happy to remove them to minimise the diff

templates/aws-stack.yml
```diff
@@ -970,85 +960,43 @@ Resources:
             Version: '2012-10-17'
             Statement:
               - Effect: Allow
-                Action:
-                  - cloudwatch:PutMetricData
+                Action: cloudwatch:PutMetricData
                 Resource: '*'
+        - PolicyName: ReadAgentToken
```
keithduncan, line 965: I think I want to get rid of this entire resource and either accept a KMS key ID as a parameter of the lambda application for it to generate the correct IAM policy, or adding an `AWS::IAM::Policy` to its generated role here.

templates/aws-stack.yml
```diff
@@ -116,6 +112,12 @@ Parameters:
     Type: String
     Default: ""
 
+  BuildkiteAgentToken:
```
keithduncan, line 115: I just tested switching to an ssm parameter, the old BuildkiteAgentToken does stick around in the CloudFormation template due to NoEcho. To remove it I had to redeploy the existing template on the console and uncheck "use existing value" for this parameter.

templates/aws-stack.yml
xiaket: Not really sure what is the benefit to make this stack dependent, as most of the use cases, we will have one token for several bk stacks. BTW, if we have two organizations within the same AWS account, the user can easily customize this value by not using the default one.

We have used `/vendors/buildkite/agent-token` and it works for us really well.

keithduncan: This resource is only created if you don’t specify your own SSM parameter path. If this stack were to create a global resource _not stack scoped_ it would conflict with subsequent deployments of the same CloudFormation template.

keithduncan:
@yob this is the change to switch from in-line CloudFormation resources to imported resources via the Serverless Application Repository.

This simplifies deploying new versions of the lambda scaler as the IAM policies and code live in the same repository and are always deployed simultaneously, and removes a maintenance point as we no longer have to copy the lambda code to new regions for customers to deploy an elastic stack there.

When I took this pass over the scaler SAR repository I renamed it from `buildkite-elastic-ci-scaler` to `buildkite-agent-scaler`, but with fresh eyes this morning I’m erring the other way and am considering renaming it back so the name more tightly scopes it to the elastic ci stack.

templates/aws-stack.yml
```diff
@@ -103,12 +105,6 @@ Parameters:
       - edge
     Default: "stable"
 
-  BuildkiteAgentToken:
-    Description: Buildkite agent registration token
-    Type: String
-    NoEcho: true
-    Default: ""
```
yob, line 110: If we're removing a parameter, I think we'll need to release it as v6.x. If we're doing **that** there's a list of possible breaking changes we've been holding off over in https://github.com/buildkite/elastic-ci-stack-for-aws/issues/792

yob (commented):
This is interesting! It seems like a good improvement.

One thing I'm not across is how moving the IAM role for the scaler into SAR works in practice. We can just update the role as required and release a new version, reference the new version here, and stack upgrades will pull in the update IAM role? Magic :sparkles:

templates/aws-stack.yml
keithduncan: This must be a merge artifact because the parameter is still present just lower, I can fix this up in a changeset

templates/aws-stack.yml
```diff
@@ -2,15 +2,17 @@
...
 Metadata:
   AWS::CloudFormation::Interface:
     ParameterGroups:
       - Label:
           default: Buildkite Configuration
         Parameters:
-        - BuildkiteAgentToken
         - BuildkiteAgentTokenParameterStorePath
         - BuildkiteAgentTokenParameterStoreKMSKey
+        - BuildkiteAgentToken
```
keithduncan, line 15: This is deliberately moved to give lower precedence to the inline token option vs using an SSM Parameter

keithduncan:
> We can just update the role as required and release a new version, reference the new version here, and stack upgrades will pull in the update IAM role? Magic ✨

Exactly! Bumping the referenced SAR version will deploy the updated lambda scaler and its role as part of its template.

I’ve double checked going through the changes here and the old `BuildkiteAgentToken` parameter is still supported throughout, though marked as deprecated and something we could remove in a major version bump. The idea is to align everything towards using an SSM Parameter and have this template create a plaintext one from the given plaintext token if needed.

This is a change though, and places SSM Parameter Store on the critical path for instance booting for everyone. Once an instance is booted and has executed the UserData script, the token is cached locally in the agent config file, so it’s not a real time dependency for accepting and running jobs.

Issue #689, opened 2020-05-19
Update version of docker-compose on agent


zermatt:
Hi, would it be possible to update the version of docker-compose baked into the agent, so the latest version of the compose file format (`3.8` at time of writing) is supported. thanks.

chloeruka:
I'll consider this along with #688 this week alongside any relevant core Docker updates. 

We're currently using Docker Compose `1.25.1`:
https://github.com/buildkite/elastic-ci-stack-for-aws/blob/04c0ced472e0845173ab955735461b90b125ae10/packer/linux/scripts/install-docker.sh#L4-L6

This would be a huge jump, but we also haven't updated this dependency since 2018 so it might be worth reviewing.

chloeruka:
Ah; whoops, now I realise you meant the file format version which is separate from the compose version – this looks like a pretty straightforward update from current to `>1.25.5` to support file format 3.8. I'll probably bump us to latest (`1.26.2`) and see how we get on.

yob:
This was completed in #694

PR #694, opened 2020-07-22
Update docker & docker-compose

chloeruka:
For #689. Updating these docker dependencies is part of our regular release cycle.

Docker release notes  (19.03.2 > 19.03.12) 
https://docs.docker.com/engine/release-notes/

Compose release notes (moving from 1.25.1 > 1.26.2)
https://docs.docker.com/compose/release-notes/

### DIFF
--- a/packer/scripts/install-docker.sh
+++ b/packer/scripts/install-docker.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 set -eu -o pipefail
 
-DOCKER_VERSION=19.03.2
+DOCKER_VERSION=19.03.12
 DOCKER_RELEASE="stable"
-DOCKER_COMPOSE_VERSION=1.25.1
+DOCKER_COMPOSE_VERSION=1.26.2
 
 # This performs a manual install of Docker.
 


sj26:
We've been bitten by unexpected changes during Docker upgrades. Is there anything significant worth mentioning? Otherwise it's probably good to bump and then test it on edge.

sj26:
(Scanning the release notes nothing stands out to me, although there are lots of security patches so 🏎💨)

chloeruka:
> I'd like to think the keeping within the 19.03.xx series means we only get bug fixes. Right?

I'd like to hope so! Not every org follows semver as accurately as I'd like, but looking at the notes it looks to just be bugfixes for docker. docker-compose however does include a minor version bump, but looking at the changes it should be pretty low impact.

DanielHeath:
Any way this could be a stack parameter for those who want to opt-in to newer versions?

I'd like to use the new 'target' feature in 1.27.x

yob:
If it was a stack parameter, would that mean each instance downloads the requested version (if required) on boot?

I suspect we might be reluctant to slow down instance boot times, but it's certainly possible to do the download yourself in a [bootstrap script](https://github.com/buildkite/elastic-ci-stack-for-aws#customizing-instances-with-a-bootstrap-script).

In this particular case, we're aiming for a new release of the stack in the next week or two. Are you up for opening a PR to bump docker compose to 1.27.x? If so, we can sneak it in before the release.

DanielHeath:
Ahh that makes sense. 

One option would be to make it a param in the docker-compose plugin (which downloads/reuses a different version if specified).

> In this particular case, we're aiming for a new release of the stack in the next week or two. Are you up for opening a PR to bump docker compose to 1.27.x? If so, we can sneak it in before the release.

Turns out I was experiencing some PEBKAC and no longer need the upgrade.

PR #698, opened 2020-08-18
Only set git-mirrors-path when EnableAgentGitMirrorsExperiment

pda:
We currently set `git-mirrors-path` even when [`EnableAgentGitMirrorsExperiment`](https://github.com/buildkite/elastic-ci-stack-for-aws/pull/555) is `false`. That would lead to unexpected activation of the feature if a newer version of https://github.com/buildkite/agent promotes `git-mirrors` from experiment status.

This patch configures `git-mirrors-path` as an empty string unless `EnableAgentGitMirrorsExperiment` is `true`.

### DIFF
--- a/packer/windows/conf/bin/bk-install-elastic-stack.ps1
+++ b/packer/windows/conf/bin/bk-install-elastic-stack.ps1
@@ -66,7 +66,7 @@ If (Test-Path Env:BUILDKITE_AGENT_TAGS) {
   $agent_metadata += $Env:BUILDKITE_AGENT_TAGS.split(",")
 }
 
-# Enable git mirrors
+# Enable git-mirrors
 If ($Env:BUILDKITE_AGENT_ENABLE_GIT_MIRRORS_EXPERIMENT -eq "true") {
   If ([string]::IsNullOrEmpty($Env:BUILDKITE_AGENT_EXPERIMENTS)) {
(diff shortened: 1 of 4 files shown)


packer/windows/conf/bin/bk-install-elastic-stack.ps1
```diff
@@ -86,7 +87,7 @@ timestamp-lines=${Env:BUILDKITE_AGENT_TIMESTAMP_LINES}
 hooks-path="C:\buildkite-agent\hooks"
 build-path="C:\buildkite-agent\builds"
 plugins-path="C:\buildkite-agent\plugins"
-git-mirrors-path="C:\buildkite-agent\git-mirrors"
+git-mirrors-path="${Env:BUILDKITE_AGENT_GIT_MIRRORS_PATH}"
```
pda, line 90: I don't know much Windows / PowerShell and was worried this might fail if the env wasn't set at all, which is the case when `BUILDKITE_AGENT_ENABLE_GIT_MIRRORS_EXPERIMENT` is false.

I've confirmed¹ that it results in `git-mirrors-path=""` in the Windows `buildkite-agent.cfg` file when that env var isn't present 👌🏼

¹ a decade of using AWS/EC2, and this is the first time I've ever connected to a windows instance 😬

pda:
I've also cherry-picked this onto the `v4-development` branch as dccac5091a881ebe91b344ef6727d327e0f8f817.

PR #699, opened 2020-08-20
InstanceType default t3.small, was t2.nano

pda:
The newer `t3` instance family is objectively better than `t2`.
The `large` size (8 GiB RAM) is more realistic than `nano` (0.5 GiB RAM) for actual workloads, and still very affordable with autoscaling.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -206,7 +206,7 @@ Parameters:
   InstanceType:
     Description: Instance type
     Type: String
-    Default: t3.small
+    Default: t3.large
     MinLength: 1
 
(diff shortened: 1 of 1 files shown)


chloeruka (approved):
Sensible, very low impact change. 🚀

chloeruka:
Moving to a large makes sense for most of the users of our stack, but I'd want to emphasise somewhere in the documentation that there's no 'minimum' agent size and you can choose a smaller instance if it makes more sense for your use-case. I wouldn't want a `t3.large` to upload something to S3 or deploy a cloudformation changeset for example.

PR #707, opened 2020-09-16
Ensure terminate instance exits gracefully

patrobinson:
We run a slightly modified version of this stack, in which it doesn't have permission to call `terminate-instance-in-auto-scaling-group`.

I will change that, but for robustness I don't think a failure here should cause the box to stay up (and the buildkite-agent to get restarted by systemd).
Just falling through to `sudo poweroff` will be sufficient.

This causes issues with scaling down, as the non-zero exit code from `aws autoscaling terminate-instance-in-auto-scaling-group` results in systemd restarting buildkite-agent, which means it takes jobs even though we're trying to drain it ready for scale down.

### DIFF
--- a/packer/linux/conf/buildkite-agent/scripts/terminate-instance
+++ b/packer/linux/conf/buildkite-agent/scripts/terminate-instance
@@ -5,4 +5,4 @@ set -euo pipefail
 instance_id=$(curl -fsSL http://169.254.169.254/latest/meta-data/instance-id)
 region=$(curl -fsSL http://169.254.169.254/latest/meta-data/placement/availability-zone | head -c -1)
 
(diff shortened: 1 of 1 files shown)


yob:
Thanks for the detailed context, this seems like a sensible change.

Iniitally I thought it'd be nice to print a warning to stderr about the permissions issue, but doing that with `set -e` is awkward and I assume `aws autoscaling terminate-instance-in-auto-scaling-group` might be printing something to stderr anyway?

What's the expected behaviour if there's a permissions issue after this merges? The instance will stop so it no-longer accept jobs, and the ASG desired size won't decrement? Does a new instance start in its place?

yob:
(none of my questions are blockers!)

patrobinson:
@yob yes, a new instance will start in it's place, but given desired count is controlled by the autoscaler I think that's preferred.
Having a seperate system adjusting desired count seperate to the autoscaler seems a recipe for unexpected behaviour.

PR #710, opened 2020-09-18
Support multiple instance types by switching to a Mixed Instance ASG

yob:
I started reviewing #651 as an excuse to improve my CloudFormation knowledge, and ended up extending it into an alternative approach that might be worth considering.

This makes the following changes to master:

1. Replaces the `SpotPrice` parameter with `OnDemandPercentage` that defaults to 100%. If this is reduced, the stack will potentially use Spot instances if they're available and cheaper than the standard OnDemand price
1. Allows the `InstanceType` parameter to be a CSV with 1- 4 types. When launching  OnDemand instances the types will be used in priority order, and when launching spot instances (if the config allows them) all types will be considered (AWS will pick the instances with the best capacity and least chanec of being terminated).

The internal implementation of the `InstanceType` CSV is pretty ugly, but at least the users of the stack don't see it. We append some extra commas before splitting it to ensure the CF `!Select` function is **always** passed a list with at least 4 items.

A nice side effect of this change is that a single stack can be configured to use Spot instances when they're available, but fall back to OnDemand instances when spot capacity is unavailable. I'm a little hazy on how Reserved Instances work, but I think the priority order can also be used to prefer instance types with reserved pricing, but then fallback to other types.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -221,7 +221,7 @@ Parameters:
     Default: 0
 
   OnDemandPercentage:
-    Description: Percentage of total instances that should launch as OnDemand. Reduce this to use some Spot Instances when they're available.
+    Description: Percentage of total instances that should launch as OnDemand. Default is 100% OnDemand - reduce this to use some Spot Instances when they're available and cheaper than the OnDemand price.
     Type: Number
     Default: 100
     MinValue: 0
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -39,9 +39,11 @@ Metadata:
         - ImageId
         - ImageIdParameter
         - InstanceType
+        - InstanceType2
+        - InstanceType3
+        - InstanceType4
         - AgentsPerInstance
         - KeyName
-        - SpotPrice
         - SecretsBucket
         - ArtifactsBucket
         - AuthorizedUsersUrl
@@ -57,13 +59,7 @@ Metadata:
         Parameters:
         - MinSize
         - MaxSize
-        - SpotAllocationStrategy
-        - OnDemandBaseCapacity
-        - OnDemandPercentageAboveBaseCapacity
-        - InstanceType1
-        - InstanceType2
-        - InstanceType3
-        - InstanceType4
+        - OnDemandPercentage
         - ScaleOutFactor
         - ScaleInIdlePeriod
         - ScaleOutForWaitingJobs
@@ -216,11 +212,6 @@ Parameters:
     Default: t3.large
     MinLength: 1
 
-  SpotPrice:
-    Description: Maximum Spot bid price to use for the instances. 0 allows you to specify mixed instances (a combination of On-Demand and Spot Instances across multiple instances types)
-    Type: String
-    Default: 0
-
   MaxSize:
     Description: Maximum number of instances
     Type: Number
@@ -231,11 +222,6 @@ Parameters:
     Description: Minimum number of instances
     Type: Number
     Default: 0
-
-  InstanceType1:
-    Description: The primary instance type to use when requesting mixed instance types.
-    Type: String
-    Default: ""
 
   InstanceType2:
     Description: The secondary instance type to use when requesting mixed instances. Omit this parmameter to only request 1 instance type.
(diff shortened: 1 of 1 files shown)


yob:
Here's a screenshot of a single stack that's launched with mix of m4.large/m5.large and spot/ondemand:

mixed-instance-asg

JuanitoFatas (approved):
Looks good to me 👍🏼 I have tried this Pull Request on real production and it works as described 👏🏼👏🏼👏🏼

yob:
I plan to rework this to keep the `SpotPrice` param, making the PR additive only. That'll make it a non-breaking change, and a candidate for a post 5.0 release (maybe 5.1?).

chloeruka:
I'm going to be testing this today 😄

chloeruka:
Looking good, I have 5 instances with 20% on demand configured with the following `InstanceType`s (in order):
1. `t3.large`
1. `t3.medium` 
1. `m5.large`
1. `t3.small`


image

Pleasingly we see 1 `t3.large` (on-demand), and 4 `t3.medium` spot instances.

chloeruka (approved):

yob:
> Pleasingly we see 1 `t3.large` (on-demand), and 4 `t3.medium` spot instances.

It'd be really neat if we could add the Spot yes/no state into agent metadata. It seems like we can't determine the spot-iness of an instance from the EC2 instance metadata though. Bummer.

vgrigoruk:
Hi @yob,

It is actually very easy to add agent metadata/tag, which will tell if the instance is spot/on-demand. Here is how we do it in `agent-bootstrap.sh`:
```
echo "tags-from-ec2-meta-data-paths=aws:instance-life-cycle=instance-life-cycle" >> /etc/buildkite-agent/buildkite-agent.cfg
```
The value of the tag will be either `on-demand` or `spot`. 

But, it turns that using both `on-demand` and `spot` instances within a single ASG (by setting `0 < OnDemandPercentage < 100`) doesn't play well. We have been trying this setup for a few days and we eventually reverted to have separate 2 stacks: one for 100% `spot` instances and another for 100% `on-demand`. The reason is simple, ASG tries to maintain the `OnDemandPercentage`, so when `DesiredCapacity` decreases (e.g. idle instances are terminated), ASG kills "random" instances (either `on-demand` or `spot`, based on the difference between actual `OnDemandPercentage` and expected `OnDemandPercentage`) to reach a target `OnDemandPercentage`.

Maybe it will be better to use [OnDemandBaseCapacity](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_InstancesDistribution.html) instead of `OnDemandPercentage`.
Alternatively, it would probably also work well with `OnDemandPercentage` if there was a `scale in protection` enabled for all instances in ASG and removing the protection right before requesting instance termination in [terminate-instance] script (https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/linux/conf/buildkite-agent/scripts/terminate-instance)

yob:
> But, it turns that using both on-demand and spot instances within a single ASG (by setting 0 < OnDemandPercentage < 100) doesn't play well. We have been trying this setup for a few days and we eventually reverted to have separate 2 stacks: one for 100% spot instances and another for 100% on-demand. The reason is simple, ASG tries to maintain the OnDemandPercentage, so when DesiredCapacity decreases (e.g. idle instances are terminated), ASG kills "random" instances (either on-demand or spot, based on the difference between actual OnDemandPercentage and expected OnDemandPercentage) to reach a target OnDemandPercentage.

Oh. This is really good feedback to get, and it's a bummer we didn't notice it earlier.

We've also had some discussion about restoring the lifecycle hooks that were dropped in the 5.0 release. They held up instance termination until the currently running job had completed. Do you think that'd help in this case as well?

yob:
> It is actually very easy to add agent metadata/tag, which will tell if the instance is spot/on-demand. Here is how we do it in agent-bootstrap.sh:
>
>    echo "tags-from-ec2-meta-data-paths=aws:instance-life-cycle=instance-life-cycle" >> /etc/buildkite-agent/buildkite-agent.cfg

This is interesting too! Based on [these docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html) the `instance-life-cycle`  metadata key was added in IMDSv2. I might open a quick PR in the agent to collect that as standard.

PR #714, opened 2020-09-25
make build: render the mappings.yml file before aws-stack.yml

yob:
Rendering `aws-stack.yml` depends on the `mappings.yml` file existing. I previously setup that dependency directly (see PR #706), however that broke CI where we want to manually manage the mappings.yml file.

This seems like it might be a way to ensure `make build` works in our development environment (as documented in the README) while also keeping the CI pipeline working.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ env-%:
 
 # -----------------------------------------
 
-build: packer build/aws-stack.yml
+build: packer build/mappings.yml build/aws-stack.yml
 
 # Build a mapping file for a single region and image id pair
 mappings-for-linux-image: env-AWS_REGION env-IMAGE_ID
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ mappings-for-windows-image: env-AWS_REGION env-IMAGE_ID
 
 # Takes the mappings files and copies them into a generated stack template
 .PHONY: build/aws-stack.yml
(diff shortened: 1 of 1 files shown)


yob:
Here's the relevant part of the CI pipeline that #706 broke: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/87d7f1dc472cc71e407fbe4167ad6f499f36d033/.buildkite/steps/publish.sh#L26

In that script we download the correct mappings.yml file from buildkite artifacts and place it in the suitable place on disk where `make build/aws-stack.yml` will find and use it. We don't want `make` to rebuild it with `us-east-1` only!

JuanitoFatas (approved):
👍🏼  Looks good to me! Thanks for detailed explanations.

yob:
Unfortunately non-master branches only push AMIs to us-east-1 (see https://github.com/buildkite/elastic-ci-stack-for-aws/blob/87d7f1dc472cc71e407fbe4167ad6f499f36d033/.buildkite/steps/copy.sh#L103-L112) so to test this fully we have to merge and see what happens

:crossed_fingers:

yob:
I've confirmed the `aws-stack.yml` built after this merged has all regions in the mappings :tada: 

See the annotation at the top of this build: https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/2102#_

PR #718, opened 2020-09-27
AMI and Lambda Scaler support more regions

JuanitoFatas:
Support more regions for AMI and Lambda Scaler.


yob:
Looks good!

After we merge, we should probably check the rendered aws-stack.yml to confirm the new regions are in there.

One thing we need to consider: currently the stack references scaler v1.0.0, which doesn't exist in some of the lambda s3 buckets. I wonder how much work it'd be to get them copied over?

yob (approved):
May as well merge, and we can follow up with coping the lambda objects

PR #719, opened 2020-09-29
bump docker and docker-compose

yob:
* docker to 19.03.13 (linux) and 19.03.12 (windows)
* docker-compose to 1.27.4 (linux)
* docker-compose on windows is installed via chocolatey, and at the time of this commit is version 1.27.2
    
v4.5.0 of the elastic stack was released with docker 19.03.12 and docker-compose 1.25.1 - the upcoming release should use at least the same versions.
    
I've bumped them a little higher that what was in v4.5.0, but only to versions I consider relatively low risk.

### DIFF
--- a/packer/linux/scripts/install-docker.sh
+++ b/packer/linux/scripts/install-docker.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 set -eu -o pipefail
 
-DOCKER_VERSION=19.03.5
+DOCKER_VERSION=19.03.13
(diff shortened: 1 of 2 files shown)


yob:
Windows install of docker-compose happens here: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/2ad801ed582e40c7a04126a23d874f0de10d5fd4/packer/windows/scripts/install-docker.ps1#L15

chloeruka (approved):
Nice!

PR #721, opened 2020-09-29
Wrap exit handler with exit code check to resolve #720

### LINKED ISSUE #720: EC2 instance terminates when buildkite-agent exits with a non-zero exit code
**What happens:** 
1. The agent exits with status 1
2. systemd runs the `terminate-instance` script (and then poweroff) to start taking the instance out of service
3. systemd restarts the agent as configured `on-failure`

**What should happen:** 
1. The agent exits with status 1
2. The exit code is evaluated for exit != 0 and the instance is kept in service for a restart
3. systemd restarts the agent

chloeruka:
**Problem**: systemd's [`ExecStopPost`](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStopPost=) gets run in all cases, including where the service exited unexpectedly (non-zero exit code) as it is intended for cleanup behaviour.

This PR implements a script to wrap `terminate-instance` (which I did not want to modify) in a script to gracefully handle the branching logic between standard agent `exit 0` (which should terminate the instance) and non-zero exit codes (which should restart the agent and leave the instance as-is).

This is a change we should consider backporting to v4.5.

### DIFF
--- a/packer/linux/conf/buildkite-agent/scripts/service-exit-handler
+++ b/packer/linux/conf/buildkite-agent/scripts/service-exit-handler
@@ -12,4 +12,3 @@ fi
 
 echo "Service exited normally, attempting to shut down EC2 instance" | logger
 ./terminate-instance
-/bin/sudo poweroff
--- a/packer/linux/conf/buildkite-agent/scripts/service-exit-handler
+++ b/packer/linux/conf/buildkite-agent/scripts/service-exit-handler
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -euo pipefail
 
 echo "Buildkite agent service stopped, entered service-exit-handler" | logger
--- a/packer/linux/conf/buildkite-agent/scripts/service-exit-handler
+++ b/packer/linux/conf/buildkite-agent/scripts/service-exit-handler
@@ -1,0 +1,15 @@
+#!/bin/sh
+set -euo pipefail
+
+echo "Buildkite agent service stopped, entered service-exit-handler" | logger
+systemctl is-failed buildkite-agent.service | logger
+echo "Service result: $SERVICE_RESULT Exit code: $EXIT_CODE Exit status: $EXIT_STATUS" | logger
+
+if [[ $EXIT_CODE != 0 ]]; then
+  echo "Service exited abnormally, systemd will try to restart it in 5 seconds" | logger
+  exit 0
+fi
+
+echo "Service exited normally, attempting to shut down EC2 instance" | logger
+./terminate-instance
+/bin/sudo poweroff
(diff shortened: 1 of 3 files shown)


lox:
Drive by comment, but be aware that the version of systemd in Amazon Linux 2 is ancient. Caused me a lot of problems with various behaviours.

yob:
Thanks, the age of systemd isn't something I'd thought to check. I was curious, so took a look at what's on the AMIs currently generated by master:

```
# yum list installed | grep systemd
rpm-plugin-systemd-inhibit.x86_64     4.11.3-40.amzn2.0.4            installed  
systemd.x86_64                        219-57.amzn2.0.12              installed  
systemd-libs.x86_64                   219-57.amzn2.0.12              installed  
systemd-sysv.x86_64                   219-57.amzn2.0.12              installed
```


... certainly quite a bit older than my debian laptop:

```
$ dpkg -l | grep systemd | tail -n 3
ii  systemd                               246.6-1                         amd64        system and service manager
ii  systemd-sysv                          246.6-1                         amd64        system and service manager - SysV links
ii  systemd-timesyncd                     246.6-1                         amd64        minimalistic service to synchronize local time with NTP servers
```

yob (commented):
This is great :heart: 

Now that the systemd unit doesn't call `poweroff`, maybe we can fold the new script into `terminate-instance`? I can see it's not called anywhere else, so the change in behaviour won't impact any other processes.

```
$ git grep terminate-instance packer/linux/
packer/linux/conf/buildkite-agent/scripts/terminate-instance:aws autoscaling terminate-instance-in-auto-scaling-group --region "$region" --instance-id "$instance_id" "--should-decrement-desired-capacity" || true
packer/linux/conf/buildkite-agent/systemd/buildkite-agent.service:ExecStopPost=/usr/local/bin/terminate-instance
packer/linux/scripts/install-buildkite-agent.sh:sudo cp /tmp/conf/buildkite-agent/scripts/terminate-instance /usr/local/bin/terminate-instance
```

With @lox's comment in mind, have we verified that this works the way the systemd docs suggest it will? It'd be good to spin up a test stack with something that will cause the agent to exit non-zero (maybe a bad token?) and verify that:

a. `$SERVICE_RESULT` is set like we expect
b. that the systemd unit restarts the agent when the agent exits with status `1` and the `ExecStopPost` exits with status `0`

chloeruka:
> Now that the systemd unit doesn't call poweroff, maybe we can fold the new script into terminate-instance? I can see it's not called anywhere else, so the change in behaviour won't impact any other processes.

I'd prefer not to combine the scripts. The new script is deciding whether or not we _should_ terminate and requires systemd's env vars as its dependencies, the second is the implementation of _how_ to terminate and has no dependencies. Even if we don't use it anywhere else, it's nice to leave an option to stack users to control the ASG scale-in from the agent. We can't know if someone is already doing this, too.

> have we verified that this works the way the systemd docs suggest it will?

Thanks for looking into it. I thought it would work since its well documented behaviour that dates back to 2016:
https://github.com/systemd/systemd/blob/ddd6dd9f2398c01f69acf163bfdedf161d7cea0d/NEWS#L4987-L4990
Unfortunately, AL2's version of systemd is _so_ old it doesn't have it according to the version number you sighted. Oof. This will have to be shelved until a new version is available.

yob:
oh no!

chloeruka:
Closing this as systemd version needs to be >= 232. Amazon Linux 2.0.20200904.0 ships with only systemd 219. Further discussion or alternative solutions should be directed to #720

PR #727, opened 2020-10-01
update agent config: tags-from-ec2 -> tags-from-ec2-meta-data

yob:
I've noticed the following warning in log output of various stacks:

> The config option `tags-from-ec2` has been renamed to `tags-from-ec2-meta-data`. Please update your configuration.

The deprecation happened in agent 3.17.0 ([august 2019](https://github.com/buildkite/agent/pull/1067)) [1]. The stack ships with a more recent version and it seems unlikely anyone will jump through hoops to downgrade, so I think it's safe to use the newer flag.


JuanitoFatas (approved):
👍🏼  Thanks for updating this!

PR #729, opened 2020-10-02
add a 10sec pause before we request instance termination

yob:
When the buildkite-agent exits cleanly (generally after an idle timeout), we ask the ASG to terminate the instance. The termination can happen *very* quickly, oftern within 2-3 seconds.

We use the legacy awslogs tool[1] to send logs to cloudwatch, and it buffers new logs for 5 seconds (or 32Kb, whichever comes first).

This results in a race condition - often the final logs from the agent as it shutdown may not make it to cloudwatch before the instance shuts down.

Long term we assume migrating from awslogs to the new cloudwatch agent might improve this situation. However, we don't plan to attempt that migration vor the v5 release and we'd like to improve the chances customers will have access to a full agent log in clouddwatch. Adding a 10 sec sleep before requesting termination is slightly-sad, but pragmatic option.

Note that this (sadly) won't resolve the race condition that prevents other useful logs from making it to cloudwatch, like the result of calling `aws autoscaling terminate-instance-in-auto-scaling-group` and systemd during shutdown

### DIFF
--- a/packer/linux/conf/buildkite-agent/scripts/terminate-instance
+++ b/packer/linux/conf/buildkite-agent/scripts/terminate-instance
@@ -2,7 +2,13 @@
 
 set -euo pipefail
 
+echo "sleeping for 10 seconds before terminating instance to allow agent logs to drain to cloudwatch..."
+
+sleep 10
+
 instance_id=$(curl -fsSL http://169.254.169.254/latest/meta-data/instance-id)
 region=$(curl -fsSL http://169.254.169.254/latest/meta-data/placement/availability-zone | head -c -1)
+
+echo "requesting instance termination..."
 
 aws autoscaling terminate-instance-in-auto-scaling-group --region "$region" --instance-id "$instance_id" "--should-decrement-desired-capacity" || true


chloeruka (commented):
Seems sensible if time is the only thing stopping us from draining the syslog. 

I worry this might introduce some delay with scaling up in the following scenario:
1. `agent a` finishes job, terminates buildkite-agent and waits 10s
2. a new job enters the dispatch queue
3. [`buildkite-agent-scalar`](https://github.com/buildkite/buildkite-agent-scaler) sets the ASG's [desired capacity](https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-manual-scaling.html) to 1
4. ASG remains with 1/1 instance until `agent a` fully terminates (up to 10s)
5. ASG sees current count < desired count and adds another instance to replace the outgoing `agent a` instance with `agent b`

yob:
Hmmm, interesting.

That race condition can happen anyway, even without the new 10 second pause right? This might make it more likely to happen.

What's the impact? `agent a` could've been kept around to run the new job (potentially with warm caches), but `agent b` (with a cold cache) runs it instead?

chloeruka:
There's one thing I forgot to consider in my hypothetical which is that the ASG will have its desired count decremented. This might incur the lag of waiting for another cycle of the agent scaler to reset the desired count. The impact is that it'll take longer to get an active agent in the pool. 

This can be mitigated by adjusting the min-size and agent terminate-after-idle timeouts longer, so it really just comes down to performance tuning. Maybe it's not a big deal, it just popped into my head as a scenario that adding an extra delay will make slightly more likely.

chloeruka (approved):

ticky:
It looks like the combination of this and #728 means that it can’t actually terminate the agent:

```
Oct 6 00:28:00 ip-10-0-x-xxx terminate-instance: requesting instance termination...
Oct 6 00:28:02 ip-10-0-x-xxx terminate-instance: An error occurred (ValidationError) when calling the TerminateInstanceInAutoScalingGroup operation: Currently, desiredSize equals minSize (2). Terminating instance without replacement will violate group's min size constraint. Either set shouldDecrementDesiredCapacity flag to false or lower group's min size.
```

PR #736, opened 2020-10-08
Fix publish script to support beta tags

chloeruka:
Previously `publish.sh` only supported pre-releases as including "-rc" in the git tag. This change allows us to just use any "-" to specify a pre-release version (e.g. `v5.0.0-beta1`)


JuanitoFatas (approved):
👍🏼

Issue #733, opened 2020-10-06
Spot termination leaves hanging agents if "stop-agent-gracefully" blocks for more than the allotted 2 minutes


bshelton229:
When an agent is interrupted due to spot termination, if it is processing a job that doesn't finish 2 minutes after the spot termination signal is received, the ec2 instance ends up terminated in ec2 without properly deregistering the agent in the buildkite control plane. This results in the agent looking like it's still processing the job for some time (minutes?) before the connection is determined to be lost, marking the agent as gone in buildkite. I believe this to be an issue with the instance being hard terminated by ec2 while the `stop-agent-gracefully` script called by `lifecycled` is still waiting for buildkite-agent to gracefully stop.

My understanding of how spot termination is handled in what's currently in master is:

1. Lifecycled is booted up and will handle the spot termination signal from the local ec2 metadata url.
2. This triggers `stop-agent-gracefully` when the spot termination signal is received.
3. `stop-agent-gracefully` calls `systemd stop buildkite-agent` which should send SIGTERM to the agent process and block for the agent to gracefully clear all in flight jobs.
4. The `ExecStopPost` systemd command will then trigger the agent to terminate by calling `terminate-instance`, which works by calling `aws autoscaling terminate-instance-in-auto-scaling-group` with the correct derived values.

Where I believe this breaks down is if spot steps in and hard terminates while `systemctl stop "buildkite-agent"` is still blocking for graceful shutdown, resulting in a hung agent in the UI until Lost Connection is detected some minutes after the instance is gone in ec2.

## Possible Solution

`systemctl stop buildkite-agent` appears to block waiting for the process to die gracefully, so I wasn't sure what the loop waiting for pids was doing in the `stop-agent-gracefully` script, but I think a possible solution is to detect from lifecycled that this termination was due to spot and after 1.5 minutes hard-terminate the agent pids. Maybe we could do the initial SIGTERM backgrounded and loop for less than 2 minutes and if the pid still isn't gone hit it with SIGTERM again. It looks like if the agent receives subsequent SIGTERM it will hard stop? - https://github.com/buildkite/agent/blob/097d299649771ad7da5cdb445090331a249c924d/clicommand/agent_start.go#L774

I haven't looked enough into lifecycled to know if it passes some env vars telling the script which termination method it's responding to. I think we'd want to block for as long as we have to when not being terminated due to spot, which has that 2 minute window.

## Replicating the Issue

[...]

PR #737, opened 2020-10-08
Send SIGQUIT before spot termination time

bshelton229:
This addresses #733 by capturing the lifecycle transition from the first script argument. If we're not terminating due to spot the script should behave as it did before. If we are terminating due to spot, we do the following:

1. Send SIGTERM to the master agent process in a non-blocking fashion
1. Calculate the unix timestamp 20 seconds before the spot termination time and consider this our STOP_TIME
1. Check to see if the agents stop gracefully before our STOP_TIME
1.  If the agents stop on their own, exit 0. If they don't, send `SIGQUIT` and let this script exit, you only have 20 seconds before termination anyway.

## Testing

I tested this by booting up a stack and manually destroying the scaling lambda as to not have anything provision agents outside this test. I then create a spot fleet requesting a single instance using the launch template from the stack, setting the networking and other settings to what the ASG would do. This should boot up a single spot instance. You can then jump onto the instance and tail the lifecycled logs `journalctl -f -t lifecycled`. Now decrement the spot flee to 0 while your test long running job is still running on the agent. This will trigger the normal spot termination process. AWS doesn't do a great job of documenting this method, but you can find it here - https://ec2spotworkshops.com/running_spark_apps_with_emr_on_spot_instances/tracking_spot_interruptions.html#verifying-that-the-notification-works.

You should see the long running job eventually forcefully killed by `SIGQUIT` 20 seconds before the spot termination will ultimately terminate the instance. You can also repeat this test, manually cancelling the long running job to test the script detecting the graceful shutdown of the agent.

### DIFF
--- a/packer/linux/conf/buildkite-agent/scripts/stop-agent-gracefully
+++ b/packer/linux/conf/buildkite-agent/scripts/stop-agent-gracefully
@@ -4,13 +4,44 @@ set -eu -o pipefail
 # Import BUILDKITE_AGENTS_PER_INSTANCE
 eval "$(cat /var/lib/buildkite-agent/cfn-env)"
 
-echo "Stopping buildkite-agent gracefully"
-systemctl stop "buildkite-agent"
+LIFECYCLE_TRANSITION=${1:-}
 
-# Need to ensure it's the buildkite-agent user, so it doesn't match this lifecycled handler script
-while pgrep -u buildkite-agent buildkite-agent > /dev/null; do
-  echo "Waiting for all buildkite-agent processes to have stopped..."
-  sleep 5
-done
+echo "Stopping buildkite-agent gracefully due to the lifecycle transition: ${LIFECYCLE_TRANSITION}"
 
-echo "All buildkite-agent processes have stopped"
+if [ $LIFECYCLE_TRANSITION = "ec2:SPOT_INSTANCE_TERMINATION" ]; then
+  # Send SIGTERM to the main buildkite-agent process in a non-blocking manner
+  # to start a graceful shutdown
+  systemctl kill --kill-who=main --signal=SIGTERM buildkite-agent
+
+  SPOT_TERMINATION_TIME=$(curl http://169.254.169.254/latest/meta-data/spot/termination-time 2>/dev/null)
+  # Convert the RFC3339 spot termination time to a unix timestamp
+  SPOT_TERMINATION_UNIX=$(date '+%s' -d ${SPOT_TERMINATION_TIME})
+  # Subtracts 20 seconds from the published spot termination time to give the
+  # agent a little time to forcefully quit running jobs. If the agent doesn't
+  # gracefully quit before this time we will send SIGQUIT.
+  STOP_BY_UNIX=$(expr ${SPOT_TERMINATION_UNIX} - 20)
+
+  echo "Waiting for agents to quit gracefully until 20 seconds before the spot termination time: ${SPOT_TERMINATION_TIME}"
+  while [ $(date '+%s') -lt $STOP_BY_UNIX ]; do
+    if ! pgrep -u buildkite-agent buildkite-agent > /dev/null; then
+      echo "All buildkite agents have stopped gracefully"
+      exit 0
+    fi
+    sleep 1
+  done
+
+  # Agents are still running and we're close to time, so send SIGQUIT. We can let
+  # this script exit as spot will hard terminate in a matter of seconds.
+  echo "Agents are still running. Sending SIGQUIT ahead of forced spot termination at ${SPOT_TERMINATION_TIME}"
+  systemctl kill --kill-who=main --signal=SIGQUIT buildkite-agent
+else
+  systemctl stop "buildkite-agent"
+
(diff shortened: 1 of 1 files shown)


packer/linux/conf/buildkite-agent/scripts/stop-agent-gracefully
```diff
@@ -4,13 +4,44 @@ set -eu -o pipefail
...
-done
+echo "Stopping buildkite-agent gracefully due to the lifecycle transition: ${LIFECYCLE_TRANSITION}"
 
-echo "All buildkite-agent processes have stopped"
+if [ $LIFECYCLE_TRANSITION = "ec2:SPOT_INSTANCE_TERMINATION" ]; then
+  # Send SIGTERM to the main buildkite-agent process in a non-blocking manner
+  # to start a graceful shutdown
+  systemctl kill --kill-who=main --signal=SIGTERM buildkite-agent
+
+  SPOT_TERMINATION_TIME=$(curl http://169.254.169.254/latest/meta-data/spot/termination-time 2>/dev/null)
```
pda, line 16: I think we'd want curl's `--fail` option, so that if the request fails it's picked up by bash's `set -e` and the script terminates, instead of proceeding with a blank `SPOT_TERMINATION_TIME`. (I _think_ terminating on failure is the lesser evil?)

> -f, --fail
> (HTTP)  Fail  silently  (no  output at all) on server errors. This is mostly done to better enable scripts etc to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will prevent curl from outputting that and return error 22.

bshelton229: Good call @pda. I just realized lifecycled passes the termination time as the third argument to the script, so I updated this to rely on that.

chloeruka (approved):
Hey @bshelton229 – Huge thanks for your contribution. I've tested this on two versions of the stack; this PR (SHA `46669ae82`) and by backporting the `stop-agent-gracefully` script to our previous stable v4.5.0. I confirmed earlier that the same bug is present in the previous version, and that the fix applies in both cases (which makes it a good candidate for backporting to v4.5).

image

image

Tested using a simple 10min sleep job based off [buildkite/bash-example](https://github.com/buildkite/bash-example/tree/test-long-running-job)

It's somewhat of an antipattern as we've implemented but `SIGTERM` [sent twice should work](https://github.com/buildkite/agent/blob/097d299649771ad7da5cdb445090331a249c924d/clicommand/agent_start.go#L770-L786) and it might be preferable to make a small change to use that instead of `SIGQUIT`. For now, since this work fixes the immediate bug and it'd require a bit more testing, I'm going to approve this PR and merge it in 👍

bshelton229:
Thanks @chloeruka. Let me know if you'd like me to change to sending SIGTERM twice in here, rather than SIGQUIT. It's still really unclear to me how folks were reporting downgrading the stack fixing the issue. Combined with our shop thinking that `-1` agent lost connection was the normal behavior for spot termination, I'm wondering if the control plane realizes agents are defunct faster in older AMIs for reasons I can't think of (tcp keepalive?) without knowing more about how the control plane manages and monitors the connections with the agents. I may have some time today to see if that's the case or not, mostly out of curiosity.

Either way, unexpected loss of connection is hopefully an exception after handling spot timeout. One question I had is, does [pool.Stop(false)](https://github.com/buildkite/agent/blob/master/clicommand/agent_start.go#L782) send cancel to running jobs, waiting for the `cancel-grace-period`, or does it kill them more forcefully. This is the main reason I give an extra 20 seconds before spot termination, but I realize that `cancel-grace-period` is configurable, so folks could run into additional edge cases if they've bumped that up.

yob:
Thanks @bshelton229!

I think we're fine with SIGQUIT. The code paths in the agent are the same, and might be best to avoid arguing signal semantics for little gain.

We're also a bit stumped that some folks have reported this as a regression - in our testing it seems to be an issue in the 4.x series as well. We're not aware of any differences in the control plane that would explain it - in both stacks agent registration, heartbeats, and deregistration are quick HTTP calls. There's no long-lived connection that might allow the servers to rapidly notice an agent going away.

Our regression test was on 4.5 - I wonder if it's worth us trying on 4.3.5 as well. That's a very common version that has had no complaints of lost agents on spot termination. In anycase, exploring that isn't a blocker for this fix.

> One question I had is, does `pool.Stop(false)` send cancel to running jobs, waiting for the cancel-grace-period, or does it kill them more forcefully

* `pool.Stop(true)` asks the agent to stop after the current job completes, but sends no signal/request/interrupt to a running job https://github.com/buildkite/agent/blob/097d299649771ad7da5cdb445090331a249c924d/agent/agent_worker.go#L245-L292
* `pool.Stop(false)` sends an interrupt to the running job process, and after `cancel-grace-period` (default: 10 secs) does a hard-termination of the job processes (if they're still running) https://github.com/buildkite/agent/blob/097d299649771ad7da5cdb445090331a249c924d/agent/agent_worker.go#L245-L292

PR #738, opened 2020-10-09
Enable SSM access and disable SSH access.

### LINKED ISSUE #620: Proposal: Remove default external SSH access; default to SSM
This is an issue to track the discussion that started in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/598. 

Right now, SSH is open by default on Linux instances, but no such RDP access is made available for Windows. Ideally, we'd want access behavior to be consistent between the two platforms. 

There were two solutions thrown out in the PR linked above:
1. Open up RDP access by default on Windows
2. Remove SSH access from Linux and add SSM access to both platforms

It seems that consensus was moving towards option 2 at the time of issue creation.

xiaket:
1. Remove SSH keyname parameter and cleaned up a bunch of stuff.
2. Add minimal iam permissions to use SSM based access. Not using the `AmazonSSMManagedInstanceCore` as that will grant additional permission that are undesirable(`ssm:GetParameter` to `*` for example)

Fixes #620

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -40,7 +40,6 @@ Metadata:
         - ImageIdParameter
         - InstanceType
         - AgentsPerInstance
-        - KeyName
         - SpotPrice
         - SecretsBucket
         - ArtifactsBucket
@@ -88,11 +87,6 @@ Metadata:
         - EnableDockerLoginPlugin
 
 Parameters:
-  KeyName:
-    Description: Optional - SSH keypair used to access the buildkite instances, setting this will enable SSH ingress
-    Type: String
-    Default: ""
-
   BuildkiteAgentRelease:
     Type: String
     AllowedValues:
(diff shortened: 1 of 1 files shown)


xiaket:
This will hopefully resolve #620

ecktom:
Hi @xiaket, I guess we also need to have the `ssm-agent` running on the nodes, don't we?
```
    yum install -y https://s3.amazonaws.com/ec2-downloads-windows/SSMAgent/latest/linux_amd64/amazon-ssm-agent.rpm
    systemctl start amazon-ssm-agent
    systemctl enable amazon-ssm-agent
```

jamesholmes-linktree:
Yes, but there's a way to do this more easily with the packer build: https://www.packer.io/docs/builders/amazon/instance#session-manager-connections

This will build the AMI with teh agent already installed.

xiaket:
@ecktom Thanks Thomas! Am I right to believe that the ssm agent is included in Amazon Linux 2?

ecktom:
@xiaket you're right, my bad. I got confused by the Amazon EKS optimized AMIs which are build on top of Amazon Linux 2 but do not include the ssm-agent by default ;)

nitrocode:
Nice contribution! We've been using ssm to access the buildkite ec2 instances successfully as well.

For the uninitiated, it's been as simple as this.

```
brew cask install session-manager-plugin

aws ssm start-session --target i-aaabbbcccddd
```

chloeruka (approved):
Apologies for the delay in us thinking about this! SSH keys came up in conversation just recently so I'm going to put my ✅ on this, but before merging I'd be keen to grab some positive feels from @yob or @pda too.

PR #740, opened 2020-10-13
update secrets plugin submodule

yob:
The full diff is [here](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/compare/2a11ab5...48c2116e91850d94aec155f8e8870a4c687e78d7?w=1
). It's mostly docs, improved logging and tidy ups with no impact on behaviour. The one exception is we now set the AWS_METADATA_SERVICE_NUM_ATTEMPTS environment variable, so the awscli will retry failed instance metadata requests a few times.

In theory instance metadata requests are very likely to fail, however with enough instances across enough customers.... we have seen it happen. The failures are intermittent and temporary, so hopefully retries will help. More details in [this PR](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/pull/30).


JuanitoFatas (approved):
🚢

pda:
I had one concern with `AWS_METADATA_SERVICE_NUM_ATTEMPTS`… it's being exported — does it leak out of the hook and into subsequent phases of the build? Maybe that's no big deal.

chloeruka (approved):
I'm satisfied we should ship this, even if it bleeds the environment variable (it's just resiliency to transient network faults, and most of the time the first request succeeds). We can address that in a later update if it becomes an issue.

PR #744, opened 2020-10-16
bump vendored docker login plugin

yob:
We're preparing to release v5.0.0 and would like the 3 vendored plugins to be up to date.

The full diff is
[here](https://github.com/buildkite-plugins/docker-login-buildkite-plugin/compare/470352092436a09f3f9f1f45afd67aebe111af55...cb92aa6f507d9cd3527889d89b1c380843be0e70). It's mostly docs, and non-functional updates.

The one exception is that support for multiple registries has been removed. However, the way we use the embedded plugin only supports a single registry. Here's how we describe using it in the [stack README](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/e703d749e16d941878bcb6dde80dd9c9ec2c5cc1/README.md)::

> If you want to push or pull from registries such as Docker Hub or Quay you
> can use the environment hook in your secrets bucket to export the following
> environment variables:
>
>     DOCKER_LOGIN_USER="the-user-name"
>     DOCKER_LOGIN_PASSWORD="the-password"
>     DOCKER_LOGIN_SERVER="" - optional. By default it will log into Docker Hub
>
> Setting these will perform a docker login before each pipeline step is run,
> allowing you to docker push to them from within your build scripts


chloeruka (approved):
LGTM! 🚀

PR #746, opened 2020-10-19
Configure Windows service exit like the unix service

### LINKED ISSUE #712: Avoid Windows instance churn when MinSize > 0
Overall the lambda scaler (experimental and opt-in during 4.x releases, the only option on master branch) has been popular with users of the stack.

We're preparing to make it the only option in the next release, but before that we hope to improve the way it behaves on stacks with MinSize > 0.

The lambda scaler only scales the ASG stack up. Scale down is handled by the agents stopping after a configurable idle period and self-terminating the instance. When MinSize > 0, each instance self terminates after the idle period, and then if the ASG is below MinSize new instances are immediately created. This can lead to brief periods of reduced (or no) agent capacity.

It'd be nice if we had a way to mostly keep MinSize instances running without the churn.

chloeruka:
There's a known issue around the lambda scaling behaviour and when the stack's MinSize is greater than 0 (see #712). On unix, the Buildkite service will exit on idletimeout, but restart as configured in the `systemd` service configuration. On Windows, we never restart, so the instance will end and the ASG will provision us a fresh instance. This is way more disruptive and time consuming, and leads to a lot of resource churn in AWS. While we continue to investigate the underlying issue, we're hopeful that configuring the Windows service the same way as the unix one will reduce the severity of the problem.

Our Windows service is configured through `nssm` (the "non-sucking" service manager). This service manager features the following properties to help us resolve this:
- The action to take when exiting (includes to restart) https://nssm.cc/usage#exit
- The delay to wait before restarting (ms) https://nssm.cc/usage#delay

These options have straightforward command equivalents
```powershell
nssm set UT2003 AppExit Default Restart
nssm set UT2003 AppRestartDelay 0
```

I'll update this PR with results of Windows agent testing for this behaviour. I will focus only on on-demand agents for this test. This PR should resolve #712

### DIFF
--- a/packer/windows/conf/bin/bk-install-elastic-stack.ps1
+++ b/packer/windows/conf/bin/bk-install-elastic-stack.ps1
@@ -170,7 +170,7 @@ nssm set buildkite-agent AppEnvironmentExtra :HOME=C:\buildkite-agent
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 nssm set buildkite-agent AppExit Default Restart
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
-nssm set buildkite-agent AppRestartDelay 5000
+nssm set buildkite-agent AppRestartDelay 10000
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 nssm set buildkite-agent AppEvents Exit/Post "powershell C:\buildkite-agent\bin\terminate-instance.ps1"
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
--- a/packer/windows/conf/bin/bk-install-elastic-stack.ps1
+++ b/packer/windows/conf/bin/bk-install-elastic-stack.ps1
@@ -168,7 +168,9 @@ nssm set buildkite-agent AppStderr C:\buildkite-agent\buildkite-agent.log
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 nssm set buildkite-agent AppEnvironmentExtra :HOME=C:\buildkite-agent
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
-nssm set buildkite-agent AppExit Default Exit
+nssm set buildkite-agent AppExit Default Restart
+If ($lastexitcode -ne 0) { Exit $lastexitcode }
+nssm set buildkite-agent AppRestartDelay 5000
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 nssm set buildkite-agent AppEvents Exit/Post "powershell C:\buildkite-agent\bin\terminate-instance.ps1"
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
--- a/packer/windows/conf/bin/bk-install-elastic-stack.ps1
+++ b/packer/windows/conf/bin/bk-install-elastic-stack.ps1
@@ -172,7 +172,7 @@ nssm set buildkite-agent AppExit Default Restart
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 nssm set buildkite-agent AppRestartDelay 5000
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
-nssm set buildkite-agent AppEvents Exit/Post "powershell C:\buildkite-agent\bin\terminate-instance.ps1 3>&1 2>&1 > C:\buildkite-agent\buildkite-agent.log"
+nssm set buildkite-agent AppEvents Exit/Post "powershell C:\buildkite-agent\bin\terminate-instance.ps1"
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 
 Restart-Service buildkite-agent
--- a/packer/windows/conf/bin/bk-install-elastic-stack.ps1
+++ b/packer/windows/conf/bin/bk-install-elastic-stack.ps1
@@ -172,7 +172,7 @@ nssm set buildkite-agent AppExit Default Restart
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 nssm set buildkite-agent AppRestartDelay 5000
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
-nssm set buildkite-agent AppEvents Exit/Post "powershell C:\buildkite-agent\bin\terminate-instance.ps1"
+nssm set buildkite-agent AppEvents Exit/Post "powershell C:\buildkite-agent\bin\terminate-instance.ps1 3>&1 2>&1 > C:\buildkite-agent\buildkite-agent.log"
 If ($lastexitcode -ne 0) { Exit $lastexitcode }
 
 Restart-Service buildkite-agent
--- a/packer/windows/conf/buildkite-agent/scripts/terminate-instance.ps1
+++ b/packer/windows/conf/buildkite-agent/scripts/terminate-instance.ps1
@@ -9,3 +9,11 @@ $Region = (Invoke-WebRequest -UseBasicParsing http://169.254.169.254/latest/meta
 
 Write-Output "terminate-instance: requesting instance termination..."
 aws autoscaling terminate-instance-in-auto-scaling-group --region "$Region" --instance-id "$InstanceId" "--should-decrement-desired-capacity"
+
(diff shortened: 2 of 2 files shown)


chloeruka:
Unfortunately this first pass didn't work. The win agent restarts and reconnects after 5 minutes as expected, but the instance shuts down anyway. Need to investigate more.

```
2020-10-20 03:58:00 NOTICE Starting buildkite-agent v3.24.0 with PID: 1556
2020-10-20 03:58:01 NOTICE The agent source code can be found here: https://github.com/buildkite/agent/v3
2020-10-20 03:58:01 NOTICE For questions and support, email us at: hello@buildkite.com
2020-10-20 03:58:01 INFO   Configuration loaded path=C:\buildkite-agent\buildkite-agent.cfg
2020-10-20 03:58:01 INFO   Running builds within a pseudoterminal (PTY) has been disabled
2020-10-20 03:58:01 INFO   Agents will disconnect after 600 seconds of inactivity
2020-10-20 03:58:01 INFO   Fetching EC2 meta-data...
2020-10-20 03:58:01 INFO   Successfully fetched EC2 meta-data
2020-10-20 03:58:01 INFO   Registering agent with Buildkite...
2020-10-20 03:58:01 INFO   Successfully registered agent "windows-churn-fix-i-00e58a76787faa6f0-1" with tags [queue=windows, docker=19.03.12, stack=windows-churn-fix, buildkite-aws-stack=v5.0.0-beta1-12-g9b676577, aws:instance-type=t3.small, aws:ami-id=ami-037db01d7c5e3ea32, aws:instance-id=i-00e58a76787faa6f0]
2020-10-20 03:58:01 INFO   Starting 1 Agent(s)
2020-10-20 03:58:01 INFO   You can press Ctrl-C to stop the agents
2020-10-20 03:58:01 INFO   windows-churn-fix-i-00e58a76787faa6f0-1 Connecting to Buildkite... 
WARNING: Waiting for service 'buildkite-agent (buildkite-agent)' to start...
DEBUG:  184+  >>>> cfn-signal `
DEBUG:  188+   --exit-code 0 ; if ( >>>> -not $?) {
DEBUG:  191+      >>>> Write-Output "Signal failed"
Signal failed
2020-10-20 03:58:01 INFO   windows-churn-fix-i-00e58a76787faa6f0-1 Waiting for work... 
DEBUG:  194+  >>>> Set-PSDebug -Off
2020-10-20 04:08:01 INFO   windows-churn-fix-i-00e58a76787faa6f0-1 All agents have been idle for 600 seconds. Disconnecting... 
2020-10-20 04:08:01 INFO   windows-churn-fix-i-00e58a76787faa6f0-1 Disconnecting... 
2020-10-20 04:08:06 NOTICE Starting buildkite-agent v3.24.0 with PID: 1592
2020-10-20 04:08:06 NOTICE The agent source code can be found here: https://github.com/buildkite/agent/v3
2020-10-20 04:08:06 NOTICE For questions and support, email us at: hello@buildkite.com
2020-10-20 04:08:06 INFO   Configuration loaded path=C:\buildkite-agent\buildkite-agent.cfg
2020-10-20 04:08:06 INFO   Running builds within a pseudoterminal (PTY) has been disabled
2020-10-20 04:08:06 INFO   Agents will disconnect after 600 seconds of inactivity
2020-10-20 04:08:06 INFO   Fetching EC2 meta-data...
2020-10-20 04:08:06 INFO   Successfully fetched EC2 meta-data
2020-10-20 04:08:06 INFO   Registering agent with Buildkite...
2020-10-20 04:08:06 INFO   Successfully registered agent "windows-churn-fix-i-00e58a76787faa6f0-1" with tags [queue=windows, docker=19.03.12, stack=windows-churn-fix, buildkite-aws-stack=v5.0.0-beta1-12-g9b676577, aws:instance-id=i-00e58a76787faa6f0, aws:instance-type=t3.small, aws:ami-id=ami-037db01d7c5e3ea32]
2020-10-20 04:08:06 INFO   Starting 1 Agent(s)
2020-10-20 04:08:06 INFO   You can press Ctrl-C to stop the agents
2020-10-20 04:08:06 INFO   windows-churn-fix-i-00e58a76787faa6f0-1 Connecting to Buildkite... 
2020-10-20 04:08:06 INFO   windows-churn-fix-i-00e58a76787faa6f0-1 Waiting for work... 
```

chloeruka:
Removing the Stop-Computer directive worked. We expect to see a bunch of buildkite-agents starting and expiring after idle timeouts (same as unix) and we see that in the logs:

image

chloeruka:
I'm happy with this as is, but the trade-off is that "stop agent" on the UI will no longer kill the instance behind the agent. It'll just kill the agent instance and cause it to restart after 5 seconds.

yob:
It's neat to see how similar that terminate-instance scripts are in bash and powershell.

In the case where the `aws autoscaling terminate-instance` command is successful, might this result in the agent process restarting just as the instance is terminated by the ASG?

yob:
I don't love this, but if my guess above is an actual issue then maybe this is an tolerable workaround/hack while we working making the agent itself responsible for triggering the instance termination.

I note the [nssm readme says](https://github.com/rticommunity/nssm/blob/master/README.txt):

> Some hooks are run synchronously and some are run asynchronously.  Hooks prefixed with an *asterisk are run synchronously.  NSSM will wait for these hooks to complete before continuing its work.
>
>   Event: Exit - Triggered when the application exits.
>   *Action: Post - Called after NSSM has cleaned up the application.

So it looks like the agent won't restart until after the Exit/Post hook has completed. If we update `terminate-instance.ps1` to have a ~30 seconds sleep after a **successful** call to `aws autoscaling terminate-instance`, that should mean the instance is terminated well before the agent process restarts and is assigned any work.

chloeruka:
> In the case where the `aws autoscaling terminate-instance` command is successful, might this result in the agent process restarting just as the instance is terminated by the ASG?

My understanding is no, because the agent waits 5 minutes to restart and even the slowest instance termination signal should reach us by then. We could guarantee it by signalling NSSM not to restart if the instance is to be brought out of service (I'd replicate this on unix with a systemd signal too)

yob:
I created a test stack with MinSize:1 and waited 10 minutes for the agent idle timeout. It worked! The agent was restarted and it matched the linux stack behaviour of restarting the agent without stopping the instance.

Annoyingly, I get the agent output (in the `/buildkite/buildkite-agent` log group) but not the logging output from `terminate-instance.ps1`. Is it possible to see that output as well?

Screenshot from 2020-10-26 13-24-24

However, I also watched what happened when an agent hit the idle timeout to confirm the instance shutdown as expected. The instance did shutdown, but a new agent process was able to start and register to receive work in the meantime. I think this means there's a race condition and it's possible a running job might be assigned to an agent just before it's hard terminated? Does this suggest the `nssm stop buildkite-agent` call isn't preventing the agent from restarting?

Screenshot from 2020-10-26 13-34-53

chloeruka:
> I created a test stack with MinSize:1 and waited 10 minutes for the agent idle timeout. It worked! The agent was restarted and it matched the linux stack behaviour of restarting the agent without stopping the instance.
> 
> Annoyingly, I get the agent output (in the `/buildkite/buildkite-agent` log group) but not the logging output from `terminate-instance.ps1`. Is it possible to see that output as well?

Possibly, I had to redirect stdio to get it, and it seemed to interfere with the function of the terminate-instance script. I can try again now that I have a little time.

> I think this means there's a race condition and it's possible a running job might be assigned to an agent just before it's hard terminated? Does this suggest the `nssm stop buildkite-agent` call isn't preventing the agent from restarting?

I think the 10 second wait might have something to do with that. If NSSM doesn't work the way that I hoped (block on `AppEvents Exit/Post`), then we might have to either lengthen the restart timeout (more agent downtime) or shorten/remove that delay.

yob:
> I think the 10 second wait might have something to do with that. If NSSM doesn't work the way that I hoped (block on AppEvents Exit/Post), then we might have to either lengthen the restart timeout (more agent downtime) or shorten/remove that delay.

I might do a test with the 10 second pause removed to see what happens

yob:
I just noticed the agent timestamps in both screenshots above show the agent restarting less than a second after it shutdown. WAT

yob (approved):

Issue #700, opened 2020-08-20
Entering a single value for AvailabilityZones fails


jmchuster:
If i try to set the value for `AvailabilityZones` to a single zone, say `us-west-2a`, it fails with

```
There was an error creating this change set
Template error: Fn::Select cannot select nonexistent value at index 1
```

I only need a single availability zone here, because the auto-scale group will try rebalancing instances across zones, killing my long-running jobs (24 hours +).

sj26:
Oh, weird, sorry! Sounds like a bug. Looks like we presume two availability zones in which to create subnets:

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/aws-stack.yml#L542-L570

What happens if you specify `us-west-2a,us-west-2a`? Is that a reasonable workaround?

Another possible workaround, could you suspend the rebalance scaling processes? This wouldn't stick around between updates, though.

Otherwise, sounds like a bug to be fixed in the CloudFormation, somehow.

lox:
> Otherwise, sounds like a bug to be fixed in the CloudFormation, somehow.

Unfortunately it can't 😢 There is no way to vary CloudFormation templates based on the length of input. It's extremely frustrating.

sj26:
@lox can we list a few and make them conditional on the presence of the az?

lox:
This is how I approached it in an experiment @sj26:

https://github.com/buildkite/elastic-ci-stack-for-aws-ecs/blob/master/templates/vpc/template.yaml#L9
https://github.com/buildkite/elastic-ci-stack-for-aws-ecs/blob/master/templates/vpc/template.yaml#L56-L66

I've been petitioning AWS for `Fn::Size` for 3+ years 😢

sj26:
... oh my

sj26:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-select.html
![... you should be certain that the index you choose is valid ...](https://user-images.githubusercontent.com/14028/90853983-0a57b400-e3bf-11ea-9b76-7045574e044b.png)
_How?!_

sj26:
I think I'm going too deep
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-transform.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html#template-macros-lambda-interface

[...]

PR #751, opened 2020-10-21
The agent autoscaling group should never rebalance availability zones

yob:
By default, if the two availability zones in the agent ASG become significantly unbalanced, the ASG will terminate some instances in the larger AZ and start some new ones in the smaller AZ.

That's helpful for an ASG serving web requests, but it's not very helpful for our agent-shared workloads - the instance termination can disrupt running jobs. With the new lambda based scaler, each instance is responsible for terminating itself and the AZs become unbalanced very easily. That's not much of a problem though - the larger AZ is likely to reduce in size relatively soon, and subsequent scale-outs will restore the balance (for a while).

Sadly there's no way to suspend the AZRebalance process via cloudformation, so I held my nose and implemented it using a custom resource. It's not as ugly as I feared, mainly because it's possible to provide the required lambda function inline.

An alternative approach would be to have our buildkite-agent-scaler lambda check the AZRebalance status each time it loops and suspend the process if required. I thought this approach might be good enough for now, and we could try the scaler option down the track if we need to.

Some resources I found useful:

1. https://www.alexdebrie.com/posts/cloudformation-custom-resources/
2. https://gist.github.com/atward/9573b9fbd3bfd6c453158c28356bec05
3. https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_SuspendProcesses.html
4. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-lambda-function-code-cfnresponsemodule.html

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -945,23 +945,23 @@ Resources:
       AssumeRolePolicyDocument:
         Version: 2012-10-17
         Statement:
-        - Effect: Allow
-          Principal:
-            Service:
-              - lambda.amazonaws.com
-          Action:
-          - sts:AssumeRole          
+          - Effect: Allow
+            Principal:
+              Service:
+                - lambda.amazonaws.com
+            Action:
+              - sts:AssumeRole
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
       Policies:
-      - PolicyName: AsgProcessModification
-        PolicyDocument:
-          Version: 2012-10-17
-          Statement:
-          - Effect: Allow
-            Action:
-            - 'autoscaling:SuspendProcesses'
-            Resource: !Sub arn:${AWS::Partition}:autoscaling:${AWS::Region}:${AWS::AccountId}:autoScalingGroup:*:autoScalingGroupName/${AgentAutoScaleGroup}
+        - PolicyName: AsgProcessModification
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - 'autoscaling:SuspendProcesses'
+                Resource: !Sub arn:${AWS::Partition}:autoscaling:${AWS::Region}:${AWS::AccountId}:autoScalingGroup:*:autoScalingGroupName/${AgentAutoScaleGroup}
 
   AzRebalancingSuspenderFunction:
     Type: AWS::Lambda::Function
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -959,9 +959,7 @@ Resources:
           - Effect: Allow
             Action:
             - 'autoscaling:SuspendProcesses'
-            Resource: !Sub
-              - "arn:aws:autoscaling:*:*:autoScalingGroup:*:autoScalingGroupName/${LogicalGroupName}"
-              - LogicalGroupName: !Ref AgentAutoScaleGroup
+            Resource: !Sub arn:${AWS::Partition}:autoscaling:${AWS::Region}:${AWS::AccountId}:autoScalingGroup:*:autoScalingGroupName/${AgentAutoScaleGroup}
 
   AzRebalancingSuspenderFunction:
     Type: AWS::Lambda::Function
@@ -1127,4 +1125,3 @@ Resources:
       Action: "lambda:InvokeFunction"
       Principal: "events.amazonaws.com"
       SourceArn: !GetAtt AutoscalingLambdaScheduledRule.Arn
-
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -939,6 +939,59 @@ Resources:
       AutoScalingReplacingUpdate:
         WillReplace: true
 
+  AsgProcessSuspenderRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+        - Action: ['sts:AssumeRole']
+          Effect: Allow
+          Principal:
+            Service: [lambda.amazonaws.com]
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+      Policies:
+      - PolicyName: AsgProcessModification
+        PolicyDocument:
+          Version: 2012-10-17
+          Statement:
+          - Effect: Allow
+            Action:
+            - 'autoscaling:SuspendProcesses'
+            Resource: '*'
+
+  AzRebalancingSuspenderFunction:
+    Type: AWS::Lambda::Function
+    Properties:
+      Description: 'Disables AZ Rebalancing on the agent ASG'
+      Code:
+        ZipFile: |
+          import cfnresponse
+          import boto3
+          def handler(event, context):
+            try:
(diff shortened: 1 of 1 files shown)


yob:
Here's a screenshot of the ASG in a newly created stack, built with this PR:

Screenshot from 2020-10-22 01-01-08

I've tested stack creation and deletion with the new custom resource, and both worked as expected.

templates/aws-stack.yml
```diff
@@ -939,6 +939,59 @@ Resources:
...
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+      Policies:
+      - PolicyName: AsgProcessModification
+        PolicyDocument:
+          Version: 2012-10-17
+          Statement:
+          - Effect: Allow
+            Action:
+            - 'autoscaling:SuspendProcesses'
+            Resource: '*'
```
yob, line 962: is this OK? Is there a way to only give permission to the specific ASG in this stack?

yob:
Related: #700

templates/aws-stack.yml
lox: I would think it should be possible!

lox:
This is great @yob, pleased at actually how brief the custom resource ends up being.

templates/aws-stack.yml
yob: I tried this:

```diff
diff --git a/templates/aws-stack.yml b/templates/aws-stack.yml
index 40490c9..a8de337 100644
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -959,7 +959,7 @@ Resources:
           - Effect: Allow
             Action:
             - 'autoscaling:SuspendProcesses'
-            Resource: '*'
+            Resource: !GetAtt AgentAutoScaleGroup.Arn
```


.. but sadly it seems the `AWS::AutoScaling::AutoScalingGroup` type doesn't have an Arn property we can get:


```
$ STACK_NAME=jh-azrebalance7 aws-vault exec bk-sandbox-admin make create-stack      
...

An error occurred (ValidationError) when calling the CreateStack operation: Template error: resource AgentAutoScaleGroup does not support attribute type Arn in Fn::GetAtt
make: *** [Makefile:114: create-stack] Error 255
```

yob: https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/548

> I would also like to see this implemented so we can write IAM policies that limit access to a specific autoscaling group - otherwise, there is no way to target a specific group (as far as I can tell there is no way to get the UUID part of the group ARN and IAM won't take a * there).

.. and then:

> Correction - Targetting specific autoscaling groups by their "friendly name" can work if we use wildcards in place of the region and account id (instead of just leaving these empty - like we do with other resources such as S3 objects). To this does work: `!Sub arn:aws:autoscaling:*:*:autoScalingGroup:*:autoScalingGroupName/${LogicalGroupName}`

pda: Huh yeah, nothing but “friendly name”:

image

The wildcard approach is pretty legit, but you can be specific on region & account:

```
Resource: !Sub arn:aws: autoscaling:${AWS::Region}:${AWS::AccountId}:autoScalingGroup:…
```

templates/aws-stack.yml
```diff
@@ -939,6 +939,61 @@ Resources:
...
+      - PolicyName: AsgProcessModification
+        PolicyDocument:
+          Version: 2012-10-17
+          Statement:
+          - Effect: Allow
+            Action:
+            - 'autoscaling:SuspendProcesses'
+            Resource: !Sub
+              - "arn:aws:autoscaling:*:*:autoScalingGroup:*:autoScalingGroupName/${LogicalGroupName}"
+              - LogicalGroupName: !Ref AgentAutoScaleGroup
```
pda, line 964:
```suggestion
            Resource: !Sub arn:${AWS::Partition}:autoscaling:${AWS::Region}:${AWS::AccountId}:autoScalingGroup:*:autoScalingGroupName/${AgentAutoScaleGroup}
```

(untested but should be good)

templates/aws-stack.yml
```diff
@@ -939,6 +939,61 @@ Resources:
...
+  AsgProcessSuspenderRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+        - Action: ['sts:AssumeRole']
+          Effect: Allow
+          Principal:
+            Service: [lambda.amazonaws.com]
```
pda, line 951: We use this style elsewhere in this template:

```suggestion
        - Effect: Allow
          Principal:
            Service:
              - lambda.amazonaws.com
          Action:
          - sts:AssumeRole          
```

templates/aws-stack.yml
```diff
@@ -939,6 +939,61 @@ Resources:
...
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole
+      Policies:
+      - PolicyName: AsgProcessModification
+        PolicyDocument:
+          Version: 2012-10-17
+          Statement:
+          - Effect: Allow
+            Action:
+            - 'autoscaling:SuspendProcesses'
+            Resource: !Sub arn:${AWS::Partition}:autoscaling:${AWS::Region}:${AWS::AccountId}:autoScalingGroup:*:autoScalingGroupName/${AgentAutoScaleGroup}
```
pda, line 964: Apologies for the trivial YAML linting comments 😅
I can never decide how much multi-line YAML arrays should be indented, there's no community consensus/standard, and I think this file uses both styles. But this new code block uses both, maybe we should pick one :)
Personally I'm for “the hyphen is indented” ala:
```yaml
Key:
  - one
  - two
  - three
```

templates/aws-stack.yml
```diff
@@ -939,6 +939,61 @@ Resources:
...
+          import cfnresponse
+          import boto3
+          def handler(event, context):
+            try:
+              if event['RequestType'] == 'Delete':
+                cfnresponse.send(event, context, cfnresponse.SUCCESS, {}, "CustomResourcePhysicalID")
+              else:
+                client = boto3.client('autoscaling')
+                props = event['ResourceProperties']
+                response = client.suspend_processes(AutoScalingGroupName=props['AutoScalingGroupName'], ScalingProcesses=['AZRebalance'])
```
pda, line 981: Do we care about checking the `response`? Logging any errors etc? Maybe not.

yob: I considered that. Based on [the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/autoscaling.html#AutoScaling.Client.suspend_processes) it seems like there may not be any error conditions we care about that aren't raised as exceptions.

templates/aws-stack.yml
yob: good point, I've pushed some indenting changes

templates/aws-stack.yml
chloeruka: API responds with either none, or an exception either [suspend_process specific](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_SuspendProcesses.html) or one of the [common API errors](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/CommonErrors.html). since we log exceptions below, we should be all good if it went boom.

chloeruka (approved):
Looks good to me!

pda:
Nice 👍🏼

PR #755, opened 2020-10-26
add timestamps to packer output

yob:
The linux build takes ~4 minutes, the windows build takes ~25 minutes.

**WHHYYYYYY?**


chloeruka (approved):

lox:
I'm going to bet "something something sysprep".

yob:
I had a quick look, and "it's the vibe". Every bit is a little bit slower than the linux equivalent, accumulating into a big overall differences. Some proper research time will be required if we want to really look for time savings, unfortunately.

PR #756, opened 2020-10-26
always use a fresh docker config for each step

yob:
This improves job isolation, preventing jobs from relying on authentication configured in other jobs.

This is based on PR #678, but we've removed the stack parameter because the v5.0.0 release is pending and we're comfortable making a small breaking change.


chloeruka (approved):
Straightforward change to improve step isolation. We've discussed this and decided it's better to make this (possibly) breaking change now as part of v5 release, and to bugfix it later if necessary. Original PR (#678) says this has been working well for them for quite some time.

lox:
Couldn't we do this in the `environment` hook, or a bootstrap script?

yob:
@lox are you suggesting this doesn't need to be baked in for everyone?

packer/linux/conf/buildkite-agent/hooks/pre-exit
```diff
@@ -2,6 +2,11 @@
 
 set -eu -o pipefail
 
+# clean up our temporary docker config
+if [[ -n "${BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY:-}" ]]; then
+  rmdir "$BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY"
+fi
```
yob, line 8: do you think it's worth checking if the directory exists as well? It shouldn't matter, but maybe there's a very small chance a customer build script will have deleted it fist (maybe by cleaning out the system tmp dir or something).

chloeruka: Yeah, added a quick commit.

yob: oh, this failed in the build too. Looks like it needs a `-rf`

yob: interesting - it looks like `rmdir` will only delete empty directories. Maybe we need to use `rm`? Or I guess we could delete the contents first.

PR #690, opened 2020-06-11
Add support for arm64

galak:
Add ability to specify ARCH=arm64 to get a deployment for arm64

PR #758, opened 2020-10-29
Experimental support for running linux ARM instances

yob:
AWS has a growing number of ARM based instance types, using their gravitron and gravitron2 CPUs.

This expands on PR #690 by @galak, changing the CI pipeline to build an `linux/arm64` AMI, and ensuring that AMI is selected when a stack is deployed with OS:linux and and ARM instance type.

A few noteworthy things:

* On linux/amd64 we install docker-compose using a binary download. Sadly there's no official arm64 binary published (see https://github.com/docker/compose/issues/7472) so I've fallen back on installing via pip (for arm64 only). This works, but it's gross to install it differently depending on arch. Hopefully we can go back to binary installs if upstream starts publishing an arm64 version
* I'm not super happy with the changes to the CI pipeline scripts yet. They work, but they feel messy.
* Explicitly listing the ARM instance types feels fragile - we'll have to update it when future instance types are announced. I'm not sure how else we could do it though

Agents running on an ARM elastic stack look like this on buildkite.com:

Screenshot from 2020-10-29 11-30-26

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -132,7 +132,7 @@ steps:
     depends_on: "linux-launch"
 
   - id: "packer-linux-arm64"
-    name: ":packer: :linux: :arm:"
+    name: ":packer: :linux: 💪"
     command: .buildkite/steps/packer.sh linux arm64
     timeout_in_minutes: 60
     retry: { automatic: { limit: 3 } }
@@ -144,7 +144,7 @@ steps:
       - "s3secrets-helper-linux-arm64"
 
   - id: "linux-arm64-launch"
-    name: ":cloudformation: :linux: :arm: Launch"
+    name: ":cloudformation: :linux: 💪 Launch"
     command: .buildkite/steps/launch.sh linux arm64
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
@@ -152,7 +152,7 @@ steps:
     depends_on: "packer-linux-arm64"
 
   - id: "linux-arm64-test"
-    name: ":cloudformation: :linux: :arm: Test"
+    name: ":cloudformation: :linux: 💪 Test"
     command: "goss validate --format documentation"
     timeout_in_minutes: 5
     agents:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -85,7 +85,7 @@ steps:
       - "bats-tests"
       - "s3secrets-helper-windows-amd64"
 
-  - id: "windows-launch"
+  - id: "launch-windows"
     name: ":cloudformation: :windows: Launch"
     command: .buildkite/steps/launch.sh windows
     agents:
@@ -93,17 +93,17 @@ steps:
     artifact_paths: "build/aws-stack.yml"
     depends_on: "packer-windows"
 
-  - id: "windows-test"
+  - id: "test-windows"
     name: ":cloudformation: :windows: Test"
     command: "docker info"
     timeout_in_minutes: 5
     agents:
       stack: "buildkite-aws-stack-test-windows-amd64-${BUILDKITE_BUILD_NUMBER}"
       queue: "testqueue-windows-amd64-${BUILDKITE_BUILD_NUMBER}"
-    depends_on: "windows-launch"
+    depends_on: "launch-windows"
 
-  - id: "packer-linux"
-    name: ":packer: :linux:"
+  - id: "packer-linux-amd64"
+    name: ":packer: :linux: AMD64"
     command: .buildkite/steps/packer.sh linux
     timeout_in_minutes: 60
     retry: { automatic: { limit: 3 } }
@@ -114,25 +114,25 @@ steps:
       - "bats-tests"
       - "s3secrets-helper-linux-amd64"
 
-  - id: "linux-launch"
-    name: ":cloudformation: :linux: Launch"
+  - id: "launch-linux-amd64"
+    name: ":cloudformation: :linux: AMD64 Launch"
     command: .buildkite/steps/launch.sh linux
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
     artifact_paths: "build/aws-stack.yml"
-    depends_on: "packer-linux"
+    depends_on: "packer-linux-amd64"
 
-  - id: "linux-test"
-    name: ":cloudformation: :linux: Test"
+  - id: "test-linux-amd64"
+    name: ":cloudformation: :linux: AMD64 Test"
     command: "goss validate --format documentation"
     timeout_in_minutes: 5
     agents:
       stack: "buildkite-aws-stack-test-linux-amd64-${BUILDKITE_BUILD_NUMBER}"
       queue: "testqueue-linux-amd64-${BUILDKITE_BUILD_NUMBER}"
-    depends_on: "linux-launch"
+    depends_on: "launch-linux-amd64"
 
   - id: "packer-linux-arm64"
-    name: ":packer: :linux: 💪"
+    name: ":packer: :linux: ARM64"
     command: .buildkite/steps/packer.sh linux arm64
     timeout_in_minutes: 60
     retry: { automatic: { limit: 3 } }
@@ -143,22 +143,22 @@ steps:
       - "bats-tests"
       - "s3secrets-helper-linux-arm64"
 
-  - id: "linux-arm64-launch"
-    name: ":cloudformation: :linux: 💪 Launch"
+  - id: "launch-linux-arm64"
+    name: ":cloudformation: :linux: ARM64 Launch"
     command: .buildkite/steps/launch.sh linux arm64
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
     artifact_paths: "build/aws-stack.yml"
     depends_on: "packer-linux-arm64"
 
-  - id: "linux-arm64-test"
-    name: ":cloudformation: :linux: 💪 Test"
+  - id: "test-linux-arm64"
+    name: ":cloudformation: :linux: ARM64 Test"
     command: "goss validate --format documentation"
     timeout_in_minutes: 5
     agents:
       stack: "buildkite-aws-stack-test-linux-arm64-${BUILDKITE_BUILD_NUMBER}"
       queue: "testqueue-linux-arm64-${BUILDKITE_BUILD_NUMBER}"
-    depends_on: "linux-arm64-launch"
+    depends_on: "launch-linux-arm64"
 
   - id: "copy-ami"
     name: ":cloudformation: 🚚 🌎"
@@ -167,9 +167,9 @@ steps:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
     artifact_paths: "build/mappings.yml"
     depends_on:
-      - "linux-test"
-      - "linux-arm64-test"
-      - "windows-test"
+      - "test-linux-amd64"
+      - "test-linux-arm64"
+      - "test-windows"
 
   - id: "publish"
     name: ":cloudformation: :rocket:"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -35,6 +35,25 @@ steps:
     artifact_paths:
       - build/s3secrets-helper-linux-amd64
 
+  - id: "s3secrets-helper-linux-arm64"
+    name: ":golang: :linux: s3secrets-helper-linux-arm64"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
+    plugins:
+      docker#v3.7.0:
+        image: "golang:1.15"
+        mount-checkout: false
+        volumes:
+          - "./build:/build:rw"
+          - "./plugins/secrets/s3secrets-helper:/s3secrets-helper:ro"
+        workdir: /s3secrets-helper
+        environment:
+          - "GOOS=linux"
+          - "GOARCH=arm64"
+        command: ["go", "build", "-o", "/build/s3secrets-helper-linux-arm64"]
+    artifact_paths:
+      - build/s3secrets-helper-linux-arm64
+
   - id: "s3secrets-helper-windows-amd64"
     name: ":golang: :windows: s3secrets-helper-windows-amd64.exe"
     agents:
@@ -79,8 +98,8 @@ steps:
     command: "docker info"
     timeout_in_minutes: 5
     agents:
-      stack: "buildkite-aws-stack-test-windows-${BUILDKITE_BUILD_NUMBER}"
-      queue: "testqueue-windows-${BUILDKITE_BUILD_NUMBER}"
+      stack: "buildkite-aws-stack-test-windows-amd64-${BUILDKITE_BUILD_NUMBER}"
+      queue: "testqueue-windows-amd64-${BUILDKITE_BUILD_NUMBER}"
     depends_on: "windows-launch"
 
   - id: "packer-linux"
@@ -108,9 +127,38 @@ steps:
     command: "goss validate --format documentation"
     timeout_in_minutes: 5
     agents:
-      stack: "buildkite-aws-stack-test-linux-${BUILDKITE_BUILD_NUMBER}"
-      queue: "testqueue-linux-${BUILDKITE_BUILD_NUMBER}"
+      stack: "buildkite-aws-stack-test-linux-amd64-${BUILDKITE_BUILD_NUMBER}"
+      queue: "testqueue-linux-amd64-${BUILDKITE_BUILD_NUMBER}"
     depends_on: "linux-launch"
+
+  - id: "packer-linux-arm64"
+    name: ":packer: :linux: :arm:"
+    command: .buildkite/steps/packer.sh linux arm64
+    timeout_in_minutes: 60
+    retry: { automatic: { limit: 3 } }
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
+    depends_on:
+      - "lint"
+      - "bats-tests"
+      - "s3secrets-helper-linux-arm64"
+
+  - id: "linux-arm64-launch"
+    name: ":cloudformation: :linux: :arm: Launch"
+    command: .buildkite/steps/launch.sh linux arm64
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
+    artifact_paths: "build/aws-stack.yml"
+    depends_on: "packer-linux-arm64"
+
+  - id: "linux-arm64-test"
+    name: ":cloudformation: :linux: :arm: Test"
+    command: "goss validate --format documentation"
+    timeout_in_minutes: 5
+    agents:
+      stack: "buildkite-aws-stack-test-linux-arm64-${BUILDKITE_BUILD_NUMBER}"
+      queue: "testqueue-linux-arm64-${BUILDKITE_BUILD_NUMBER}"
+    depends_on: "linux-arm64-launch"
 
   - id: "copy-ami"
     name: ":cloudformation: 🚚 🌎"
@@ -120,6 +168,7 @@ steps:
     artifact_paths: "build/mappings.yml"
     depends_on:
       - "linux-test"
+      - "linux-arm64-test"
       - "windows-test"
 
   - id: "publish"
--- a/.buildkite/steps/copy.sh
+++ b/.buildkite/steps/copy.sh
@@ -138,13 +138,13 @@ windows_amd64_source_image_name=$(get_image_name "$windows_amd64_source_image_id
 # Copy to all other regions
 for region in ${ALL_REGIONS[*]}; do
   if [[ $region != "$source_region" ]] ; then
-    echo "--- Copying :linux: $linux_amd64_source_image_id to $region" >&2
+    echo "--- :linux: Copying Linux AMD64 $linux_amd64_source_image_id to $region" >&2
     IMAGES+=("$(copy_ami_to_region "$linux_amd64_source_image_id" "$source_region" "$region" "${linux_amd64_source_image_name}-${region}")")
 
-    echo "--- Copying :linux: :arm: $linux_arm64_source_image_id to $region" >&2
+    echo "--- :linux: Copying Linux ARM64 $linux_arm64_source_image_id to $region" >&2
     IMAGES+=("$(copy_ami_to_region "$linux_arm64_source_image_id" "$source_region" "$region" "${linux_arm64_source_image_name}-${region}")")
 
-    echo "--- Copying :windows: $windows_amd64_source_image_id to $region" >&2
+    echo "--- :windows: Copying Windows AMD64 $windows_amd64_source_image_id to $region" >&2
     IMAGES+=("$(copy_ami_to_region "$windows_amd64_source_image_id" "$source_region" "$region" "${windows_amd64_source_image_name}-${region}")")
   else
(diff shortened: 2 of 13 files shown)


.buildkite/pipeline.yml
```diff
@@ -68,10 +68,38 @@ steps:
...
+    timeout_in_minutes: 60
+    retry: { automatic: { limit: 3 } }
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
+    depends_on:
+      - "lint"
+      - "bats-tests"
+
+  - id: "linux-arm64-launch"
+    name: ":cloudformation: :linux: :arm: Launch"
```
yob, line 87: oh, turns out we don't support the `:arm:` emoji.

templates/aws-stack.yml
```diff
@@ -485,6 +485,17 @@ Conditions:
...
+    UsingArmInstances:
+      !Or
+        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "m6g" ]
+        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "m6gd" ]
+        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "t4g" ]
+        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "a1" ]
+        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "c6g" ]
+        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "c6gd" ]
+        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "r6g" ]
+        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "r6gd" ]
```
yob, line 497: :vomiting_face:

templates/aws-stack.yml
```diff
@@ -810,7 +821,7 @@ Resources:
               SpotOptions:
                 MaxPrice: !Ref SpotPrice
             - !Ref "AWS::NoValue"
-          ImageId: !If [ HasImageId, !Ref ImageId, !If [ HasImageIdParameter, !GetAtt ImageIdParameterStack.Outputs.ImageId, !FindInMap [ AWSRegion2AMI, !Ref 'AWS::Region', !Ref InstanceOperatingSystem ] ] ]
+          ImageId: !If [ HasImageId, !Ref ImageId, !If [ HasImageIdParameter, !GetAtt ImageIdParameterStack.Outputs.ImageId, !If [ UseWindowsAgents, !FindInMap [ AWSRegion2AMI, !Ref 'AWS::Region', 'windows'], !If [ UsingArmInstances, !FindInMap [ AWSRegion2AMI, !Ref 'AWS::Region', 'linuxarm'], !FindInMap [ AWSRegion2AMI, !Ref 'AWS::Region', 'linux'] ] ] ] ]
```
yob, line 824: :vomiting_face:

chloeruka: maybe at this point we should multi-line this so it's easier to reason about?

chloeruka:
This is looking great. The seamless support for ARM InstanceTypes without a parameter is a huge win in my books. Will this work for multi-instance stacks? (#635)

.buildkite/pipeline.yml
chloeruka: hm! there is one that might work for an emoji. i can make a pull to buildkite/emojis

image

yob: Maybe we're  restricted in using it? I'm not sure what our policies on corporate emjois are

https://www.arm.com/company/policies/trademarks/guidelines-corporate-logo

> If you do not have permission to use the Arm logo please refer to the Arm Logo Permission Requests page.
>
> You do not need to seek permission if:
>
>    You already have written permission from Arm.
>    You are a journalist, media outlet or blog using the logo for editorial purposes.
>    You are a charity or other not-for-profit organisation that is being supported by Arm.

templates/aws-stack.yml
yob: Yer, good call. I'll make that change

packer/linux/scripts/install-buildkite-agent.sh
```diff
@@ -12,14 +20,17 @@ sudo useradd --base-dir /var/lib --uid 2000 buildkite-agent
...
 echo "Downloading buildkite-agent v${AGENT_VERSION} stable..."
-sudo curl -Lsf -o /usr/bin/buildkite-agent-stable \
-  "https://download.buildkite.com/agent/stable/${AGENT_VERSION}/buildkite-agent-linux-amd64"
-sudo chmod +x /usr/bin/buildkite-agent-stable
+mkdir -p buildkite-agent
+curl -Lsf -o buildkite-agent-linux.tar.gz \
+  "https://github.com/buildkite/agent/releases/download/v${AGENT_VERSION}/buildkite-agent-linux-${ARCH}-${AGENT_VERSION}.tar.gz"
+tar -C buildkite-agent -xvzf buildkite-agent-linux.tar.gz
+sudo cp buildkite-agent/buildkite-agent /usr/bin/buildkite-agent-stable
+rm -fr buildkite-agent buildkite-agent-linux.tar.gz
```
yob, line 28: Hmm, I just noticed this changes from downloading the binary to downloading and unpacking a tarball. @galak, was that because this URL didn't work at the time? https://download.buildkite.com/agent/stable/3.25.0/buildkite-agent-linux-arm64

yob:
> Will this work for multi-instance stacks

It chooses the arm64/amd64 AMI for all instances based on the first instance type, so it will work with multi-instance ASG - provided all the instance types use a consistent architecture.

.buildkite/pipeline.yml
```diff
@@ -68,10 +68,38 @@ steps:
...
+    depends_on: "packer-linux-arm64"
+
+  - id: "linux-arm64-test"
+    name: ":cloudformation: :linux: 💪 Test"
+    command: "goss validate --format documentation"
+    timeout_in_minutes: 5
+    agents:
+      stack: "buildkite-aws-stack-test-linux-arm64-${BUILDKITE_BUILD_NUMBER}"
+      queue: "testqueue-linux-arm64-${BUILDKITE_BUILD_NUMBER}"
+    depends_on: "linux-arm64-launch"
```
yob, line 101: the build DAG now looks like this:

Screenshot from 2020-10-30 11-42-41

.buildkite/steps/copy.sh
```diff
@@ -110,15 +112,16 @@ if [[ $BUILDKITE_BRANCH != "master" ]] && [[ "$BUILDKITE_TAG" != "$BUILDKITE_BRA
   cat << EOF > "$mapping_file"
 Mappings:
   AWSRegion2AMI:
-    ${AWS_REGION} : { linux: $linux_source_image_id, windows: $windows_source_image_id }
+    ${AWS_REGION} : { linux: $linux_amd64_source_image_id, linuxarm: $linux_arm64_source_image_id, windows: $windows_amd64_source_image_id }
 EOF
   exit 0
 fi
```
yob, line 118: From this point on, the copy.sh script doesn't run on non-master branches. I haven't tested that copying all AMIs from us-east-1 to other region works.

.buildkite/pipeline.yml
pda: I imagine it's fair use of some sort, and it'll be fine. We probably haven't added it because nobody has suggested/done it. _I am not a lawyer etc etc._

Hmm.

> Do not use the logo or logotype as an element in titles, headlines or text.
> Do not place competing visual elements (including but not limited to other trademarks) close to the logo.

It'd be hard to stick to those. 🤷‍♂️

I've submitted a request to Arm via their Arm Logo Permission Requests page…

> Buildkite is a Continuous Integration platform. Our customers increasingly build software for Arm-based processors. We would like permission to add the Arm corporate logo to our emoji repository[1]. This emoji would allow software engineers using our platform to succinctly indicate that a particular software build target is for Arm processors.
> [1]: https://github.com/buildkite/emojis

nitrocode:
> * On linux/amd64 we install docker-compose using a binary download. Sadly there's no official arm64 binary published (see docker/compose#7472) so I've fallen back on installing via pip (for arm64 only). This works, but it's gross to install it differently depending on arch. Hopefully we can go back to binary installs if upstream starts publishing an arm64 version

I'm ok with this. 

> * I'm not super happy with the changes to the CI pipeline scripts yet. They work, but they feel messy.

Done is better than perfect, no? 😄 

> * Explicitly listing the ARM instance types feels fragile - we'll have to update it when future instance types are announced. I'm not sure how else we could do it though

It's too bad cloudformation doesn't support a string function that allows you to get the last character of a string (You could create a string macro but that might be overkill). If so, you could retrieve the last character and check if it's a `g` and assume it's a graviton/arm instance.

Sigh. Another reason for terraform: https://github.com/buildkite/elastic-ci-stack-for-aws/issues/246#issuecomment-694592688

We could use a map with a `FindInMap` function and just bump minor versions of the stack upon each update. That would be fairly simple for now.

yob:
I've rebased onto master to resolve a conflict with #772 and force pushed.

pda:
> On linux/amd64 we install docker-compose using a binary download. Sadly there's no official arm64 binary published (see docker/compose#7472) so I've fallen back on installing via pip (for arm64 only). This works, but it's gross to install it differently depending on arch. Hopefully we can go back to binary installs if upstream starts publishing an arm64 version

Seems like maybe `docker-compose` is broken?

> RuntimeError: You are linking against OpenSSL 1.0.2, which is no longer supported by the OpenSSL project. To use this version of cryptography you need to upgrade to a newer version of OpenSSL. For this version only you can also set the environment variable CRYPTOGRAPHY_ALLOW_OPENSSL_102 to allow OpenSSL 1.0.2.

```
[ec2-user@ip-10-0-1-215 ~]$ docker-compose --version
Traceback (most recent call last):
  File "/usr/local/bin/docker-compose", line 7, in <module>
    from compose.cli.main import main
  File "/usr/local/lib/python3.7/site-packages/compose/cli/main.py", line 14, in <module>
    import docker.errors
  File "/usr/local/lib/python3.7/site-packages/docker/__init__.py", line 2, in <module>
    from .api import APIClient
  File "/usr/local/lib/python3.7/site-packages/docker/api/__init__.py", line 2, in <module>
    from .client import APIClient
  File "/usr/local/lib/python3.7/site-packages/docker/api/client.py", line 10, in <module>
    from .. import auth
  File "/usr/local/lib/python3.7/site-packages/docker/auth.py", line 9, in <module>
    from .utils import config
  File "/usr/local/lib/python3.7/site-packages/docker/utils/__init__.py", line 3, in <module>
    from .decorators import check_resource, minimum_version, update_headers
  File "/usr/local/lib/python3.7/site-packages/docker/utils/decorators.py", line 4, in <module>
    from . import utils
  File "/usr/local/lib/python3.7/site-packages/docker/utils/utils.py", line 13, in <module>
    from .. import tls
  File "/usr/local/lib/python3.7/site-packages/docker/tls.py", line 5, in <module>
    from .transport import SSLHTTPAdapter
  File "/usr/local/lib/python3.7/site-packages/docker/transport/__init__.py", line 11, in <module>
    from .sshconn import SSHHTTPAdapter
  File "/usr/local/lib/python3.7/site-packages/docker/transport/sshconn.py", line 1, in <module>
    import paramiko
  File "/usr/local/lib/python3.7/site-packages/paramiko/__init__.py", line 22, in <module>
    from paramiko.transport import SecurityOptions, Transport
  File "/usr/local/lib/python3.7/site-packages/paramiko/transport.py", line 129, in <module>
    class Transport(threading.Thread, ClosingContextManager):
  File "/usr/local/lib/python3.7/site-packages/paramiko/transport.py", line 190, in Transport
    if KexCurve25519.is_available():
  File "/usr/local/lib/python3.7/site-packages/paramiko/kex_curve25519.py", line 30, in is_available
    X25519PrivateKey.generate()
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/primitives/asymmetric/x25519.py", line 39, in generate
    from cryptography.hazmat.backends.openssl.backend import backend
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/backends/openssl/__init__.py", line 7, in <module>
    from cryptography.hazmat.backends.openssl.backend import backend
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/backends/openssl/backend.py", line 117, in <module>
    from cryptography.hazmat.bindings.openssl import binding
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 222, in <module>
    _verify_openssl_version(Binding.lib)
  File "/usr/local/lib64/python3.7/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 183, in _verify_openssl_version
    "You are linking against OpenSSL 1.0.2, which is no longer "
RuntimeError: You are linking against OpenSSL 1.0.2, which is no longer supported by the OpenSSL project. To use this version of cryptography you need to upgrade to a newer version of OpenSSL. For this version only you can also set the environment variable CRYPTOGRAPHY_ALLOW_OPENSSL_102 to allow OpenSSL 1.0.2.
```

yob:
Ergh. It's such a shame we can't just download a binary version of docker-compose like we do on amd64.

This is an incompatibility between a python package and the ye olde version of openssl that ships on amazon linux 2.

I'm surprised to see the exception though - I thought this new line in the environment hook would squash it..

https://github.com/buildkite/elastic-ci-stack-for-aws/pull/758/files#diff-2792228b7bb1e85bf0415ed131115c08df8eb461cbb7a68f55bf3c7b3ccbd44bR15

pda:
> Edit: Oh, I see. your test was via an ssh session.

Ahhh yeah I was vaguely wondering if that would make a difference, but thought probably not.
Seems like it would be nice to fix anyway, but I guess not a huge deal.

yob:
We could fix it by adding that ENV bar to user sessions some way, but it
feels increasingly fragile.

At some point a new AMI build will pull in a versions of the python package
that drops support altogether 😬

pda:
I did a bit of digging…
Looks like Amazon Linux 2 ships old `openssl` 1.0.2 but also [newer `openssl11` v1.1.1](https://forums.aws.amazon.com/ann.jspa?annID=7562).
Looks like docker-compose depends on paramiko which depends on cryptography which depends on `libssl.so.10` which implicitly pins it to v1.0.x, so the newer `openssl11` package doesn't help; it provides `libssl.so.1.1`
I also couldn't see any easy way to get this python stuff to use gnutls or any other alternative library.

pda:
That said, https://cryptography.io/en/3.2/installation/ says it does support OpenSSL 1.1 so I don't know why it's so determined to load `libssl.so.10`, even when I delete that symlink (which, incidentally, breaks yum and sudo, so this test machine is dead).

yob:
I've pushed a change that installs the openssl1 package (and headers) before install docker-compose, and now I'm able to ssh in and print the version:

```
$ docker-compose version
docker-compose version 1.27.4, build unknown
docker-py version: 4.4.0
CPython version: 3.7.9
OpenSSL version: OpenSSL 1.0.2k-fips  26 Jan 2017
```

... weird that it still  reports openssl 1.0.2 at runtime though 🤔

Here's the openssl bits that are installed when I ssh in:

```
$ yum list installed | grep openssl
openssl.aarch64                       1:1.0.2k-19.amzn2.0.3          installed  
openssl-libs.aarch64                  1:1.0.2k-19.amzn2.0.3          installed  
openssl11.aarch64                     1:1.1.1c-15.amzn2.0.1          @amzn2-core
openssl11-devel.aarch64               1:1.1.1c-15.amzn2.0.1          @amzn2-core
openssl11-libs.aarch64                1:1.1.1c-15.amzn2.0.1          @amzn2-core
```

pda:
As far as I know, I don't suffer from dyslexia, but I've been mixing up amd64/arm64 terribly.

I believe the official name of the ARM 64-bit architecture is [AArch64](https://en.wikipedia.org/wiki/AArch64), while ARM64 is [unofficial/accidental](https://www.phoronix.com/scan.php?page=news_item&px=MTY5ODk).

The industry remains a bit split (e.g. Go calls it arm64), but given AArch64 is the official name, and Amazon Linux 2 calls it aarch64, perhaps we should use that name too?

Proposal: `s/arm64/aarch64/ ` anywhere that we control. Thoughts?

yob:
I get arm64/amd64 a bit confused when reading, but I find the symmetry
useful when writing.

We read the code more than we wrote it though and I assume I'll grow more
familiar with how many A's to use in aaaaarch64, so no objections from me 👍

pda:
I think I talked myself out of changing things to AArch64.
But I did push a commit that clearly labels the build steps as AMD64 or ARM64.

.buildkite/pipeline.yml
```diff
@@ -35,6 +35,25 @@ steps:
...
+          - "./build:/build:rw"
+          - "./plugins/secrets/s3secrets-helper:/s3secrets-helper:ro"
+        workdir: /s3secrets-helper
+        environment:
+          - "GOOS=linux"
+          - "GOARCH=arm64"
+        command: ["go", "build", "-o", "/build/s3secrets-helper-linux-arm64"]
+    artifact_paths:
+      - build/s3secrets-helper-linux-arm64
+
```
pda, line 56: Side-note:
It's a shame this is so many lines of code, copied for each os/arch.
But, it is nicely self-contained, not needing any scripts etc.

.buildkite/steps/copy.sh
```diff
@@ -128,19 +131,23 @@ if aws s3 cp "${s3_mappings_cache}" "$mapping_file" ; then
...
 for region in ${ALL_REGIONS[*]}; do
   if [[ $region != "$source_region" ]] ; then
-    echo "--- Copying :linux: $linux_source_image_id to $region" >&2
-    IMAGES+=("$(copy_ami_to_region "$linux_source_image_id" "$source_region" "$region" "${linux_source_image_name}-${region}")")
+    echo "--- Copying :linux: $linux_amd64_source_image_id to $region" >&2
+    IMAGES+=("$(copy_ami_to_region "$linux_amd64_source_image_id" "$source_region" "$region" "${linux_amd64_source_image_name}-${region}")")
 
-    echo "--- Copying :windows: $windows_source_image_id to $region" >&2
-    IMAGES+=("$(copy_ami_to_region "$windows_source_image_id" "$source_region" "$region" "${windows_source_image_name}-${region}")")
+    echo "--- Copying :linux: :arm: $linux_arm64_source_image_id to $region" >&2
```
pda: I pushed 9de77ec46e337a11687097841ed3eba4bb6cfcb3 to further avoid using 💪🏼 for ARM64.

Makefile
```diff
@@ -10,6 +10,9 @@ PACKER_WINDOWS_FILES = $(exec find packer/windows)
 AWS_REGION ?= us-east-1
 AMZN_LINUX2_AMI ?= $(shell aws ec2 describe-images --region $(AWS_REGION) --owners amazon --filters 'Name=name,Values=amzn2-ami-hvm-2.0.????????-x86_64-gp2' 'Name=state,Values=available' --output json | jq -r '.Images | sort_by(.CreationDate) | last(.[]).ImageId')
 
+ARM64_INSTANCE_TYPE = m6g.xlarge
+X86_INSTANCE_TYPE = c5.xlarge
```
pda, line 14: Hmm. We use AMD64 rather than X86_64 everywhere else.
However this file deals with upstream Amazon Linux 2 AMIs, which use x86_64.
I wonder if this variable should be:
* `X86_INSTANCE_TYPE` (current, introduced in this PR)
* `X86_64_INSTANCE_TYPE` (Because X86 != X86_64, and aligned with AWS / Amazon Linux 2)
*`AMD64_INSTANCE_TYPE` (aligned with all the references that we control)

Very minor detail, it probably doesn't matter.

packer/linux/buildkite-ami.json
```diff
@@ -1,23 +1,27 @@
 {
   "variables": {
-    "region": "us-east-1"
+    "region": "us-east-1",
+    "arch": "x86_64",
+    "goarch": "amd64",
```
pda, line 5: AMD64 vs X86_64 and AArch64 vs ARM64 will grate on me until the end of time 😅

packer/linux/scripts/install-buildkite-agent.sh
```diff
@@ -3,6 +3,14 @@ set -eu -o pipefail
 
 AGENT_VERSION=3.25.0
 
+MACHINE="$(uname -m)"
+
+case "${MACHINE}" in
+	x86_64)    ARCH=amd64;;
+	aarch64)   ARCH=arm64;;
```
pda, line 10: heh

packer/linux/scripts/install-docker.sh
```diff
@@ -30,18 +31,22 @@ sudo curl -Lfs -o /etc/systemd/system/docker.socket https://raw.githubuserconten
...
+if [ "${MACHINE}" == "x86_64" ]; then
+	echo "Downloading docker-compose..."
+	sudo curl -Lsf -o /usr/bin/docker-compose https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-Linux-x86_64
+	sudo chmod +x /usr/bin/docker-compose
+	docker-compose --version
+elif [[ "${MACHINE}" == "aarch64" ]]; then
+  sudo yum install -y gcc-c++ libffi-devel openssl11 openssl11-devel python3-devel
+  sudo pip3 install docker-compose
+	docker-compose version
+fi
```
pda, line 43: I wonder if we want an `else` warning/error here.
Probably not important right now.

.buildkite/steps/copy.sh
```diff
@@ -110,15 +112,16 @@ if [[ $BUILDKITE_BRANCH != "master" ]] && [[ "$BUILDKITE_TAG" != "$BUILDKITE_BRA
   cat << EOF > "$mapping_file"
 Mappings:
   AWSRegion2AMI:
-    ${AWS_REGION} : { linux: $linux_source_image_id, windows: $windows_source_image_id }
+    ${AWS_REGION} : { linux: $linux_amd64_source_image_id, linuxarm: $linux_arm64_source_image_id, windows: $windows_amd64_source_image_id }
```
pda, line 115: `linuxarm`… should we use something like `linux-arm64` or `linux_arm64`?
Or is this referencing something outside our control?

pda (approved):
This is bloody fantastic.
Apologies for taking a long time to review.
In the end, all I have to offer are some niggles about naming, none of which are really blockers.

pda:
Successful build:
image

Agent looks good:
image

keithpitt:
Woo!

.buildkite/steps/copy.sh
yob: These map keys are totally arbitrary and can be whatever we want. I'll update them.

packer/linux/scripts/install-docker.sh
yob: good idea

yob:
:boom:

etaoins:
I just moved a pretty complex Buildkite pipeline (using `docker`, docker-compose`, `docker-login` & `seek-jobs/docker-ecr-cache` plugins along with `aws-cli` on the agent) to the elastic ARM instances:
<https://github.com/etaoins/arret/blob/master/.buildkite/pipeline.yml>

There are still some steps on AMD64 due to actually needing to test x86-64 and lack of support in `vscode-test`.

Moving a build step from `t3.medium` to `t4g.medium` yielded a ~10-15% speedup while the instances are ~20% cheaper. There were no hiccups with the agents at all; everything worked as it did on AMD64.

Awesome work everyone! 🎉

yob:
@etaoins thanks for sharing your experience with the new ARM images, it sounds encouraging!

PR #762, opened 2020-11-03
Retreive agent token from param store on windows agents

### LINKED ISSUE #761: Windows agent doesn't retreive agent token from Parameter Path if specified
When running `bk-install-elastic-stack.ps1` on windows agent bootstrap the script does not check if BUILDKITE_AGENT_TOKEN is set or if BUILDKITE_AGENT_TOKEN_PATH is set and just blindly sets:
``` token="${Env:BUILDKITE_AGENT_TOKEN}" ```

I will have a PR for this shortly - just logging this as an issue incase anyone else is struggling with windows agents in this scenario.

chrisfowles:
fixes #761 

Gets agent token from param store if `BuildkiteAgentTokenParameterStorePath` is set and `BuildkiteAgentToken` is not for windows instances.

### DIFF
--- a/packer/windows/conf/bin/bk-install-elastic-stack.ps1
+++ b/packer/windows/conf/bin/bk-install-elastic-stack.ps1
@@ -77,9 +77,8 @@ If ($Env:BUILDKITE_AGENT_ENABLE_GIT_MIRRORS_EXPERIMENT -eq "true") {
   $Env:BUILDKITE_AGENT_GIT_MIRRORS_PATH = "C:\buildkite-agent\git-mirrors"
 }
 
-# Get token from ssm param (if we have a path) if it's not set explicitly
-If (($null -eq $Env:BUILDKITE_AGENT_TOKEN -or $Env:BUILDKITE_AGENT_TOKEN -eq "") -and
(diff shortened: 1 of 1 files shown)


yob:
Thanks @chrisfowles :heart_eyes:  Missing this on the windows side was definitely an oversight, and confirmed on a test stack based on this PR   that the fix works.

My only thought is that current behaviour on linux is: use the SSM path if it's provided, overriding any manually provided token https://github.com/buildkite/elastic-ci-stack-for-aws/blob/857e3f72aa0e002d9dcd113b4871f93327bc822a/packer/linux/conf/bin/bk-install-elastic-stack.sh#L107-L109

The difference is small, but maybe it's worth using the same behaviour as linux to keep the two code paths as similar as we can?

chrisfowles:
Cheers! I've updated the logic to match the linux install @yob. 

There's possibly some discussion around if this is the "correct" logic, as I'd typically expect an explicitly defined value to override a retreived value for application configuration. However it's probably not worth a breaking change on the linux side for just this qualm ;)

yob:
lovely, thanks again.

I agree - I'd probably lean towards your initial behavior if we were starting from scratch, but I think the downsides small enough that consistency with the (only recently) released behaviour is worth it.

I might backport this to a new `5-0-stable` branch as well, and if no other significant bugs show up over the next few days then release a 5.0.1 stack.

yob:
We've released this in 5.0.1. Thanks again!

Issue #676, opened 2020-03-11
Consider adding Amazon ECR Docker Credential Helper


pda:
We could perhaps use this instead of the ECR plugin:
https://github.com/awslabs/amazon-ecr-credential-helper

I think it would be a more elegant solution, but there might be other considerations…?

toolmantim:
I don’t see why not

bshelton229:
I got a PR started (#763). I'm not sure if it's the correct approach, but it's a start. We'd like to explore this as maintaining a cache between steps would help us not run into rate limits occasionally as well as speed things up. It also means we don't need to know all the ECR registry IDS ahead of time or put them explicitly in a separate plugin definition in our steps.

jamestelfer:
At the moment if the `ecr-login` helper is configured at the `credsStore` or `credHelper` level, an explicit `docker login` causes it to fail.

We could work around this:

1. Require steps to switch from explicit login to using the helper (via an explicit plugin reference)
2. Work around the issue via a helper script as suggested in the [`ecr-credential-helper` issue](https://github.com/awslabs/amazon-ecr-credential-helper/issues/154#issuecomment-472988526)
3. Allow the helper to behave differently and ignore these hooks.

We're working on (3) by getting this behaviour changed in the ecr-helper itself (https://github.com/awslabs/amazon-ecr-credential-helper/pull/847)

This should allow for the elastic stack to install the helper in the base AMI, and optionally allow it to be configured in the Docker `config.json`.

I recommend configuring it for an allow list of ECR repos, rather than as the default. Configuring `ecr-login` using `credStore` isn't the optimal way to work with this plugin, as it would compromise other repo usage.

It would be better to install it for an allow list of repositories, like:

```json
  "credHelpers": {
    "<account_id_here>.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
    "public.ecr.aws": "ecr-login"
  },
```

This could perhaps be configured in the Elastic stack parameters, or become part of the default `environment` hook.

Ref: https://github.com/awslabs/amazon-ecr-credential-helper/pull/847

jamestelfer:
AWS have just [released v0.9.0](https://github.com/awslabs/amazon-ecr-credential-helper/releases/tag/v0.9.0) with the referenced issue fixed. 

[...]

PR #763, opened 2020-11-03
Optionally enable the ecr credential helper plugin

bshelton229:
This allows optionally enabling https://github.com/awslabs/amazon-ecr-credential-helper as mentioned in #676 . This plugin's cache should write to the appropriate home folder and should be shared between builds as it's not in the ephemeral `DOCKER_HOME`. That's my desired behavior, but I'm not 100% sure that's what everyone would want. I think this has the benefit over the ECR plugin of both maintaining a cache between builds and also working against any ECR repo you have access to without having to pre-populate the appropriate registry IDs. I should note I've had a little bit of trouble with this plugin's interactions with docker-compose, but that was years ago. I believe the docker project has resolved those issues with how it calls credential helpers at this point. This worked with the docker and docker-compose plugins for me in testing.

Overview:

- Installs the credential helper no matter the plugin setting. The binary won't do anything when it's not used and it's only 8mb and some change.
- Writes an empty `{}` `config.yaml` in the temp `DOCKER_HOME` so it can be modified later by various things if necessary.
- Follows the standard 

Known issues:

I have almost no Windows knowledge. I'm hoping to get some feedback on this general pattern before attempting to get this working in Windows at well.

I'm not sure if this is better off being made into a more general plugin and then incorporated into this stack via git submodules or not? I opted to bake it right in.

This configuration also precludes any other docker auth, like dockerhub. To make that work it would need to configure the `credHelpers` object to only work on certain configured ECR URLs. That would probably be best done in a separate plugin that was incorporated and configured via a git submodule? We currently only use ECR as an authenticated repo, so this meets that need.

### DIFF
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -48,6 +48,7 @@ PLUGINS_ENABLED=()
 [[ $SECRETS_PLUGIN_ENABLED == "true" ]] && PLUGINS_ENABLED+=("secrets")
 [[ $ECR_PLUGIN_ENABLED == "true" ]] && PLUGINS_ENABLED+=("ecr")
 [[ $DOCKER_LOGIN_PLUGIN_ENABLED == "true" ]] && PLUGINS_ENABLED+=("docker-login")
+[[ ${ECR_CRED_HELPER_PLUGIN_ENABLED:-} == "true" ]] && PLUGINS_ENABLED+=("ecr-cred-helper")
 
 # cfn-env is sourced by the environment hook in builds
 cat << EOF > /var/lib/buildkite-agent/cfn-env
--- a/packer/linux/conf/buildkite-agent/hooks/environment
+++ b/packer/linux/conf/buildkite-agent/hooks/environment
@@ -8,6 +8,9 @@ source ~/cfn-env
 BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY=$(mktemp -d)
 export BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY
 export DOCKER_CONFIG="$BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY"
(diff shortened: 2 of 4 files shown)


yob:
disclaimer: I'm not super familiar with docker credential stores or helpers, so some of this comment might be naive.

This is great, thanks for starting the discussion. Automatically using the instance credentials when accessing ECR seems like a great usability win.

> Installs the credential helper no matter the plugin setting

:+1: 

> I think this has the benefit over the ECR plugin of both maintaining a cache between builds

Could you describe the caching in more detail? What will be cached across builds, the credentials?

> I'm not sure if this is better off being made into a more general plugin and then incorporated into this stack via git submodules or not? I opted to bake it right in.
> This configuration also precludes any other docker auth, like dockerhub. To make that work it would need to configure the credHelpers object to only work on certain configured ECR URLs. That would probably be best done in a separate plugin that was incorporated and configured via a git submodule? We currently only use ECR as an authenticated repo, so this meets that need.

Given the new rate limits being imposed on unauthenticated docker hub pulls as of yesterday, I think it'd be a shame to make it impossible to authenticate to docker hub.

I guess that means `credHelpers` are our best bet, but we'd need a way to specify the specific registries that will use `ecr-login` :thinking: 

I'd want to run this idea by some colleagues before we commit to it, but what if we added a new ENV var to the existing ecr-login plugin that looks something like this: `BUILDKITE_PLUGIN_ECR_CREDHELPER_REGISTRIES="aws_account_id1.dkr.ecr.region.amazonaws.com,aws_account_id2.dkr.ecr.region.amazonaws.com"`?

That'd be useful for folks who use the ecr plugin directly. For folks using it via the elastic stack though, it'd mean setting the new ENV in an environment hook. Similar to the [existing elastic stack docs](https://github.com/buildkite/elastic-ci-stack-for-aws#docker-registry-support). It's a shame wild card credHelpers aren't possible (like `*.amazonaws.com`).

Would that work for your team? I know it doesn't align with this comment :(

> I think this has the benefit over the ECR plugin of both maintaining a cache between builds and also working against any ECR repo you have access to without having to pre-populate the appropriate registry IDs

PR #766, opened 2020-11-10
Bump JQ_VERSION to 1.6

nitrocode:


yob:
My only reservation here is we're trying to land first-class arm support (in #758), and currently that includes switching to installing jq 1.5 via yum:

https://github.com/buildkite/elastic-ci-stack-for-aws/pull/758/files#diff-40214ac00fcda1e5241174fd97f51ea3697f17ebebe62f2065913cf79f2dfe78R51

Is there a feature in 1.6 that you're keen on?

nitrocode:
Nah i just thought it would be good to update.

PR #767, opened 2020-11-11
Support IAM Permissions Boundaries

nitrocode:
This makes it a lot easier to give users access to create their own queue by allowing a permission boundary to be attached.

We currently do not allow iam role creation or attachment of policies without a permission boundary set.

For more context, see this [aws blog post on "Delegate permission management to developers by using IAM permissions boundaries"](https://aws.amazon.com/blogs/security/delegate-permission-management-to-developers-using-iam-permissions-boundaries/) and [AWS 2019 Reinvent "AWS Identity: Permission boundaries & delegation"](https://d1.awsstatic.com/events/reinvent/2019/REPEAT_1_AWS_identity_Permission_boundaries_&_delegation_SEC402-R1.pdf).

@yob @lox

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -299,6 +299,11 @@ Parameters:
     Description: Optional - A name for the IAM Role attached to the Instance Profile
     Default: ""
 
+  InstanceRolePermissionsBoundaryARN:
+    Type: String
+    Description: The ARN of the policy used to set the permissions boundary for the role.
+    Default: ""
+
   InstanceOperatingSystem:
     Type: String
     Description: The operating system to run on the instances
@@ -435,6 +440,9 @@ Conditions:
 
     SetInstanceRoleName:
       !Not [ !Equals [ !Ref InstanceRoleName, "" ] ]
+
+    SetInstanceRolePermissionsBoundaryARN:
+      !Not [ !Equals [ !Ref InstanceRolePermissionsBoundaryARN, "" ] ]
 
     UseSpecifiedSecretsBucket:
       !Not [ !Equals [ !Ref SecretsBucket, "" ] ]
@@ -635,6 +643,7 @@ Resources:
     Type: AWS::IAM::Role
     Properties:
       RoleName: !If [ SetInstanceRoleName, !Ref InstanceRoleName, !Sub "${AWS::StackName}-Role" ]
+      PermissionsBoundary: !If [ SetInstanceRolePermissionsBoundaryARN, !Ref InstanceRolePermissionsBoundaryARN ]
       ManagedPolicyArns: !If
           - HasManagedPolicies
           # Support multiple policies to attach by merging the values together and splitting on ','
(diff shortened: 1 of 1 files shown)


s672 (approved):

rb-nr:
@yob @pda @chloeruka thoughts on this ?

sj26:
This sounds like a good idea, but I'm not sure I understand permission boundaries well enough to fold it in.

@chloeruka and @pda, could you take a look?

nitrocode:
It's a pretty cool feature where you can define the maximum permissions an iam role could potentially use without exceeding it.

Here's an image explaining it https://d2908q01vomqb2.cloudfront.net/22d200f8670dbdb3e253a90eee5098477c95c23d/2018/07/03/delegated-admin-02.png and the aws blog I linked in the OP is good for background information on its benefits.

sj26:
I guess I don't understand why you wouldn't consolidate it to a single IAM policy — why have two policies which overlay?

nitrocode:
Heres a practical use case. We have devs that have access to create iam roles but only if they attach an iam permission boundary. This is a good way to setup guard rails for our users and still give them access to iam.

By allowing the attachment of a permission boundary, we can give them access to create iam roles without needing elevated permissions.

sj26:
Ah yes, I hadn't cottoned on to the idea of using it to restrict definer privileges — very useful!

chloeruka (approved):
Approving this with the caveat that I'm not an expert on permission boundaries and AWS seems to pitch them as being a last-line power-user feature if other methods aren't sufficient, but having brushed past them before this change makes a lot of sense, particularly for large-shaped organisations deploying this stack in broadly shared AWS accounts.

> We currently do not allow iam role creation or attachment of policies without a permission boundary set.

Is this why the Boundary is also attached to `AutoscalingLambdaExecutionRole` and `AsgProcessSuspenderRole`? The latter is limited only to the stack ASG resource, but the former does set `autoscaling:SetDesiredCapacity` to '*' which might be something we want to revisit to look more like `AsgProcessSuspenderRole`'s resource restriction.

nitrocode:
Hi chloeruka. Thank you for approving.

I added the boundary to every IAM role so we could use our current boundary which should cover all of the permissions required for each role.

Yes that permission may need to be revisited but luckily with the perm boundary, we can restrict that using the boundary instead of the perms directly on the iam role.

pda (approved):

chloeruka:
We've had an internal discussion and decided to merge this, thanks for your contribution and patience @nitrocode!

yob:
Oops, this failing the build with this error:

```
test -f build/mappings.yml
awk '{ \
	if ($0 ~ /AWSRegion2AMI:/ && system("test -f build/mappings.yml") == 0) { \
		system("grep -v Mappings: build/mappings.yml") \
	} else { \
		print \
	}\
}' templates/aws-stack.yml | sed "s/%v/v5.2.0-7-g69ed7a5/" > build/aws-stack.yml
aws cloudformation validate-template \
	--output text \
	--template-body "file:///var/lib/buildkite-agent/builds/buildkite-i-0a312fd9d71661a3b-4/buildkite-aws-stack/buildkite-aws-stack/build/aws-stack.yml"

An error occurred (ValidationError) when calling the ValidateTemplate operation: Template error: Fn::If requires a list argument with three elements
```

Apologies for not picking this up earlier @nitrocode!

I'll revert this PR to get the build green again, and I'll keep an eye out for a revised PR that I can merge :heart:

nitrocode:
Yuck. Thanks for catching that. I put in another pr. Is there a way we can get the repos pr branch checks to test that prior to merge?

PR #773, opened 2020-11-24
plugins/ecr updated to v2.2.0: ensure log group header

pda:
> **Ensure a log output heading is always shown**
> Otherwise the log output (and timing information) of this plugin is grouped into whatever previous heading existed.
For elastic-stack it ends up included under the S3 secrets heading.
— https://github.com/buildkite-plugins/ecr-buildkite-plugin/pull/55

The highlighted portion of this output came from the ECR plugin, but is currently grouped into the “Downloading secrets …” header:
image


pda:
After this PR:
image

The “Fixing permissions” doesn't belong in this group, but that needs to be fixed in the next plugin/hook/bootstrap.

Honestly the main thing I was trying to achieve in this PR was making the s3secrets timing info correct 👌🏼

yob:
:+1:

PR #775, opened 2020-11-26
docs: ✏️ Add reference to BUILDKITE_PLUGIN_S3_SECRETS_BUCKET_PRE

maatthc:
The [AWS S3 Secrets Buildkite Plugin](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks) allows the location of the secret files on the S3 bucket to be defined using the environment variable `BUILDKITE_PLUGIN_S3_SECRETS_BUCKET_PREFIX` ([code](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/blob/67c4e227b5e591608794776d6330c7fe4ea2273a/s3secrets-helper/secrets/secrets.go#L35)). 

Therefore, instead of using the pipeline name, e.g:
`s3://{bucket_name}/{pipeline-slug}/private_ssh_key`

We can point it to any other location, e.g:
`s3://{bucket_name}/{BUILDKITE_PLUGIN_S3_SECRETS_BUCKET_PREFIX}/private_ssh_key`

This is a very handy feature when dealing with multiple pipelines in a monorepo, as it allow pipelines to share the same secrets (if that is desired).

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ Features:
 
 <!-- toc -->
 
-- [Contents](#contents)
 - [Getting Started](#getting-started)
 - [Build Secrets](#build-secrets)
 - [What’s On Each Machine?](#whats-on-each-machine)
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ Features:
 
 <!-- toc -->
 
+- [Contents](#contents)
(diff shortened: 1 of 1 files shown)


pda:
Thanks @maatthc 👍🏼

PR #779, opened 2020-12-07
Fix typo in the mappings file rendered during CI

yob:
In PR #758 we changed the AMI mappings keys to linxuamd64/linuxarm64/windows. However, we left an older format (linux_arm64/linux_amd64) in the version of the template rendered during CI.


JuanitoFatas (approved):
Looks good to me!

JuanitoFatas (approved):

yob (commented):
I've created a test stack with the CF template from this branch and confirmed it works as it should.

PR #780, opened 2020-12-07
Restore the ability to change InstanceType on a running stack

### LINKED ISSUE #774: Changing InstanceType on a running master stack fails
I haven't looked into this fully yet, but just recording something I noticed this morning before I forget.

When I change **just** the InstanceType on a running 5.0.1 stack, the changeset looks like this and the update rolls out just fine:

change-instancetype-5-0-1

However, when I attempt the same InstanceType change on a running stack from current master (a8c10c6) , the changeset looks like this and it fails with an opaque error (`Internal failure`):

Screenshot from 2020-11-25 12-09-03

Screenshot from 2020-11-25 12-47-57

I manually bisected on master, and the changeset difference seems to appear when #710 merges :thinking:

yob:
In PR #710 we added the option to select multiple instance types by providing the InstanceType stack param as a CSV.

Unfortunately, after merging we discovered that stack updates that changed only the InstanceType stack param would fail with a cryptic error:

> Internal Failure

Helpful!

After some experimentation, I realised my original PR was incorrectly skipping the instance type overrides for types that weren't provided by the user.

Turns out, it's possible to use YAML for 15 years and *still* get confused about lists 😳

Fixes #774

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -970,17 +970,17 @@ Resources:
             Version: !GetAtt "AgentLaunchTemplate.LatestVersionNumber"
           Overrides:
           - InstanceType: !Select [ "0", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ]
-          - InstanceType: !If
+          - !If
             - UseInstanceType2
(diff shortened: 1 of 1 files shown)


yob (commented):
I've created a number of test stacks, then upgraded to andfrom various instance types. Everything worked as expected, with no `Internal Error`s :tada:

PR #768, opened 2020-11-11
Add a boolean to override replacing the asg

nitrocode:
Should solve this https://github.com/buildkite/elastic-ci-stack-for-aws/issues/764

Please review when time permits

@yob @chloeruka @lox

PR #781, opened 2020-12-07
Adjust the way we set IAM role for the process suspender

### LINKED ISSUE #764: ASG removed and recreated when updating the launch template parameters
I'm a bit of a cloud formation noob but i have noticed that the iam role persists between stack updates. If i change values that have to do with the launch template, it destroys my launch template, my instances, and my asg. I noticed that the resources affected are usually the ones suffixed with the cf stack id. My iam role does not have this suffix.

In terraform, it would simply update the launch template version and assign the version to the existing asg. The instances could then be refreshed to get the latest changes from the template. If it works in terraform then this should be available in cf.

Edit: in the cloudformation stack

* The [`LaunchTemplateName`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-launchtemplate.html) is unset so it must be auto generating that
* The [`AutoScalingGroupName`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html) is also unset

If the above two parameters are hard coded like the iam role name, is there another parameter to set the asg or LT to use the latest version of the LT or is that the default in cloud formation? 

Providing these values will allow us to update our stacks without having to check if our asg has a desired count greater than zero. If we update while the count is greater than zero then current builds are interrupted.

yob:
The old way works, but any time the ASG is recreated it gets a new name and the IAM role also needs to be updated. It'd be nice to skip the IAM changes so some common stack updates (like changing only InstanceType) can be performed by users without permissions to make IAM changes. 

This converts the resource in the role to use a wildcard suffix, so the role has permission to suspend the process on any ASG called `{stack-name}-AgentAutoScaleGroup-*`.

With this change, updating a stack with only an InstanceType change requires no IAM changes:

Screenshot from 2020-12-08 08-48-24

Fixes #764 

Alternative to #768

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -1026,7 +1026,7 @@ Resources:
               - Effect: Allow
                 Action:
                   - 'autoscaling:SuspendProcesses'
-                Resource: !Sub arn:${AWS::Partition}:autoscaling:${AWS::Region}:${AWS::AccountId}:autoScalingGroup:*:autoScalingGroupName/${AgentAutoScaleGroup}
+                Resource: !Sub arn:${AWS::Partition}:autoscaling:${AWS::Region}:${AWS::AccountId}:autoScalingGroup:*:autoScalingGroupName/${AWS::StackName}-AgentAutoScaleGroup-*
 
   AzRebalancingSuspenderFunction:
     Type: AWS::Lambda::Function


yob:
I've tested this a number of times and I'm feeling pretty confident that it:

* avoids the need for IAM permissions for basic stack updates (like changing the instance type)
* doesn't break anything

PR #784, opened 2020-12-21
default to gp3 volumes

yob:
AWS [announced gp3 EBS volumes earlier this month](https://aws.amazon.com/about-aws/whats-new/2020/12/introducing-new-amazon-ebs-general-purpose-volumes-gp3/). They're slightly cheaper and have [the same or better performance than gp2](https://aws.amazon.com/ebs/general-purpose/), so we may as well default to the new type.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -272,7 +272,7 @@ Parameters:
   RootVolumeType:
     Description: Type of root volume to use
     Type: String
-    Default: "gp2"
+    Default: "gp3"
 
   SecurityGroupId:
     Type: String


pda:
When building elastic-stack AMIs, we search for the latest upstream `amzn2-ami-hvm-2.0.*-gp2` as a base. Note the `gp2` in the name`. But currently there's no `…-gp3` equivalent. Maybe one day there will be, though 🤷‍♂️ I guess we could use `amzn2-ami-hvm-2.0.*-gp?` or `amzn2-ami-hvm-2.0.*-gp*` or something. Anyway, fine for now, just a note for future.

PR #786, opened 2021-01-06
Default to IMDSv2 for instance metadata calls

### LINKED ISSUE #647: Enable IMSv2
We're planning to enforce the instance metadata endpoint v2 (IMSv2): https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#configuring-instance-metadata-options

It would be ideal if the elastic CI stack either had this on by default, or added an option for opting into IMSv2

holmesjr:
This resolves #647

All manual metadata curl calls in the build scripts are now preceded by a call to get a token. The YAML template now defaults to enforcing this, with an option to allow older v1 calls. 

The ability to turn off metadata calls entirely has not been included, as the docker GC script requires it.

### DIFF
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -12,8 +12,9 @@ on_error() {
 	local errorLine="$1"
 
 	if [[ $exitCode != 0 ]] ; then
+	  TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
 		aws autoscaling set-instance-health \
-			--instance-id "$(curl http://169.254.169.254/latest/meta-data/instance-id)" \
+			--instance-id "$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id)" \
 			--health-status Unhealthy || true
 	fi
 
--- a/packer/linux/conf/buildkite-agent/scripts/terminate-instance
+++ b/packer/linux/conf/buildkite-agent/scripts/terminate-instance
@@ -6,8 +6,9 @@ echo "sleeping for 10 seconds before terminating instance to allow agent logs to
 
 sleep 10
 
-instance_id=$(curl -fsSL http://169.254.169.254/latest/meta-data/instance-id)
-region=$(curl -fsSL http://169.254.169.254/latest/meta-data/placement/availability-zone | head -c -1)
+TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
+instance_id=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" -fsSL http://169.254.169.254/latest/meta-data/instance-id)
+region=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" -fsSL http://169.254.169.254/latest/meta-data/placement/availability-zone | head -c -1)
 
 echo "requesting instance termination..."
 
--- a/packer/linux/conf/docker/cron.hourly/docker-low-disk-gc
+++ b/packer/linux/conf/docker/cron.hourly/docker-low-disk-gc
@@ -13,8 +13,10 @@ mark_instance_unhealthy() {
 
   # mark the instance for termination
   echo "Marking instance as unhealthy"
+
+  TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
   aws autoscaling set-instance-health \
-    --instance-id "$(curl http://169.254.169.254/latest/meta-data/instance-id)" \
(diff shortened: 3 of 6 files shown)


holmesjr:
Note one call here that didn't change: https://github.com/buildkite/elastic-ci-stack-for-aws/pull/786/files#diff-e24d3687ef938e86d3a2b1f4129ef9f285f76319211f2181b253ce299e8c7621R31 

If the version of cloud-utils is recent enough, it should be using v2 calls. If not, it needs to be updated.

yob (commented):
Thanks @holmesjr!

We're still a bit short staffed over the summer holidays, but we'll try and give feedback by next week.

My initial thought is maybe we can default the new param to optional so this is a non-breaking change and can be merged into a 5.x release, and then we can default the param to `required` when a 6.x release eventually happens. Hold off on making that change until I get more eyes on this though :heart:

pda:
FYI, I don't know if this is useful, but when I was doing IMDSv2 from bash in a branch elsewhere, it looked like this:

```bash
# global var cache for IMDSv2 token
imds_token=""

# curl options used by imds functions
imds_curlopt=(--silent --show-error --fail --connect-timeout 1 --retry 2)

imds_get_token() {
  echo >&2 "ec2-metadata: acquiring IMDSv2 token"
  imds_token="$(curl "${imds_curlopt[@]}" \
    -H "X-aws-ec2-metadata-token-ttl-seconds: 300" \
    -X PUT "http://169.254.169.254/latest/api/token")"
  if [[ -z $imds_token ]]; then
    echo >&2 "ec2-metadata: failed to acquire IMDSv2 token"
    return 1
  fi
}

imds_get() {
  local path="/latest/meta-data/$1"
  echo >&2 "ec2-metadata: getting $path"
  curl "${imds_curlopt[@]}" -H "X-aws-ec2-metadata-token: $imds_token" "http://169.254.169.254$path"
}

usage_example() {
  if ! imds_get_token; then return 1; fi
  local role; role=$(imds_get iam/security-credentials | head -n1)
}
```

Note I never shipped that branch, maybe it's no good, but your PR just triggered a memory of it and I figured I'd share it just in case it's useful.

yob:
@pda does the new golang-based secrets plugin already use tokens for metadata requests? It'd be Not Fun if making tokens required broke fetching secrets from S3

pda:
I imagine so, it's using a recent AWS SDK. I haven't explicitly checked.

jamesholmes-linktree:
Yeah, last time I looked it was just using the SDK, so it should be fine.

holmesjr:
Yeah, last time I looked it was just using the SDK, so it should be fine.

holmesjr:
Sorry to be a pain - is there anything I can do to help get this reviewed?

pda:
Sorry for being slow over here @holmesjr :(

> My initial thought is maybe we can default the new param to optional so this is a non-breaking change and can be merged into a 5.x release, and then we can default the param to required when a 6.x release eventually happens.

I think I agree with @yob about making `optional` the default, so that existing stacks aren't forced to IMDSv2 without opting in. But I'm also wondering if we can reduce the parameter count; we're getting pretty close to the CloudFormation limit per template.

What do you think about either of these approaches?

A: Only add an `IMDSv2` boolean parameter, default `false`. The next major elastic-stack would remove the parameter and hard-code it to `true`. `MetadataOptions.HttpPutResponseHopLimit` would be hard-coded to `1` until if/when somebody needs multiple hops.

B: Only add a `IMDSHopLimit` optional parameter, default `""` (empty). Leaving it empty keeps IMDSv1 by passing `AWS::NoValue` to those `MetadataOptions`. Setting it to e.g. `1` forces IMDSv2.

Option A adds one parameter to elastic-stack v5 and removes it again in v6, but can't control hop limit.
Option B adds one parameter forever.

Any thoughts?

pda:
Also — would it help you if we shipped the code that adds `X-aws-ec2-metadata-token` headers immediately, with a separate PR for the CloudFormation parameters to make IMDSv2 mandatory? Or is it only useful to you once it's all merged?

holmesjr:
> Option A adds one parameter to elastic-stack v5 and removes it again in v6, but can't control hop limit.
> Option B adds one parameter forever.
> 
> Any thoughts?

Yeah, I understand the issues with params. While I have no need for multiple hops, I imagine someone somewhere might and we'd end up adding it in anyway. Maybe option B is better for that reason, but I'd actually prefer option A.

>Also — would it help you if we shipped the code that adds X-aws-ec2-metadata-token headers immediately, with a separate PR for the CloudFormation parameters to make IMDSv2 mandatory? Or is it only useful to you once it's all merged?

In terms of the security improvement, mandatory IMDSv2 is the goal. The other part is just what was necessary to enable it. So for my use case, it's not useful until the whole lot's merged in.

pda:
👆🏼 I've split this into two PRs; I think #788 is uncontroversial and can merge ASAP. I'll wait for a bit more feedback for #789, but I think it'll be fine to merge.

yob:
Thanks for do the hard work on this @holmesjr, and sorry it took us a little while to get to it.

This has merged in #788 and #789, with the only change being that the new parameter defaults to `optional`.

PR #788, opened 2021-01-25
Use IMDSv2 session tokens

pda:
Extracted from #786 (thanks @holmesjr!), this pull request uses [IMDSv2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) session tokens when calling the EC2 Instance Metadata Service. This paves the way for making IMDSv2 (optionally?) mandatory in buildkite/elastic-ci-stack-for-aws.

References:

[Add defense in depth against open firewalls, reverse proxies, and SSRF vulnerabilities with enhancements to the EC2 Instance Metadata Service](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/) — the November 2019 announcement of IMDSv2, including why it's important for security.

[Transition to using Instance Metadata Service Version 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html#instance-metadata-transition-to-version-2) — a guide to transition, including mention of CloudWatch metrics which can be used to confirm all IMDSv1 usage has been converted to IMDSv2.

### DIFF
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -12,8 +12,9 @@ on_error() {
 	local errorLine="$1"
 
 	if [[ $exitCode != 0 ]] ; then
+	  TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 60")
 		aws autoscaling set-instance-health \
-			--instance-id "$(curl http://169.254.169.254/latest/meta-data/instance-id)" \
+			--instance-id "$(curl -H "X-aws-ec2-metadata-token: $TOKEN" http://169.254.169.254/latest/meta-data/instance-id)" \
 			--health-status Unhealthy || true
 	fi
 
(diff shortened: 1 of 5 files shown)


chloeruka (approved):
This PR will work based on the assumption the LaunchTemplate is using [default MetadataOptions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata-metadataoptions.html#cfn-ec2-launchtemplate-launchtemplatedata-metadataoptions-httptokens), which is to say, that both IMDSv1 and IMDSv2 endpoints are available. There shouldn't ever be a situation where the IMDSv2 endpoint is NOT available, so this feels like a safe PR to merge and to smoke-test that we support this ahead of making a decision about additional stack parameters.

yob (approved):
I agree this is a good thing to do and should be non-controversial.

let's merge and dog food it on our own stacks over the next few days :+1:

PR #791, opened 2021-02-02
Add `CancelGracePeriod` param to aws template

chrisdothtml:
[It seems](https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-cancel-grace-period) this value isn't modifiable via environment variables, so it would be nice to be able to modify this value in the cloudformation config

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -46,6 +46,7 @@ Metadata:
         - ArtifactsBucket
         - AuthorizedUsersUrl
         - BootstrapScriptUrl
+        - CancelGracePeriod
         - RootVolumeSize
         - RootVolumeName
         - RootVolumeType
@@ -183,6 +184,12 @@ Parameters:
     Description: Optional - HTTPS or S3 URL to run on each instance during boot
     Type: String
     Default: ""
+
+  CancelGracePeriod:
+    Description: Optional - The number of seconds a canceled or timed out job is given to gracefully terminate and upload its artifacts
+    Type: Number
+    Default: 10
+    MinValue: 1
 
   AuthorizedUsersUrl:
     Description: Optional - HTTPS or S3 URL to periodically download ssh authorized_keys from, setting this will enable SSH ingress
@@ -900,6 +907,7 @@ Resources:
                   $Env:BUILDKITE_QUEUE="${BuildkiteQueue}"
                   $Env:BUILDKITE_AGENT_ENABLE_GIT_MIRRORS_EXPERIMENT="${EnableAgentGitMirrorsExperiment}"
                   $Env:BUILDKITE_ELASTIC_BOOTSTRAP_SCRIPT="${BootstrapScriptUrl}"
+                  $Env:BUILDKITE_CANCEL_GRACE_PERIOD="${CancelGracePeriod}"
                   $Env:BUILDKITE_AUTHORIZED_USERS_URL="${AuthorizedUsersUrl}"
                   $Env:BUILDKITE_ECR_POLICY="${ECRAccessPolicy}"
                   $Env:BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB="${BuildkiteTerminateInstanceAfterJob}"
@@ -941,6 +949,7 @@ Resources:
                   BUILDKITE_QUEUE="${BuildkiteQueue}" \
                   BUILDKITE_AGENT_ENABLE_GIT_MIRRORS_EXPERIMENT=${EnableAgentGitMirrorsExperiment} \
                   BUILDKITE_ELASTIC_BOOTSTRAP_SCRIPT="${BootstrapScriptUrl}" \
+                  BUILDKITE_CANCEL_GRACE_PERIOD="${CancelGracePeriod}" \
                   BUILDKITE_AUTHORIZED_USERS_URL="${AuthorizedUsersUrl}" \
                   BUILDKITE_ECR_POLICY=${ECRAccessPolicy} \
                   BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB=${BuildkiteTerminateInstanceAfterJob} \


pda:
Thanks for the PR @chrisdothtml (and I'm also looking into the broader issues you're having with cancel-grace-period).

I'm a bit nervous about adding a CloudFormation parameter for this (and potentially every other) agent configuration value. I'm wondering if we should bring in some other way to override agent configuration. For example an SSM Parameter Store namespace could be specified, and the values therein could be written to the agent configuration during boot. Do you think that would work for your use-case? (any thoughts @yob @chloeruka @lox?)

If we _do_ stick with this CloudFormation parameter, I believe you'll need to update `packer/linux/conf/bin/bk-install-elastic-stack.sh` and `packer/windows/conf/bin/bk-install-elastic-stack.ps1` to explicitly read it from env and write it to the config.

yob:
I share the concern about a parameter count explosion if we expose all the agent flags, but it's also not a great situation at the moment. Changing flags with a bootstrap script is possible, but super awkward.

> For example an SSM Parameter Store namespace could be specified, and the values therein could be written to the agent configuration during boot. Do

I'm not familiar enough with SSM Parameter Store to know the tradeoffs involved here, but at face value it sounds like a neat solution with lots of flexibility for the future.

pda:
The main trade-off I can think of, and not specific to Parameter Store, is that the parameters/config would be outside the CloudFormation model. Machines would pick up whatever config was set at their individual time of boot, rather than what is defined in the stack itself. So if the parameters are changed over time, a stack could end up with different config on different machines. I don't really see this as a big problem, though, and sometimes it might be advantageous.

chrisdothtml:
Agreed that doing agent configuration in the bootstrap is super awkward and that adding CloudFormation params for every agent config key would be really noisy. How about instead of an SSM param store you could just specify an S3 url for the agent config file, similar to the existing mechanism in `BootstrapScriptUrl`? Something like `AgentConfigUrl`

yob:
One complication with both the SSM and S3 approach is that there's a handful of config options that maybe do need their own stack parameter (like agent token and `disconnect-after-idle-timeout`) or arguably shouldn't be configurable at all (like `*-path`). I guess we could explicitly exclude/override those values with stack parameters?

pda:
The SSM/S3/whatever config (let's call it “user config” for now) would certainly need to be merged somehow with the base config file and the stack-driven config keys. But maybe the user config can just take highest precedence, and it's up to the user to not set values that are meant to be controlled elsewhere?

One benefit of SSM Parameter Store over S3 is that people could more easily use an outer CloudFormation / Terraform stack to manage the user config key/values, if they wanted. But it's true that a file in S3 would be more fitting with the existing elastic-stack tech, and perhaps more approachable to people unfamiliar with SSM.

dominics:
@pda Would you be amenable, if I or @yob continue this PR, to overlooking the possible slippery slope toward exposing _all_ the configuration values (which I think we all agree is suboptimal)? Might we be able to add support for adjusting (only) this agent configuration value (using the existing system), first, before we look at different ways of configuring all the other useful agent configuration values (which are...?)

One reason, in particular, that `cancel-grace-period` should be tunable (before/as opposed to making _everything_ tunable) is that the default is deficient for a number of scenarios (e.g. in mine, it cancels Terraform runs mid-flight and leaves a lock file behind). You can see this in [how the original PRs/issues](https://github.com/buildkite/agent/pull/700) talk a lot about 30 second or two minutes being a better default than 10 seconds, and how lox mentions it should probably be configurable *per-step*

I just think worrying about needing to expose all the configuration values is probably premature. YAGNI: I've not needed anything except `cancel-grace-period` (or things _already_ exposed via CloudFormation parameters) in all the time we've been using the AWS stack (years).

dominics:
For ease of testing: #881 is this PR rebased, with the fixes that were asked for to the `bk-install-elastic-stack` scripts (to write out the config value). Did this mainly so I can use it myself. Although, I'm curious, @yob said:

> Changing flags with a bootstrap script is possible, but super awkward.

Could you describe how? Maybe I can use that? Do you have to kill the agent process that's running the bootstrap script and relaunch it somehow? Is there a signal to get it to re-read its configuration?

AFAICT by testing, by the time the bootstrap script is running, it's too late for most sensible methods to set this flag. So all that's left is custom _cloud-init_ code, which is another level of difficulty? Is there an easier way?

keithduncan:
> Could anyone describe how? Maybe I can use that? Do you have to relaunch the agent process somehow? Is there a signal to get it to re-read its configuration?

The [BootstrapScriptUrl](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/aws-stack.yml#L185-L188) parameter is [executed](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/4b8a67ee26b85d5877b87e3117a789cc7fdc4b90/packer/linux/conf/bin/bk-install-elastic-stack.sh#L135-L139) on every instance boot [after the agent config is written](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/4b8a67ee26b85d5877b87e3117a789cc7fdc4b90/packer/linux/conf/bin/bk-install-elastic-stack.sh#L103-L119) but [before the agent is started](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/4b8a67ee26b85d5877b87e3117a789cc7fdc4b90/packer/linux/conf/bin/bk-install-elastic-stack.sh#L161-L162). So you should be able to append settings to the config file using the bootstrap script before the agent process ever starts up.

dominics:
> append settings to the config file using the bootstrap script

Brilliant, the missing link was that we always tried to set environment variables using the bootstrap before. A pretty easy workaround after all!

```
echo "cancel-grace-period=30" >> /etc/buildkite-agent/buildkite-agent.cfg
```

DrJosh9000:
https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1258

PR #793, opened 2021-02-04
Adds ScheduledAction for scaling min/max instances up and down (e.g. during business hours)

chrisfowles:
This adds a scheduled action to the auto-scaling group to automatically scale up and/or down the min instance count, for example to set a minimum number of agents during business hours.

### DIFF
--- a/.buildkite/steps/publish.sh
+++ b/.buildkite/steps/publish.sh
@@ -28,8 +28,10 @@ buildkite-agent artifact download build/mappings.yml build/
 echo "--- Fetching latest git tags"
 git fetch --tags
 
-echo "--- Building :cloudformation: templates"
+echo "--- Building :cloudformation: CloudFormation templates"
 make build/aws-stack.yml
+
+echo "--- Uploading :cloudformation: CloudFormation templates"
 
 # Publish the top-level mappings only on when we see the most recent tag on master
 if is_latest_tag ; then
@@ -40,8 +42,6 @@ if is_latest_tag ; then
 else
   echo "Skipping publishing latest, '$BUILDKITE_TAG' doesn't match '$(git describe origin/master --tags --match='v*')'"
 fi
-
-echo "--- Uploading :cloudformation: templates"
 
 # Publish the most recent commit from each branch
 s3_upload_templates "${BUILDKITE_BRANCH}/"


templates/aws-stack.yml
```diff
@@ -1016,6 +1051,24 @@ Resources:
       AutoScalingReplacingUpdate:
         WillReplace: true
 
+  AutoScalingStartSchedule:
+    Type: AWS::AutoScaling::ScheduledAction
+    Condition: HasStartSchedule
+    Properties: 
+      AutoScalingGroupName: !Ref AgentAutoScaleGroup
+      MinSize: !Ref ScheduleMinSize
```
pda, line 1059: I wonder if `ScheduleMinSize ` should get the same conditional treatment as `ScheduleMaxSize`.
Some teams might want to adjust their MaxSize on a schedule but leave MinSize zero regardless.
Would that make sense?

templates/aws-stack.yml
```diff
@@ -400,6 +404,28 @@ Parameters:
       - "false"
     Default: "false"
 
+  StartSchedule:
+    Type: String
+    Description: "The recurring schedule for changing agent count, in Unix cron syntax format."
+    Default: ""
+    AllowedPattern: "^$|^(((\\d+,)+\\d+|(\\d+(\\/|-)\\d+)|\\d+|\\*) ?){5,7}$"
```
pda, line 411: Any citation for this regexp? Or have you written it just for this? I don't see anything wrong with it, but regular expressions can be a bit “write only” :) And unfortunately we can't really unit test it here. Anyway, pretty low impact if it does/doesn't allow something valid, we can iterate on it.

For reference — the [`AWS::AutoScaling::ScheduledAction` documentation for `Recurrence`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-as-scheduledaction.html#aws-resource-as-scheduledaction-properties) says:

> The recurring schedule for this action, in Unix cron syntax format. For more information about cron syntax, see [Crontab](http://crontab.org).

templates/aws-stack.yml
chrisfowles: I can see that use case - I'll add it in.

templates/aws-stack.yml
chrisfowles: Written by myself unfortunately (for both of us). Mainly added in as the error you get here if it's wrong is pretty obscure, I think the only way I found out what was wrong was by tracing the error message back to the library that AWS is using for the scheduling.

I've tested manually with https://regexr.com/ a fair bit, but happy to remove this check if you want to err on the side of not accidentally blocking valid expressions.

joshross12:
Are there plans to add this feature to the next release?

keithduncan:
Hi @joshross12, there are no plans to release this in the next version specifically, but the broader theme of more configurable and specific scaling options like this is something we’re interested in expanding.

Our challenge is how to package and present the exploding number of configuration options in a way that doesn’t overload a single template for all customers. You voice counts and thank you for letting us know that this is a particular scaling mode you’d like to see available 😃

PR #795, opened 2021-02-08
arm64: docker-compose→paramiko→cryptography version pinned

pda:
docker-compose depends on the cryptography package, [v3.4 of which introduces a build dependency on rust](https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#34---2021-02-07); let's avoid that for now by constraining the install to `cryptography<3.4`.

This should be unpinned ASAP; hopefully `docker-compose` will offer binary download for arm64 at some point:
https://github.com/docker/compose/issues/7472

### DIFF
--- a/packer/linux/scripts/install-docker.sh
+++ b/packer/linux/scripts/install-docker.sh
@@ -38,7 +38,15 @@ if [ "${MACHINE}" == "x86_64" ]; then
 	docker-compose --version
 elif [[ "${MACHINE}" == "aarch64" ]]; then
   sudo yum install -y gcc-c++ libffi-devel openssl11 openssl11-devel python3-devel
-  sudo pip3 install docker-compose
+
+  # docker-compose depends on the cryptography package, v3.4 of which
+  # introduces a build dependency on rust; let's avoid that for now.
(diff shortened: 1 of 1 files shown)


pda:
```
Collecting cryptography<3.4 (from -c /tmp/docker-compose-pip-constraint (line 1))
  Downloading https://files.pythonhosted.org/packages/d4/85/…/cryptography-3.3.2.tar.gz (539kB)
```
— https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/2278#39326237-03a5-44fb-b4c5-27d3ba9e8b3b

matthewd (approved):

yob:
thanks for squashing this bug :bug:

pda:
Notes for posterity:
We subsequently read about the availability of _wheels_ (python/pip precompiled binary components? 🤷‍♂️) being available for linux/aarch64, which would avoid the rust build dependency, but it seems the version of pip & pip3 in Amazon Linux 2 LTS is too old to use them, for whatever reason. But that's probably the solution we should pursue to get this dependency unpinned.

PR #796, opened 2021-02-08
CI: fix cleanup process

pda:
CI for this project includes launching test CloudFormation stacks. There's a cleanup step at the end which attempts to delete the stacks launched by that build, and also searches for older stacks/buckets etc that were missed by previous cleanups. However, the stack naming evolved since the cleanup script was written, so it wasn't finding the stacks/buckets/etc.

This PR updates the name patterns to match the newer stacks/buckets.

### DIFF
--- a/.buildkite/steps/cleanup.sh
+++ b/.buildkite/steps/cleanup.sh
@@ -45,7 +45,7 @@ aws s3api list-buckets \
   --output text \
   --query "$(printf 'Buckets[?CreationDate<`%s`].[Name]' "$cutoff_date" )" \
   | xargs -n1 \
-  | grep -E 'buildkite-aws-stack-test-(\d+-)?managedsecrets' \
+  | grep -E 'buildkite-aws-stack-test-.*-managedsecretsbucket' \
   | xargs -n1 -t -I% aws s3 rb s3://% --force
 
 echo "--- Deleting old cloudformation stacks"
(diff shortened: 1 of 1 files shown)


.buildkite/steps/cleanup.sh
```diff
@@ -44,15 +45,15 @@ aws s3api list-buckets \
...
+  | grep -E 'buildkite-aws-stack-test-.*-managedsecretsbucket' \
   | xargs -n1 -t -I% aws s3 rb s3://% --force
 
 echo "--- Deleting old cloudformation stacks"
 aws cloudformation describe-stacks \
   --output text \
   --query "$(printf 'Stacks[?CreationTime<`%s`].[StackName]' "$cutoff_date" )" \
   | xargs -n1 \
-  | grep -E 'buildkite-aws-stack-test-\d+' \
+  | grep -E 'buildkite-aws-stack-test-(linux|windows)-(amd64|arm64)-\d+' \
```
pda, line 56: Looks like `\d` etc only ever worked on BSD grep (macOS local dev), not GNU grep on the real CI servers.

```
bash-4.2# grep --version | head -n1
grep (GNU grep) 2.20
bash-4.2# echo foo-123 | grep -E 'foo-\d+' | wc -l
0
```

```
pda@paulbookpro ~ $ grep --version | head -n1
grep (BSD grep) 2.5.1-FreeBSD
pda@paulbookpro ~ $ echo foo-123 | grep -E 'foo-\d+' | wc -l
       1
```

pda: Fixed in 6f94605

yob (approved):
Oops. I must admit I hadn't been paying much attention to the logs of the cleanup step. Thanks for digging into it!

Issue #798, opened 2021-02-11
Bug: Have to add DOCKER_CLI_EXPERIMENTAL to get docker manifests pushed


nitrocode:
## Problem

The `--platform` option works

```
...
docker build -t snip.dkr.ecr.us-west-2.amazonaws.com/snip --build-arg BASEIMAGE=arm64v8/ubuntu --platform linux/arm64 -f Dockerfile.staging ../
Sending build context to Docker daemon  134.2MB Docker daemon  557.1kB
[2021-02-11T16:31:47Z] Step 1/26 : ARG BASEIMAGE
...
```

The `manifest create` did not

```
...
[2021-02-11T16:49:23Z] docker manifest create snip.dkr.ecr.us-west-2.amazonaws.com/snip --amend snip.dkr.ecr.us-west-2.amazonaws.com/snip --amend snip.dkr.ecr.us-west-2.amazonaws.com/snip
[2021-02-11T16:49:23Z] docker manifest create is only supported on a Docker cli with experimental cli features enabled
...
```

## Workaround

Adding the `DOCKER_CLI_EXPERIMENTAL=enabled` to the pipeline steps seems to work.

```
      env:
          DOCKER_CLI_EXPERIMENTAL: enabled
```

We have `EnableDockerExperimental=true` set in the cloudformation stack but this does not add the above variable. It adds `DOCKER_EXPERIMENTAL` instead.

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/b84d760a66f544c5f582702ba207d280fbac20a1/templates/aws-stack.yml#L934-L938

## Proposed solution

Should both `DOCKER_CLI_EXPERIMENTAL` and `DOCKER_EXPERIMENTAL` be added if `EnableDockerExperimental=true` ? 

or would it be better to add the following to `/etc/docker/daemon.json`

```json
{
  "experimental": true
}
```

mogggggg:
I agree, having the `EnableDockerExperimental ` parameter set the experimental mode on both the daemon and the CLI would be useful for us too.

yob:
> Or would it be better to add the following to /etc/docker/daemon.json

To enable experimental cli behaviour, does that need to go into `/etc/docker/daemon.json` or `~/.docker/config.json`?

PR #801, opened 2021-02-25
Add DOCKER_CLI_EXPERIMENTAL env variable

nitrocode:
Closes https://github.com/buildkite/elastic-ci-stack-for-aws/issues/798

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -895,6 +895,7 @@ Resources:
                   <powershell>
                   $Env:DOCKER_USERNS_REMAP="${EnableDockerUserNamespaceRemap}"
                   $Env:DOCKER_EXPERIMENTAL="${EnableDockerExperimental}"
+                  $Env:DOCKER_CLI_EXPERIMENTAL="${EnableDockerExperimental}"
                   powershell -file C:\buildkite-agent\bin\bk-configure-docker.ps1 >> C:\buildkite-agent\elastic-stack.log
 
                   $Env:BUILDKITE_STACK_NAME="${AWS::StackName}"
@@ -934,6 +935,7 @@ Resources:
                   Content-Type: text/cloud-boothook; charset="us-ascii"
                   DOCKER_USERNS_REMAP=${EnableDockerUserNamespaceRemap} \
                   DOCKER_EXPERIMENTAL=${EnableDockerExperimental} \
+                  DOCKER_CLI_EXPERIMENTAL=${EnableDockerExperimental} \
                     /usr/local/bin/bk-configure-docker.sh
                   --==BOUNDARY==
                   Content-Type: text/x-shellscript; charset="us-ascii"


templates/aws-stack.yml
```diff
@@ -934,6 +935,7 @@ Resources:
                   Content-Type: text/cloud-boothook; charset="us-ascii"
                   DOCKER_USERNS_REMAP=${EnableDockerUserNamespaceRemap} \
                   DOCKER_EXPERIMENTAL=${EnableDockerExperimental} \
+                  DOCKER_CLI_EXPERIMENTAL=${EnableDockerExperimental} \
                     /usr/local/bin/bk-configure-docker.sh
```
yob, line 939: It's not immediately obvious, but I think this is passing `DOCKER_EXPERIMENTAL` to the `bk-configure-docker.sh` script, which then writes the experimental flag to `docker.json`: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/5ba33e8793a5cd6a7f60facc6806d949bfa8e10f/packer/linux/conf/bin/bk-configure-docker.sh#L18.

I'm not super familiar with `DOCKER_CLI_EXPERIMENTAL`, but based on the issue description in #798 is having experimental=true in `daemon.json` enough to enable cli experiments?

nitrocode: Interesting. It looks like this has been deprecated as of [v20.10.3](https://github.com/docker/cli/releases/tag/v20.10.3) of docker (See https://github.com/docker/cli/commit/48822564b6ed11e8e9278f71ba1949eaeb3392c9)

> The `DOCKER_CLI_EXPERIMENTAL` environment variable and the corresponding `experimental` field in the CLI configuration file are deprecated. Experimental features will be enabled by default, and these configuration options will no longer be functional.

However, it does seem like `DOCKER_EXPERIMENTAL` may have been the older name and the newer name is `DOCKER_CLI_EXPERIMENTAL` on the docker 19.x branch.

yob: Hmmmm. If the experimental features are available by default in docker 20.10.x, maybe that's an extra point for the idea of releasing a stack with newer docker. Maybe along with the other changes we're considering for a 6.x release.

In the short term, I'm still a bit confused about enabling the experimental cli  features in the current stack.

It looks like we could write `experimental: true` to ~/.docker/config.json, however the location of that config file changes per job (since #756) so we can't do it at instance boot time.

Having `DOCKER_CLI_EXPERIMENTAL` in the job execution environment might be easier, but unless I'm reading this PR wrong I don't think this diff will achieve that.

nitrocode: Yes, the upgrade to docker 20.x would be the best. Perhaps an experimental feature for the next 5.x could be using the latest docker to prep for that as a default for 6.x?

You're right if there are 2 docker configs then im unsure if a new ec2 env var on boot would solve this. Currently the experimental flag gets set to true in /etc/docker/daemon.json already but then the docket cli experimental flag needs to be set as well, probably for the new docker config.

Short term potential fixes. Thoughts?

* we could figure out how to also append the new job specific docker config with the same values.
* an experimental feature for the next 5.x could be using the latest docker to prep for that as a default for 6.x

PR #803, opened 2021-02-25
remove AMI lookup in the Makefile

yob:
Prior to PR #636 we looked up the source AMI to use for packer builds in the Makefile. However in #636 we instead configured packer with some AMI filter rules and allow it to choose the base AMI itself.

These bits of the Makefile are unused.


sj26 (approved):
Nice

PR #805, opened 2021-02-25
Add missing else condition to iam permission boundary

nitrocode:
Original pr https://github.com/buildkite/elastic-ci-stack-for-aws/pull/767

Reverted https://github.com/buildkite/elastic-ci-stack-for-aws/pull/802

Added missing else condition. If the boundary is not supplied, the `!Ref "AWS:NoValue"` should null out the argument.

Opened this pr as the last one https://github.com/buildkite/elastic-ci-stack-for-aws/pull/804 closed due to branch rename.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -308,6 +308,11 @@ Parameters:
     Description: Optional - A name for the IAM Role attached to the Instance Profile
     Default: ""
 
+  InstanceRolePermissionsBoundaryARN:
+    Type: String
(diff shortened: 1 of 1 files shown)


yob:
lovely, thanks!

nitrocode:
Thanks for merging. When do you think this change could be cut into a release?

yob:
I suspect we'll wait for another agent release. Given it will include a functional change as well, I assume we'll be looking at `5.3.0`

Issue #792, opened 2021-02-03
A list of breaking changes we're holding back for 6.0


yob:
We have a few breaking changes we might make in 6.0, whenever that ends up being. I considered making a GitHub milestone for this list, but we don't have a culture of using them and I worried we might overlook it.

* [x] update awscli to v2 #688
* [ ] remove the `IMDSv2Tokens` param added in #789 and just make v2 tokens required all the t ime
* [x] remove deprecated `SpotPrice` param #710
* [ ] Enable SSM access and disable SSH access #738 
* [x] Rename SecurityGroupId to SecurityGroupIds?
* [x] Make instance names default to the stack name
* [x] Rename base branch from master to main - this would cause people using master branch releases to stop receiving builds

See also: [Agent: A list of breaking changes we're holding back for 4.0](https://github.com/buildkite/agent/issues/1391)

nitrocode:
Renaming `SecurityGroupId` to `SecurityGroupIds` would be good too

woodhull:
Update docker to 20.x?

chloeruka:
> Update docker to 20.x?

@woodhull Docker bumps are not usually considered breaking, are there significant implications of moving to v20?

@yob I'd also like to see #738 if we don't merge it sooner.

yob:
Thanks all - list updated

jdub:
If you're doing `SecurityGroupIds`, perhaps `ManagedPolicyARNs`? Just discovered it. 😆

nitrocode:
@chloeruka that wouldn't be a breaking change since ssh access would still be available just through ssm instead of port 22.

If it's not a breaking change, could it be added prior to a 6.0 release?

chloeruka:
It's not so much the availability of equivalent/replacement features, but removal of parameters that would make us consider it more appropriate for a major version. Changing things around might require users adjust parameters or automation surrounding their stacks. We want low friction Cloudformation upgrades since it's currently tied to Agent updates.

nitrocode:
I'd like to second the docker `20.x` upgrade as that would allow us to no longer have to set environment variables `DOCKER_EXPERIMENTAL` and `DOCKER_CLI_EXPERIMENTAL` which would deprecate the `EnableDockerExperimental` parameter on the stack.

triarius:
v6 is out, so I'm going to close this.

PR #808, opened 2021-03-03
[MOD] Upgrade Docker and docker-compose to most recent

ovcharenko:
Any obstacles for such upgrade?

### DIFF
--- a/packer/linux/scripts/install-docker.sh
+++ b/packer/linux/scripts/install-docker.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 set -eu -o pipefail
 
-DOCKER_VERSION=19.03.13
+DOCKER_VERSION=20.10.4
 DOCKER_RELEASE="stable"
-DOCKER_COMPOSE_VERSION=1.27.4
+DOCKER_COMPOSE_VERSION=1.28.5
 MACHINE=$(uname -m)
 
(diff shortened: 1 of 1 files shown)


yob (commented):
The docker-compose change is definitely uncontroversial.

There has been some discussion in #792 about whether upgrading docker might be a breaking change that should wait for a 6.0 release. I don't think we know enough about the 19.03 to 20.10 upgrade to answer that yet - do you feel across how compatible this would be for all users of the stack?

ovcharenko:
> The docker-compose change is definitely uncontroversial.
> 
> There has been some discussion in #792 about whether upgrading docker might be a breaking change that should wait for a 6.0 release. I don't think we know enough about the 19.03 to 20.10 upgrade to answer that yet - do you feel across how compatible this would be for all users of the stack?

We didn't notice any downside on our projects. But still asking community for possible cases...

chloeruka:
There's an apple silicon compatibility note in 20.10.6 that might make that worth considering over 20.10.4

yob:
Apologies @ovcharenko applying these upgrades on master, but over in #825 and #826. The main difference on those PRs is applying matching upgrades to our windows AMIs.

Thanks for the nudge to get it done though :heart: 

 The upgrades are included in 5.3.0.

PR #557, opened 2019-03-28
Optionally mount Instance Storage on boot

lox:
Since lots of instances seem to have fancy fast instance storage now, this mounts whatever is available at `/ephemeral` and uses that for docker storage.  

Closes #464.

PR #809, opened 2021-03-09
ensure that custom build paths can have their permissions fixed

albertywu:
### Issue

The Buildkite `BuildPath` [config](https://github.com/buildkite/agent/blob/main/agent/job_runner.go#L494) allows us to set a custom build path. However, this build path is hardcoded to `/var/lib/buildkite-agent/builds` in the [fix-buildkite-agent-builds-permissions script](https://github.com/uber-workflow/elastic-ci-stack-for-aws/blob/master/packer/linux/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions#L77), which results in this script not setting permissions properly.

We ran into this issue when we set a custom build path, and subsequently noticed that some docker artifacts on host could not be cleaned up properly by `buildkite-agent`, because some files in the workspace (residual artifacts of docker) were owned by other users.

---

### Fix

The fix is to add the build path as a parameter to the `fix-buildkite-agent-builds-permissions` script.

---

### Test Plan

`docker-compose -f docker-compose.unit-tests.yml run unit-tests` ran successfully

### DIFF
--- a/packer/linux/conf/buildkite-agent/hooks/environment
+++ b/packer/linux/conf/buildkite-agent/hooks/environment
@@ -118,7 +118,7 @@ if [[ "${DOCKER_USERNS_REMAP:-false}" == "false" ]] ; then
   # => "my-pipeline"
 
   # Now we can pass this to the sudo script which will validate it before safely chmodding:
-  echo "Fixing permissions for '${AGENT_DIR}/${ORG_DIR}/${PIPELINE_DIR}'..."
-  sudo /usr/bin/fix-buildkite-agent-builds-permissions "${AGENT_DIR}" "${ORG_DIR}" "${PIPELINE_DIR}"
+  echo "Fixing permissions for '${BUILDKITE_BUILD_PATH}/${AGENT_DIR}/${ORG_DIR}/${PIPELINE_DIR}'..."
+  sudo /usr/bin/fix-buildkite-agent-builds-permissions "${BUILDKITE_BUILD_PATH}" "${AGENT_DIR}" "${ORG_DIR}" "${PIPELINE_DIR}"
   echo
 fi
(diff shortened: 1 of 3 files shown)


sj26:
Ooh, this is an interesting edge to the problem! It's very locked down so we don't allow chmod-ing or chown-ing arbitrary file paths. Is there a way that `$BUILDKITE_BUILD_PATH` can be controlled and changed before this script is run which might allow running this script on arbitrary paths?

albertywu:
> Ooh, this is an interesting edge to the problem! It's very locked down because we don't want to allow chmod-ing or chown-ing arbitrary file paths. Is there a way that $BUILDKITE_BUILD_PATH can be controlled and changed before this script is run which might allow running this script on arbitrary paths?

I don't believe so. As I understand it:

1) `$BUILDKITE_BUILD_PATH` is set on the [env](https://github.com/buildkite/agent/blob/main/agent/job_runner.go#L494) of the bootstrap process.
2) the permissions-fixing script runs in the agent [env hook](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/linux/conf/buildkite-agent/hooks/environment#L122), which should run prior to all other hooks per the docs [here](https://buildkite.com/docs/agent/v3/hooks#available-hooks)

Since there's no user-runnable code or hooks that can modify the `env` of the buildkite-agent bootstrap process *prior* to the agent env hook, I don't think it's possible to remap `$BUILDKITE_BUILD_PATH` for a running job.

yob:
We remain a bit nervous about this @albertywu, mainly the risk of us overlooking a way for `BUILDKITE_BUILD_PATH` to be changed and arbitrary files to be chmod.

I believe the root issue that led to this PR was exploring this was exploring ways to use instance local storage for the builds directory to improve performance. Work that we started in #557, but haven't managed to land yet. We'd **really** love to get that merged though, it's a great feature that will be useful to many customers.

I think our ideal path forward would be to re-visit #557 and allow stacks to opt in to using instance local storage, without a need to customise the build path or change `fix-buildkite-agent-builds-permissions`.

PR #811, opened 2021-03-14
Replace awslogs with the cloudwatch-agent

### LINKED ISSUE #709: CreateLogGroup service limits
The elastic stack exports logs to cloudwatch logs. The official aws cloudwatch logs exporter seems to call CreateLogGroup for each exported log on each host as it boots, and for some customers this is leading to hitting service limits and creating elastic stacks failing

It looks like we're using the awslogs cloudwatch logs agent:
https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v4-development/packer/scripts/install-awslogs.sh
https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/UsePreviousCloudWatchLogsAgent.html
​
​and configuring it to pump files up to log groups in  a pretty conventional way:
​https://github.com/buildkite/elastic-ci-stack-for-aws/blob/v4-development/packer/conf/awslogs/awslogs.conf

We're exporting groups including:

- /buildkite/buildkite-agent   
- /buildkite/cfn-init   
- /buildkite/cloud-init   
- /buildkite/cloud-init/output   
- /buildkite/docker-daemon   
- /buildkite/elastic-stack   
- /buildkite/elastic-stack-init   
- /buildkite/lifecycled   
- /buildkite/system

​The reference docs suggest that this will only create the log group if it doesn't exist:
​https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AgentReference.html

The implementation which has helpfully been uploaded here seems to be always create, and swallow errors:
https://github.com/jinty/awscli-cwlogs-debian/blob/c8e4a1d5a0d9ec771581967e4de63407b8d0e9ac/cwlogs/push.py#L1314-L1324

But the call is made, so the limits are utilized.

### LINKED ISSUE #713: Migrate from awslogs to CloudWatch Agent
We currently use the deprecated aws logs tool to send various system and agent logs to cloudwatch.

awslogs has a scale issue for large customers (see #709), and in any case we ideally don't want to ship a deprecated tool. The replacement tool is [Cloudwatch Agent](https://aws.amazon.com/about-aws/whats-new/2020/09/amazon-cloudwatch-agent-now-open-source-and-included-with-amazon-linux-2/) which can do logs (and metrics via a statsd interface).

yob:
We've long used awslogs to send logs from elastic stack instances to cloudwatch logs. However, it's deprecated and AWS now recommend using the "cloudwatch agent". For example, there's currently a banner in the [awslogs docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AgentReference.html) that says:

> This reference is for the older CloudWatch Logs agent, which is on the
> path to deprecation. We strongly recommend that you use the unified
> CloudWatch agent instead

Our windows AMIs already use the cloudwatch agent, and this finally updates the two linux AMIs (amd64/arm64) to use it as well. Although windows was already using cloudwatch logs, I have renamed a few files to keep them consistent across linux and windows.

The new linux config file was generated by booting the 5.2.0 linux/amd64 stack AMI, installing the agent via yum, and then running the wizard to convert the legacy awslogs.conf:

    /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-config-wizard

The new agent can apparently do do tracing and metrics (including running a local statsd compatible interface). I've left them disabled for now, and focused just on replacing the deprecated awslogs.

We believe this might also resolve a known issue (#709) where the awslogs tool calls the CreateLogGroup API endpoint over and over. For large Buildkite customers, this can result in the regional quota for CreateLogGroup being hit and some logs not being recorded.

Fixes #713
Fixes #709


yob:
While developing this, I created a new elastic stack based on this branch and confirmed the 7 log groups are still being updated. I wasn't sure if extra permissions would be required on the instance IAM role, but apparently not.

I assume we might need extra permissions if we wanted to experiment with tracing or metrics.

chloeruka (approved):
Nice! Looks good to me! 🚀

packer/linux/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -31,20 +31,6 @@ trap 'on_error $LINENO' ERR
...
-cwlogs = cwlogs
-[default]
-region = $AWS_REGION
-EOF
-
-systemctl enable awslogsd.service
-
-# Start logging daemons as soon as possible to ensure failures in this script get sent
-systemctl restart rsyslog
-systemctl restart awslogsd
```
yob, line 46: I removed all this on-boot stuff because the cloudwatch-agent is installed and configured in the AMI, and there should be no further config required at boot time.

pda:
Nice! 👌🏼

PR #813, opened 2021-03-16
Bugfix: bugfix/812/BuildkiteAgentTokenParameterStorePath validation (#812)

### LINKED ISSUE #812: Template bug - relying on a parameter to have a `/` in front of it
So AWS SSM parameter doesn't necessarily has to have a leading slash to it, e.g. `buildkite-token` is a perfectly valid SSM Parameter.

However, the AWS stack template seems to assume otherwise:
https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/aws-stack.yml#L707

Leading to invalid references in roles, like:
```
"Resource": "arn:aws:ssm:ap-southeast-2:103639821168:parameterbuildkite-token-default"
                                                                                                                   ^^^
``` 
, instead of:
```
"Resource": "arn:aws:ssm:ap-southeast-2:103639821168:parameter/buildkite-token-default"
                                                                                                                   ^^^
```

jufemaiz:
Resolves #812 

Ref: https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-su-create.html

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -113,6 +113,7 @@ Parameters:
     Description: AWS SSM path to the Buildkite agent registration token (this takes precedence over BuildkiteAgentToken)
     Type: String
     Default: ""
+    AllowedPattern: "^/[a-zA-Z0-9_.-]+$"
 
   BuildkiteAgentTokenParameterStoreKMSKey:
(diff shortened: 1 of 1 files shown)


yob:
Thanks @jufemaiz. I had to tweak the AllowedPattern a little to allow blank strings as well (630f8335d8085ceb55c0b048dcbbc34824994f4c).

Issue #817, opened 2021-03-23
Allow buildkite to use ssm to run commands into another instance with session-manager-plugin


nitrocode:
We run packer scripts to create new AMIs using this configuration which allows us to use SSM

```
      "ssh_interface": "session_manager",
      "communicator": "ssh",
```

The buildkite instances do not have the [`session-manager-plugin`](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html) installed and require this to access this feature. 

It would be nice if this plugin could come preinstalled on the buildkite x86 and arm AMIs so we don't have to install it ourselves.

Similar but different issues https://github.com/buildkite/elastic-ci-stack-for-aws/issues/620 https://github.com/buildkite/elastic-ci-stack-for-aws/pull/738

yob:
Hi @nitrocode,

I must admit to being a bit unfamiliar with session manager.

If we added session-manager-plugin to the elastic stack AMI, would it allow remote access into the instance (eg. what #620 and #738 are proposing) or is it about starting outbound sessions to other instances?

nitrocode:
It would only allow outbound sessions to other instances.

Those other proposals would allow ssm to be exclusively accessed by ssm which would allow them to remove sg rule port 22 ingress.

This plugin pre installed would allow us to remove the ssh security group from new instances spun up in the packing process.

yob:
That sounds like a reasonable addition to me. We're pretty cautious about adding new tools to the base image, but AWS native tools that help the stack interact with the AWS eco-system are less controversial.

We'd be happy to review a PR if you open one.

I note the [installation instructions](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html#install-plugin-linux) recommend downloading an rpm from `latest`. If you can find  URLs with a fixed version that would be preferable. We prefer to have the version locked and make updates intentional (and changelog-able), rather than each AMI build automatically fetching the latest version.

nitrocode:
Looks like the latest version is `1.2.30.0` and can be pinned like so.

https://s3.amazonaws.com/session-manager-downloads/plugin/1.2.30.0/linux_64bit/session-manager-plugin.rpm
https://s3.amazonaws.com/session-manager-downloads/plugin/1.2.30.0/linux_arm64/session-manager-plugin.rpm

Id imagine that if the plugin were to be added that it should be added to both windows and Linux.

PR #818, opened 2021-03-24
Session manager plugin

nitrocode:
Closes https://github.com/buildkite/elastic-ci-stack-for-aws/issues/817

Not sure how to install the executable silently on Windows.

### DIFF
--- a/packer/linux/buildkite-ami.json
+++ b/packer/linux/buildkite-ami.json
@@ -68,6 +68,10 @@
     },
     {
       "type": "shell",
+      "script": "scripts/install-session-manager-plugin.sh"
+    },
+    {
+      "type": "shell",
       "inline": [
         "rm /home/ec2-user/.ssh/authorized_keys"
       ]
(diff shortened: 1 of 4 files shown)


yob:
I've run a build and confirmed this works as expected.

is it possible to rebase onto master (to get #819) and remove the merge commits?

yob:
One thing I haven't done is actually test the plugin works. We don't really use SSM yet though, so that test wouldn't be straight forward. Have you been able to run a build with one of these AMIs and confirm the plugin works as intended?

nitrocode:
> I've run a build and confirmed this works as expected.

Perfect.

> is it possible to rebase onto master (to get #819) and remove the merge commits?

Done.

> One thing I haven't done is actually test the plugin works. We don't really use SSM yet though, so that test wouldn't be straight forward. Have you been able to run a build with one of these AMIs and confirm the plugin works as intended?

I did not build an AMI but I did run the script on an x86 and an arm machine to make sure it installs correctly. Then I gave the iam role of my instance access to `ssm:StartSession` across `*` resources. Finally, I ran this command from the buildkite ec2 instance where I installed the `session-manager-plugin` and it worked as expected.

```bash
$ aws --region us-west-2 ssm start-session --target i-snip

Starting session with SessionId:
```

yob:
Lovely, thanks!

PR #825, opened 2021-04-15
Update docker-compose on all operating systems to 1.28.x

yob:
* Linux 1.28.6
* Windows 1.28.5 (the latest available on chocolatey)

There's no specific features we need, we just want to do regular small bumps so the elastic stack never falls behind, and also avoid big changes in behaviour from release to release

1.29.1 is the latest released version, but I stuck with 1.28.x so we can have consistent minor versions across linux and windows. There's so new features in 1.29.x that can't wait for the next elastic stack release.


chloeruka (approved):
LGTM!

yob:
rebased onto master to resolve a conflict and force pushed

PR #829, opened 2021-04-27
Remove IAM roles for SNS and SQS

chloeruka:
These appear to be vestigial since we no longer use SNS/SQS queue for agent lifecycle operations when we removed that for lifecycled in #135.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -737,12 +737,6 @@ Resources:
               - logs:PutLogEvents
               - logs:DescribeLogStreams
             Resource: "*"
-          - Effect: Allow
-            Action:
-              - sqs:*
-              - sns:Unsubscribe
-              - sns:Subscribe
-            Resource: "*"
       Roles:
         - !Ref IAMRole
 


sj26:
I think we are using lifecycled again, but without the SNS/SQS part, only to handle graceful termination:
https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/linux/scripts/install-lifecycled.sh
https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/linux/conf/bin/bk-install-elastic-stack.sh#L141-L148

So I believe it's safe to remove these permissions. Nice.

sj26 (approved):

yob (approved):
> So I believe it's safe to remove these permissions. Nice.

This matches my understanding as well.

The only caveat is there's an open question about whether we need to bring lifecycle hooks back to help ensure graceful agent termination in some non-scale down situations (AZ rebalancing, with mixed on-demand/spot ASGs when some on-demand instances should be terminated and replaced with spot, etc).

I think this PR is fine though - we can restore the permissions if we ever decide to add lifecycle hooks back in.

PR #830, opened 2021-04-27
Fix linter issue with steps/copy.sh

chloeruka:
Not sure how this slipped by!


chloeruka:
Ah, no, this will cause a different issue. Oh bash.

PR #831, opened 2021-04-27
Fix linter issue in steps/copy.sh

chloeruka:
Shellcheck isn't happy that we use default behaviour to split on a bash array. You'll be directed toward _adding_ quotes in [SC2048](https://www.shellcheck.net/wiki/SC2048) and then be warned to _remove_ them to comply with [SC2066](https://github.com/koalaman/shellcheck/wiki/SC2066). The answer is to silence the warning, unfortunately.

### DIFF
--- a/.buildkite/steps/copy.sh
+++ b/.buildkite/steps/copy.sh
@@ -136,6 +136,7 @@ linux_arm64_source_image_name=$(get_image_name "$linux_arm64_source_image_id" "$
 windows_amd64_source_image_name=$(get_image_name "$windows_amd64_source_image_id" "$source_region")
 
 # Copy to all other regions
+# shellcheck disable=SC2048
 for region in ${ALL_REGIONS[*]}; do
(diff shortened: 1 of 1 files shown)


yob (approved):
weird, have we changed this script recently or has a new version of shellcheck started pinging us for something that previously wasn't an issue?

yob:
> has a new version of shellcheck started pinging us for something that previously wasn't an issue?

I found this in the build output

> Unable to find image 'koalaman/shellcheck:latest' locally

oh mutable docker tags, you are not my friend :upside_down_face:

sj26:
Is this popping up in shellcheck because we're using the `latest` docker image? Perhaps it's a new check? I've noticed they've started presuming bash 4 in a bunch of places which might not have it (so are being a bit stricter about using arrays etc).

chloeruka:
Yeah, shellcheck updated; I've been bitten by this elsewhere.

PR #835, opened 2021-04-30
allow multiple / characters in BuildkiteAgentTokenParameterStorePath

### LINKED ISSUE #834: BuildkiteAgentTokenParameterStorePath does not support multiple hierarchies in v5.3.0
Hi team!
As mentioned in [Slack](https://buildkite-community.slack.com/archives/C19BG77H9/p1619670223006000) the CloudFormation `AllowedPattern` regex was [changed](https://github.com/buildkite/elastic-ci-stack-for-aws/commit/630f8335d8085ceb55c0b048dcbbc34824994f4c) in v5.3.0 which has broken parameters using multiple hierarchies.

As an example our parameter used to be `/build/buildkite-agent-token` but we've had to update it to `/buildkite-agent-token` as part of the upgrade.

Not a huge issue for us, but I'm sure there will be other people affected 🙏

yob:
In #813 we added a regexp validation to this stack parameter that was intended to check the parameter is either an empty string or a path that starts with a / character.

The list of characters we permitted comes from [1]. However, we forgot to include / as a permitted character beyond the first. It's valid (and common) for paths to have multiple / to indicate hierarchy and we should support it.

Fixes #834

[1] https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-su-create.html

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -113,7 +113,7 @@ Parameters:
     Description: AWS SSM path to the Buildkite agent registration token (this takes precedence over BuildkiteAgentToken). Expects a leading slash ('/').
     Type: String
     Default: ""
-    AllowedPattern: "^$|^/[a-zA-Z0-9_.-]+$"
+    AllowedPattern: "^$|^/[a-zA-Z0-9_.-/]+$"
 
   BuildkiteAgentTokenParameterStoreKMSKey:
     Description: AWS KMS key ID used to encrypt the SSM parameter (if encrypted)


yob:
I've tested the template from this branch with four different values in `BuildkiteAgentTokenParameterStorePath`:

1. `/foo` - passed validation
2. `/foo/bar` - passed validation
3. `foo` - failed validation
4. empty string - passed validation

nitrocode:
This is blocking us from using the new version.

LGTM 🚀

chloeruka (approved):
Nice! Let's ship a bugfix.

nitrocode:
Could we get a patch release such as v5.3.1?

vgrigoruk:
Hey @yob 
I have just tried to use the latest template from master branch (which contains [this commit](https://github.com/buildkite/elastic-ci-stack-for-aws/pull/835/files)) and it still doesn't work for us (regex validation fails). 
We store agent token under `/vendors/buildkite/agent-token` SSM path.
We ended up changing the regexp in our copy of aws-stack.yml to `"^$|^\\/[a-zA-Z0-9_\\.\\-\\/]+$"` and now the validation passes.
Not sure if this is because we deploy CloudFormation stacks via Terraform...

nitrocode:
Good catch. I tested the regex out on regexr and regexplanet but not in my stack. I'm also using the cf stack resource in terraform but I don't think this issue is specific to terraform.

The AllowedPattern uses the java.net.regex library behind the scenes and I can see other cases of it needing to use a double backslash to escape a forward slash when the regex is in quotes and single backslash if the regex is without quotes.

Do you also need to escape the period and dash characters as well for it to pass validation?

nitrocode:
For what it's worth, I just applied the `master` yaml using terraform and it worked for me and my `BuildkiteAgentTokenParameterStorePath = /buildkite/agent_token`

```
  # module.packer_queue.aws_cloudformation_stack.queue will be updated in-place
  ~ resource "aws_cloudformation_stack" "queue" {
        id               = "buildkite-packer"
        name             = "buildkite-packer"
        tags             = {
            "Name"        = "packer"
            "application" = "buildkite"
            "env"         = "testing"
            "team"        = "sre"
        }
      ~ template_url     = "https://s3.amazonaws.com/buildkite-aws-stack/v5.3.0/aws-stack.yml" -> "https://s3.amazonaws.com/buildkite-aws-stack/master/aws-stack.yml"
        # (6 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

module.packer_queue.aws_cloudformation_stack.queue: Modifying... [id=buildkite-packer]
module.packer_queue.aws_cloudformation_stack.queue: Still modifying... [id=buildkite-packer, 10s elapsed]
module.packer_queue.aws_cloudformation_stack.queue: Still modifying... [id=buildkite-packer, 20s elapsed]
module.packer_queue.aws_cloudformation_stack.queue: Still modifying... [id=buildkite-packer, 30s elapsed]
module.packer_queue.aws_cloudformation_stack.queue: Still modifying... [id=buildkite-packer, 40s elapsed]
module.packer_queue.aws_cloudformation_stack.queue: Modifications complete after 50s [id=buildkite-packer]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
Releasing state lock. This may take a few moments...
```

nitrocode:
If I use an input of `BuildkiteAgentTokenParameterStorePath = /vendors/buildkite/agent-token`, it fails with the validation issue that you're experiencing.

```
Error: error updating CloudFormation stack (buildkite-packer): ValidationError: Parameter 'BuildkiteAgentTokenParameterStorePath' must match pattern ^$|^/[a-zA-Z0-9_.-/]+$
```

nitrocode:
It looks like the dash in your input was the only character that needed to be escaped.

Correct regex expression: `"^$|^/[a-zA-Z0-9_.\\-/]+$"`

RegexPlanet: https://www.regexplanet.com/cookbook/ahJzfnJlZ2V4cGxhbmV0LWhyZHNyEwsSBlJlY2lwZRiAgIDy0PGBCgw/index.html

image

yob:
Thanks all, my bad for not including:

5. `/foo/bar/baz`

in my tests :disappointed:

yob:
We've booked to to cut a 5.3.1 bugfix release tomorrow AU time (~15 hours from now). Sorry for the wait!

yob:
[5.3.1 has been released](https://github.com/buildkite/elastic-ci-stack-for-aws/releases/tag/v5.3.1)

Issue #822, opened 2021-04-08
Use asg warm pools for faster buildkite job starts


nitrocode:
https://aws.amazon.com/about-aws/whats-new/2021/04/amazon-ec2-auto-scaling-introduces-warm-pools-accelerate-scale-out-while-saving-money/

If we could keep X instances warmed up, we could start jobs much faster without having to set the min count on the asg to something non-zero.

chloeruka:
Woah! You're quick. Yeah this feature looks great; especially for Windows instances that take a while to boot. According to the announcement it's not yet available on Cloudformation, but when it is we'll take a look into it.

nitrocode:
https://docs.aws.amazon.com/en_us/AWSCloudFormation/latest/UserGuide/aws-resource-autoscaling-warmpool.html

nitrocode:
The only param that would require an additional input to use the warm pool would be the min size of it. We use an asg min size of 0 and max size of 10 so a warm pool min size of 3 seems reasonable.

The [cloudformation docs have been released](https://docs.aws.amazon.com/en_us/AWSCloudFormation/latest/UserGuide/aws-resource-autoscaling-warmpool.html).

```yaml
Resources:
  myWarmPool: 
    Type: AWS::AutoScaling::WarmPool
    Properties:
      AutoScalingGroupName: !Ref myASG
      MinSize: 30
      PoolState: Stopped
```

theonlysinjin:
This is fantastic.
I've been using the Cloudwatch metrics to scale ASG when there are no idle agents (ie all are busy now), to help beat the agent start time.
Though it seems our agents now start (with the update) in just under 2 minutes which is pretty good.

dieend:
Warm Pool is now available in CloudFormation https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-autoscaling-warmpool.html

Right now I can't manually update the created ASG from the current template because the template configuring MixedInstancesPolicy

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/7b3d02cf2de7cdfdcb1e08b7275371529e2a4e56/templates/aws-stack.yml#L1038-L1060

We have to remove them if we'd like to use WarmPool

aiven-amartin:
Any progress/caveats on this issue? Would be really nice to have this option rather than trying to bake git checkouts/other slow workspace parts into the AMI itself.

alex-shakouri-ai:
[...]

PR #838, opened 2021-05-04
Add Warm Pool

nitrocode:
Closes https://github.com/buildkite/elastic-ci-stack-for-aws/issues/822

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -406,6 +406,11 @@ Parameters:
       - "false"
     Default: "false"
 
+  WarmPoolMinSize:
+    Type: Number
+    Description: Minimum number of instances in warm pool
+    Default: 0
+
 Outputs:
   VpcId:
     Value:
@@ -533,6 +538,9 @@ Conditions:
         - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "c6gn" ]
         - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "r6g" ]
         - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "r6gd" ]
+
+    UseWarmPool:
+      !Not [ !Equals [ !Ref WarmPoolMinSize, 0 ] ]
 
 Mappings:
   ECRManagedPolicy:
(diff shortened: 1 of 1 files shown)


nitrocode:
cc: @yob @chloeruka

keithduncan:
Hi @nitrocode, thank you for opening this pull request! I agree this could be a valuable feature to help subtract any latency associated with our EC2 `UserData` and the `BootstrapScriptUrl` from ASG scale out time.

I’ve started looking at how the warm pool will behave and interact with the rest of our stack. One concern I have is how to prevent warm pool instances from starting their `buildkite-agent` and pulling work, only for the job to be interrupted by the instance being `stopped` by the ASG. I’ve looked at the ASG events around instances moving in and out of the warm pool but haven’t been able to think up a reliable way to stall agent start up. What are your thoughts on this?

nitrocode:
I was hoping there were warm pool specific lifecycle events but there are only [instance launching and instance terminating events](https://docs.aws.amazon.com/autoscaling/ec2/userguide/lifecycle-hooks.html#lifecycle-hooks-availability) to hook into.

Perhaps there is a way to detect if an ec2 is part of the warm pool and if so skip starting the agent.

keithduncan:
Interestingly the limitations section of [Warm pools for Amazon EC2 Auto Scaling](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-warm-pools.html) calls out ECS and EKS managed node pools as having a similar issue:

> If you try using warm pools with Amazon Elastic Container Service (Amazon ECS) or Elastic Kubernetes Service (Amazon EKS) managed node groups, there is a chance that these services will schedule jobs on an instance before it reaches the warm pool.

The best idea for managing the systemd unit I’ve had so far is to receive those events in a Lambda and lean on the SSM agent to execute the state change on the host. Though I’d still be concerned about the prevalence of race conditions in that set up 🤔

Maybe the best approach here is to ask AWS for guidance on how to warm pool a workload like the `buildkite-agent` that doesn’t use a load balancer?

joshross12:
Are there plans to add this feature to the next release?

keithduncan:
Hi @joshross12 likewise this feature isn’t slated for a particular release and there are some technical hurdles to over come before we can land it.

Specifically here, my plan is to experiment with using the SSM Agent to manage the status of the buildkite-agent systemd service, whether it should be running or not, and would welcome suggestions for how that would look and how to ensure the process is reliable.

dieend:
Cross-posting this information https://github.com/buildkite/elastic-ci-stack-for-aws/issues/822#issuecomment-964356545
We have to remove MixedInstancesPolicy from ASG config to allow enable WarmPool for that ASG

PR #839, opened 2021-05-06
Max instance lifetime

nitrocode:
Related to PR https://github.com/buildkite/elastic-ci-stack-for-aws/pull/836

If this number is set to 86400 or greater, the instance will terminated.


jessebye:
@nitrocode there's a merge conflict, would you mind updating your branch?

jessebye:
@keithduncan @yob anything blocking this? Just curious if this needs changes or is good to ship.

keithduncan:
> anything blocking this?

Better support for ASG initiated instance termination is the blocker I’m afraid.

> Couldn't this still be added? The default behavior would be the same, and if a user enabled this, they are doing so at their own risk...

They could! And they are the approach I’m expecting us to use for this.

Unfortunately I’m not comfortable merging this without a reliable way to handle ASG initiated instance termination. The absence of this today already causes strife so I’m very hesitant to add _another_ source of it. For those who are comfortable with the risk, it’s possible to apply a forked copy of this template to your stacks.

I’d also be thrilled to collaborate on a Lambda +SSM based lifecycle hooks implementation for the ASG events 😄 I believe it could solve several problems at once (e.g. #927 #838) if proven to be reliable, and enable features like agent instance quarantine/cordoning.

templates/aws-stack.yml
```diff
@@ -1054,6 +1054,8 @@ Resources:
         - OldestLaunchConfiguration
         - ClosestToNextInstanceHour
       MaxInstanceLifetime: !If [UseMaxInstanceLifetime, !Ref MaxInstanceLifetime, !Ref "AWS::NoValue"]
+      NewInstancesProtectedFromScaleIn: true
```
keithduncan, line 1057: @nitrocode I suspect (but haven’t tested) adding this by merging master will break the `MaxInstanceLifetime` functionality because none of the instances are eligible for ASG initiated termination. I’m not sure if you or anyone else are running stacks from this branch’s template but wanted to call it out in case you are 😄

raylu:
I'm not sure why #964 was closed, but we'd like to see this feature merged

PR #844, opened 2021-05-11
Create secrets bucket only when needed

vgrigoruk:
We would like to avoid managed secret buckets creation in case of `EnableSecretsPlugin = false`

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -447,7 +447,9 @@ Conditions:
       !Equals [ !Ref SecurityGroupId, "" ]
 
     CreateSecretsBucket:
-      !Equals [ !Ref SecretsBucket, "" ]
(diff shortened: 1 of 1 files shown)


yob (commented):
Confirming I understand the motivation here: Currently we will create a secrets bucket if the `SecretsBucket` param is blank, and it's blank by default. All stacks that don't provide a customer `SecretsBucket` will have a new bucket created, even if they've disabled the secrets plugin.

With this change, we'll skip creating the bucket if `EnableSecretsBucket` is false - regardless of what `SecretsBucket` is set to.

vgrigoruk:
Hi @yob, exactly as you described.

yob (approved):

yob:
Thanks!

PR #845, opened 2021-05-11
Fail with a specific exit code for environment hook failures

KevinGrandon:
Right now there is no specific exit code when the elastic stack fails in a step, making it difficult to retry.This change will enable pipelines to auto-retry on infra failures.

### DIFF
--- a/packer/linux/conf/buildkite-agent/hooks/environment
+++ b/packer/linux/conf/buildkite-agent/hooks/environment
@@ -1,124 +1,131 @@
 #!/bin/bash
 set -eu -o pipefail
 
-# shellcheck source=/dev/null
-source ~/cfn-env
+setup_env() {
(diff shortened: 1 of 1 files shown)


keithduncan (changes requested):
Hi @KevinGrandon, thanks for contributing this pull request.

For myself and future reviewers I’ve found the diff to be easier to scan with whitespace ignored 😄 https://github.com/buildkite/elastic-ci-stack-for-aws/pull/845/files?diff=unified&w=1#

Just one note on the error message and this looks good to merge, thanks again!

triarius (approved):
I'm going to merge this for v6 and apply the suggestion in a follow-up.

triarius:
Hmm, the merge conflicts mean it's easier to make a complete replacement. We need to move this to the v6 branch because it's technically a breaking change.

PR #855, opened 2021-06-11
Add Secrets Manager and Multi region secret guide

### LINKED ISSUE #842: SSM agent token has to be recreated in every region
It would be nice to be able to set the ssm token once in a single region and allow buildkite queues cross region to access it.

I think if we can had a parameter that would allow specifying the region of the ssm token, it would be enough to prevent having to recreate the same ssm param in every region.

keithduncan:
Adds docs on using AWS Secrets Manager to store your Buildkite Agent token.

Fixes #842 

cc @harrietgrace @plaindocs


yob (commented):
love it :+1: 

Worth a review from @harrietgrace and/or @plaindocs as well though I reckon!

PR #858, opened 2021-06-17
Increase the HttpPutResponseHopLimit from 1 to 2

### LINKED ISSUE #857: IMDSv2Tokens="required" doesn't work with docker containers
Hi there, I set `IMDSv2Tokens` to `required` in the latest version of the stack, as we're using all the latest versions of the AWS SDK in our builds so I didn't expect any problems, but I noticed that steps we run inside docker containers that are attempting to access the CI instance's credentials now fail.

I'm pretty sure I've encountered this problem before with other tooling, and it's because the docker networking means the default hop limit of `1` won't work, as going through the docker networking stack adds an additional hop. (https://github.com/fog/fog-aws/issues/560)

As mentioned in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/789, the ability to configure the hop limit is probably required after all, as using docker containers that access AWS instance credentials I think is a not-uncommon use-case.

If you're indeed planning to make it the default in 6.0, you'll probably need to change the default hop limit to something higher, or issue some clear guidance for people who run docker containers in this way.

keithduncan:
Small change to allow Docker containers running on an instance and not using an awsvpc CNI to receive responses from the IDMSv2 service.

Fixes #857

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -851,6 +851,8 @@ Resources:
           InstanceType: !Select [ "0", !Split [ ",", !Join [ ",", [ !Ref InstanceType, "", "", "" ] ] ] ]
           MetadataOptions:
             HttpTokens: !Ref IMDSv2Tokens
+            # Allow containers using a Docker network on the host to receive IDMSv2 responses
+            HttpPutResponseHopLimit: 2
           ImageId: !If
             - HasImageId
             - !Ref ImageId


yob:
Purely out of curiosity: what security issue is the hop count limit intended to mitigate against?

keithduncan:
There are some details in the release post https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/

It’s for defending against poorly configured IP layer forwarding/proxying like NAT or otherwise. Setting it to 2 would theoretically allow such software on the elastic ci stack AMI to again exfiltrate the session from the instance, though in practice I’d hope (😅) that kind of software isn’t getting installed on those hosts. Another scenario, though still unlikely in the scheme of things, is a vulnerability in the instance’s IP stack that can be exploited to act as an IP proxy.

In that case you’d have to already be in a privileged network position to benefit from a TTL of 2 vs 1. The packets will pass through additional Amazon hardware on their way out of the VPC which _should_ decrement and drop the packets before they land somewhere useful.

yob (approved):

PR #859, opened 2021-06-17
Drop the aws prefix from the default cost tag prefix

### LINKED ISSUE #856: The tag key aws:createdBy is not valid in cost allocation tags
Tried to deploy v.5.3.2 with Cost allocation tags on and got this CF error
cost allocation tag settings default except
```
  {
    "ParameterKey": "EnableCostAllocationTags",
    "ParameterValue": "true"
  },
```
```
The tag key ?aws:createdBy? is not valid. Tag keys starting with 'aws:' are reserved for internal use. (Service: AmazonEC2; Status Code: 400; Error Code: InvalidTagKey.Malformed; Request ID: 8686fdf5-82c0-4b0e-96ab-c7b44ad3f2db; Proxy: null)
```
image

keithduncan:
See https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/allocation-tag-restrictions.html for docs on the restrictions, you cannot create a tag starting with `aws:`. I’ll create a stack with this to see if you’re supposed to add the `user:` prefix or if you can cost allocate over any tag name and the `user:` part is just presentation in the cost allocation dashboard / reports.

Fixes #856

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -392,7 +392,7 @@ Parameters:
   CostAllocationTagName:
     Type: String
     Description: The name of the Cost Allocation Tag used for billing purposes
-    Default: "aws:createdBy"
+    Default: "user:createdBy"
 
   CostAllocationTagValue:
     Type: String
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -392,7 +392,7 @@ Parameters:
   CostAllocationTagName:
     Type: String
     Description: The name of the Cost Allocation Tag used for billing purposes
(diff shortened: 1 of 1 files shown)


yob (approved):
I have not booted a stack with this change to confirm it works, but this seems legit.

Are you still planning to test it to see if the user prefix is required?

keithduncan:
Yep I still want to boot it to make sure the default works this time 😄

keithduncan:
Have tested a branch deploy of the CloudFormation template and turning it on without changing the default value worked:

Screenshot 2021-06-24 at 18 04 06

PR #863, opened 2021-06-22
Clarify which user the SSH access applies to

Temikus:
This took some time spelunking around your code to figure out which user the access applies to so I decided to clarify this in the params to simplify things for people who may run into the same issue.


keithduncan:
This is great @tekumara, thanks for putting this change together 😄

PR #865, opened 2021-06-28
Add support policy docs

keithduncan:
While the template and AMIs are entangled I think this is probably the most sensible support policy we can come up with.

I think there may be a future in making the AMI builds more readily forkable, perhaps with EC2 Image Builder, for folks to track software components differently to how we do in the main release series.

Down the road I’d also like to add a Windows Server 2019 arm64 build to complete our ARM line up at the Elastic CI Stack for AWS layer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,8 @@ We aim to support new regions within one month of general availability.
 
 We build and deploy the following AMIs to all our supported regions:
 
-- Amazon Linux 2 LTS (x86_64)
-- Amazon Linux 2 LTS (arm64)
+- Amazon Linux 2 (x86_64)
+- Amazon Linux 2 (arm64)
 - Windows Server 2019 (x86_64)
 
 ## Questions and support
--- a/README.md
+++ b/README.md
@@ -77,15 +77,13 @@ We support all AWS Regions, except China and US GovCloud.
 
 We aim to support new regions within one month of general availability.
 
-### Included Software
+### Operating Systems
 
 We build and deploy the following AMIs to all our supported regions:
 
 - Amazon Linux 2 LTS (x86_64)
 - Amazon Linux 2 LTS (arm64)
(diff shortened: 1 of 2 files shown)


yob (approved):

keithduncan:
@yob I’ve added some blurb about agent versions and shipping diffs EOM, let me know what you think before we merge 😄

README.md
```diff
@@ -67,6 +67,36 @@ If you need to build your own AMI (because you've changed something in the `pack
...
+We build and deploy the following AMIs to all our supported regions:
+
+- Amazon Linux 2 (x86_64)
+- Amazon Linux 2 (arm64)
+- Windows Server 2019 (x86_64)
+
+### Buildkite Agent
+
+The Elastic CI Stack template [published from the master branch](https://s3.amazonaws.com/buildkite-aws-stack/latest/aws-stack.yml)
+tracks the latest Buildkite Agent release.
```
yob, line 94: the only caveat with this is that the master branch template only works in us-east-1

yob (approved):

README.md
keithduncan: I looked into this and the [latest master build](https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/2389#0cecf4c3-ebfb-4cfb-bfc0-bdde6b8ccad3) did copy them. I checked the [copy step](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/.buildkite/steps/copy.sh#L109) and it does the copy for `master` or tagged builds 🥳

So branch builds don’t have their AMI copied but master and tags do, that’s what I would have done if it didn’t already work that way 😄

keithduncan: Just tested deploying master to ap-southeast-2 too and it worked, hooray!

yob: oh, neat! Sorry for the bad advice 😳

Issue #765, opened 2020-11-09
Install the docker buildx plugin with a builder to build arm containers


nitrocode:
In order to start building containers for amd64 and arm (and additional platforms), it would be nice to pre-install the [buildx](https://github.com/docker/buildx) binary.

#### Install buildx

```
# install docker buildx globally
DOCKER_DIR=/usr/libexec/docker
## get latest version or pin it to v0.4.2
BUILDX_VERSION=$(curl --silent "https://api.github.com/repos/docker/buildx/releases/latest" | jq -r '.tag_name')
mkdir -p $DOCKER_DIR/cli-plugins

## check architecture
UNAME_ARCH=`uname -m`
case $UNAME_ARCH in
  aarch64)
    BUILDX_ARCH="arm64";
    ;;
  amd64)
    BUILDX_ARCH="amd64";
    ;;
  *)
    BUILDX_ARCH="amd64";
    ;;
esac

wget \
  -O $DOCKER_DIR/cli-plugins/docker-buildx \
  -nv https://github.com/docker/buildx/releases/download/$BUILDX_VERSION/buildx-$BUILDX_VERSION.linux-$BUILDX_ARCH
chmod a+x $DOCKER_DIR/cli-plugins/docker-buildx

## Install in the buildkite-agent user
# install arm
runuser -l buildkite-agent -c 'docker run --privileged --rm tonistiigi/binfmt --install all'
# create docker builder
runuser -l buildkite-agent -c 'docker buildx create --platform linux/amd64,linux/arm64,linux/arm/v7 --name mybuilder --use'
runuser -l buildkite-agent -c 'docker buildx inspect --bootstrap'
```

Sanity check: SSH into buildkite ec2 instance

```
$ sudo su - buildkite-agent
$ docker buildx version
github.com/docker/buildx v0.4.2 fb7b670b764764dc4716df3eba07ffdae4cc47b2
$ docker buildx ls
NAME/NODE    DRIVER/ENDPOINT             STATUS  PLATFORMS
mybuilder *  docker-container
  mybuilder0 unix:///var/run/docker.sock running linux/amd64*, linux/arm64*, linux/arm/v7*, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v6
default      docker
  default    default                     running linux/amd64, linux/arm64, linux/riscv64, linux/ppc64le, linux/s390x, linux/386, linux/arm/v7, linux/arm/v6
```

[...]

PR #870, opened 2021-06-30
Install Docker using CentOS yum repositories

keithduncan:
Replace the manual Docker binary install with a yum based install.

This requires adding the CentOS 7 repositories so that the Docker repository package deps can be satisfied, and is the part I’m least confident in relying on 😬

I’ve tested these steps on a t3.large and t4g.large to cover both x86_64 and aarch64 and had it work in both places.

Fixes https://github.com/buildkite/elastic-ci-stack-for-aws/issues/765

### DIFF
--- a/packer/linux/scripts/install-docker.sh
+++ b/packer/linux/scripts/install-docker.sh
@@ -35,17 +35,15 @@ gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
 priority=1
 EOF
 
-# Download the signing key for the repository
-if [ "${MACHINE}" = "x86_64" ]
-then
-  centos_key="https://centos.org/keys/RPM-GPG-KEY-CentOS-7"
-elif [ "${MACHINE}" = "aarch64" ]
-then
-  centos_key="https://centos.org/keys/RPM-GPG-KEY-CentOS-7-aarch64"
-fi
+import_rpm_key() {
+  local key="${1}"
+  local url="${2}"
+  sudo curl --fail --location --output "/etc/pki/rpm-gpg/${key}" "${url}"
+  sudo rpm -import "/etc/pki/rpm-gpg/${key}"
+}
 
-sudo curl --fail --location --output /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7 "${centos_key}"
-sudo rpm -import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
+import_rpm_key "RPM-GPG-KEY-CentOS-7" "https://centos.org/keys/RPM-GPG-KEY-CentOS-7"
+import_rpm_key "RPM-GPG-KEY-CentOS-7-aarch64" "https://centos.org/keys/RPM-GPG-KEY-CentOS-7-aarch64"
 
 # Do the install
 sudo yum install -y -q "docker-ce-${DOCKER_VERSION}" "docker-ce-cli-${DOCKER_VERSION}" containerd.io
--- a/packer/linux/scripts/install-docker.sh
+++ b/packer/linux/scripts/install-docker.sh
@@ -36,7 +36,15 @@ priority=1
 EOF
 
 # Download the signing key for the repository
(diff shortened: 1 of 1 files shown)


packer/linux/scripts/install-docker.sh
```diff
@@ -6,28 +6,46 @@ DOCKER_RELEASE="stable"
...
 
-# Manual install ala https://docs.docker.com/engine/installation/binaries/
-curl -Lsf -o docker.tgz "https://download.docker.com/linux/static/${DOCKER_RELEASE}/${MACHINE}/docker-${DOCKER_VERSION}.tgz"
-tar -xvzf docker.tgz
-sudo mv docker/* /usr/bin
-rm docker.tgz
+# Add the Docker yum repository, but override the url to get CentOS 7 not
+# CentOS 2 (the default with Amazon Linux 2 applied to the interpolation)
+sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+sudo yum-config-manager --setopt="docker-ce-stable.baseurl=https://download.docker.com/linux/centos/7/$MACHINE/stable" --save
```
keithduncan, line 16: The default interpolation of this repo file gives us `CentOS 2` because of Amazon Linux 2, this fixes up the URL to work

packer/linux/scripts/install-docker.sh
```diff
@@ -6,28 +6,46 @@ DOCKER_RELEASE="stable"
...
+
+# Add CentOS 7 Base and Extras to the yum repos list, needed for deps for
+# packages from the Docker yum repo
+cat << 'EOF' | sudo tee /etc/yum.repos.d/centos.repo
+[base]
+name=CentOS-7 - Base
+mirrorlist=http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=os
```
keithduncan, line 23: Could interpolate our `$MACHINE` var instead of placing the `$basearch` literal in here, don’t feel strongly either way as the AMI isn’t likely to be rolled forward across releases or arches like a long lived install would be 🤷🏼‍♂️

packer/linux/scripts/install-docker.sh
```diff
@@ -6,28 +6,46 @@ DOCKER_RELEASE="stable"
...
+mirrorlist=http://mirrorlist.centos.org/?release=7&arch=$basearch&repo=extras
+enabled=1
+gpgcheck=1
+priority=1
+EOF
+
+import_rpm_key() {
+  local key="${1}"
+  local url="${2}"
+  sudo curl --fail --location --output "/etc/pki/rpm-gpg/${key}" "${url}"
```
keithduncan, line 39: TODO delete this file after import

packer/linux/scripts/install-docker.sh
```diff
@@ -6,28 +6,46 @@ DOCKER_RELEASE="stable"
...
+import_rpm_key "RPM-GPG-KEY-CentOS-7" "https://centos.org/keys/RPM-GPG-KEY-CentOS-7"
+import_rpm_key "RPM-GPG-KEY-CentOS-7-aarch64" "https://centos.org/keys/RPM-GPG-KEY-CentOS-7-aarch64"
 
-sudo mkdir -p /etc/docker
-sudo cp /tmp/conf/docker/daemon.json /etc/docker/daemon.json
-sudo cp /tmp/conf/docker/subuid /etc/subuid
-sudo cp /tmp/conf/docker/subgid /etc/subgid
-sudo chown -R ec2-user:docker /etc/docker
+# Do the install
+sudo yum install -y -q "docker-ce-${DOCKER_VERSION}" "docker-ce-cli-${DOCKER_VERSION}" containerd.io
```
keithduncan, line 47: This is lifted from the Docker install docs with the version suffixes added, probably don’t need the literal call out to `containerd.io` which is likely a dep. Can test again without it.

keithduncan:
Tests are showing subsequent `yum install` commands are failing with these repos added. Looking into that now.

sj26:
Looks like a can of worms, sorry!

keithduncan:
It sadly does 😢 it would have been a preferable way to land updates plus having upstream package requirements codified too but alas, not yet! 🙁

yob:
now we know!

PR #873, opened 2021-07-05
Add cross-parameter rules

### LINKED ISSUE #648: Buildkite agent fails with default template options
Using the default parameters/template using the `Launch Stack` button, buildkite agent immediately fails with `ROLLBACK`

keithduncan:
Adds a [`Rules`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/rules-section-structure.html) section to validate cross-parameter values:

- One of BuildkiteAgentToken or BuildkiteAgentTokenParameterStorePath must be specified

Fixes #648

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -407,6 +407,21 @@ Parameters:
       - "false"
     Default: "false"
 
+Rules:
+  HasToken:
+    Assertions:
+      - Assert:
+          !Or
+            - !Not
+              - !Equals
+                - !Ref BuildkiteAgentToken
+                - ""
+            - !Not
+              - !Equals
+                - !Ref BuildkiteAgentTokenParameterStorePath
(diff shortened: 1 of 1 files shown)


keithduncan:
Sadly I couldn’t get validation of the VPC Parameters working because the stack parameters are optional and thus untyped, disabling the use of the `!MemberOf` functions.

Also sadly the validation still happens quite late in the template apply process, I had hoped it would be run at the parameter input stage before progressing.

Screenshot 2021-07-09 at 13 43 13

The benefit of this change is it now tells you what you are missing since we can’t make either of the token parameters required while both are supported.

yob (approved):
Neat!

it's a shame the VPC one wasn't possible, but now that we have this initial check in place I'm sure we'll come up with other rules that will be useful :+1:

Issue #891, opened 2021-08-18
Cross-region deploy made difficult when using Elastic CI stack


toothbrush:
Hey folks, we've just adopted this Elastic CI stack instead of our legacy/hand-built one, and it mostly is working well!

We ran into a surprising issue though, when attempting to do a cross-region deploy.  The situation is that we rely on `AWS_DEFAULT_REGION` to point deploy commands at the right target (that's probably not the best, but bear with me, legacy decisions that are now in many repos all over the place 🤷).

One of our pipelines looks roughly as follows – commands such as `aws` inside the Makefile rely on `AWS_DEFAULT_REGION` pointing where we want to deploy things:

```yaml
steps:
  - name: us-east-1
    command: make deploy-app
    env:
      AWS_DEFAULT_REGION: 'us-east-1'
      KUBERNETES_DEPLOY_TARGET: 'https://kubernetes-production.us-east-1.our.domain'
    agents:
      queue: deploy-prod

  - name: us-west-2
    command: make deploy-app
    env:
      AWS_DEFAULT_REGION: 'us-west-2'
      KUBERNETES_DEPLOY_TARGET: 'https://kubernetes-production.us-west-2.our.domain'
    agents:
      queue: deploy-prod
```

However, when the steps run i see that in both cases, `AWS_DEFAULT_REGION == us-east-1` – that's where our Buildkite agents run.

I suspect this code is running and somehow overriding the `env:` block of our `pipeline.yml`, but there must surely be a way to allow overriding some of these variables?  

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/368a03a14b51a44a87af733a96cc193c7b10a3eb/templates/aws-stack.yml#L981-L985

I've currently made a workaround for one of our repositories, but this feels like quite a footgun – if i set a value in the pipeline's `env:` block, i expect it to be respected in the executed steps.

cc: @novatan-rb @jonhiggs

toothbrush:
I've just double-checked using a small repro:

```yaml
steps:
  - name: us-east-1
    command: env
    env:
      TEST_FOO: bar
      AWS_DEFAULT_REGION: us-east-1

  - name: us-west-2
    command: env
    env:
      TEST_FOO: baz
      AWS_DEFAULT_REGION: us-west-2
```

Concentrating on the `us-west-2` step, if i look at the "Environment" tab of the step, it looks as expected:

[...]

PR #892, opened 2021-08-20
Don't clobber user's env vars.

toothbrush:
This change makes a start at addressing the issue raised in https://github.com/buildkite/elastic-ci-stack-for-aws/issues/891 - where a user sets an environment variable in their pipeline config but it unexpectedly gets clobbered by Buildkite's defaults, which are determined at AMI-build-time.

We introduce a helper function, `set_unless_present`, which will raise a warning if a variable it wants to set already exists, and leave it alone.


toothbrush:
I hoped to write this stuff in a more POSIX-compliant way, but it seemed as if the cleanest way was to use a number of Bash-isms instead (`if [[ -v ...`, `declare`, and in particular `${!var}`).  It seems this script is only intended to run in Bash, but i wanted to flag that.

packer/linux/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -37,16 +37,46 @@ PLUGINS_ENABLED=()
...
+	# using an env: block in their pipeline.yml.  This little helper is sets the environment variables
+	# buildkite-agent and plugins expect, except if a user want to override them, for example to do a
+	# deployment to a region other than where the Buildkite agent lives.
+	function set_unless_present() {
+	    local target=$1
+	    local value=$2
+
+	    if [[ -v "${target}" ]]; then
+	        echo "^^^ +++"
+	        echo "⚠️ ${target} already set, NOT overriding! (current value: \"${!target}\")"
```
toothbrush, line 59:
```suggestion
	        echo "⚠️ ${target} modified by user, NOT modifying! (current value: \"${!target}\")"
```

I'm not 100% sure what's clearest here.  Perhaps you folks can shed light on what would be most obvious to the user.

keithduncan: I think I like the "set" and "overriding" language here.

I would like to include _where_ it has already been set from. We can’t deduce where it’s from with 100% certainty because by the time the bootstrap and by extension the environment hook runs it could be inherited from the agent’s process environment or set by the Buildkite Step / Job environment as in your case 🤔 Maybe we just need to call out both:

```suggestion
	        echo "⚠️ ${target} already set, NOT overriding! (current value \"${!target}\", set by Buildkite Step env configuration or inherited from the buildkite-agent process environment)"
```

packer/linux/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -37,16 +37,46 @@ PLUGINS_ENABLED=()
...
+	}
+EOF
+
+# Step 2: Populate the default variable values.  This time, we append to the file, and allow
+# variable substitution.
+cat << EOF >> /var/lib/buildkite-agent/cfn-env
+
+set_unless_present "AWS_DEFAULT_REGION" "$AWS_REGION"
+set_unless_present "AWS_REGION" "$AWS_REGION"
+set_unless_present "BUILDKITE_AGENTS_PER_INSTANCE" "$BUILDKITE_AGENTS_PER_INSTANCE"
```
keithduncan, line 73: I think we should only apply the `set_unless_present` function to the AWS environment variables above ☝️ 

These others are more a description of the stack you find yourself in rather than a configurable option that changes the behaviour of your job, e.g. setting `BUILDKITE_AGENTS_PER_INSTANCE` doesn’t spawn more agents, setting `DOCKER_VERSION` doesn’t route the Docker CLI commands to a second install etc

What do you think? 😄

keithduncan:
Thank you so much for this pull request @toothbrush! I’m excited to get this merged for you 😄

packer/linux/conf/bin/bk-install-elastic-stack.sh
toothbrush: Yes!  Good point.  I was just mechanistically replacing-all, without thinking much about the role of the various variables.  I'll update as you suggest.

packer/linux/conf/bin/bk-install-elastic-stack.sh
toothbrush: Yeah, the uncertainty of "where this value comes from" is why i wanted input 👍 Will update the copy!

packer/linux/conf/bin/bk-install-elastic-stack.sh
toothbrush: Would you say the same for `PLUGINS_ENABLED` and `BUILDKITE_ECR_POLICY`, they're both more a statement of fact than a tweakable knob?

packer/windows/conf/bin/bk-install-elastic-stack.ps1
```diff
@@ -36,16 +36,50 @@ If ($Env:ECR_PLUGIN_ENABLED -eq "true") { $PLUGINS_ENABLED += "ecr" }
...
+    local value=$2
+
+    echo "export ${target}=\"${value}\""
+    declare -gx "${target}=${value}"
+}
+'@
+
+Add-Content -Path C:\buildkite-agent\cfn-env -Value @"
+
+set_always         "BUILDKITE_AGENTS_PER_INSTANCE" "$Env:BUILDKITE_AGENTS_PER_INSTANCE"
```
toothbrush, line 74: Right, i've reversed the behaviour again, and now i'm clobbering all variables _except_ `AWS_*`, as you suggested.

Initially i just went back to the original `export A=B` style, but it occurred to me that now, there's a regression: folks will stop seeing that particular variables have been set, because i removed the `cat ~/cfn-env` call... I thought about reinstating that, but figured a small helper function that simply echoes and sets the variable couldn't hurt...

If you think this is too much crazy complexity and magic, i'm happy to remove that.  I do like the way the code reads exactly like a human would explain what's happening..?

toothbrush: The other worry i had was alphabetisation.. 😅 I guess it doesn't matter whether i sort these lines as they are, or by target variable.  I should probably chill out.

keithduncan: > If you think this is too much crazy complexity and magic, i'm happy to remove that. I do like the way the code reads exactly like a human would explain what's happening..?

I love it! Nicely self documenting the expected behaviour of each variable 🙌

keithduncan (approved):

keithduncan:
Thank you for those changes, I’m going to do some branch deploy testing of my own before merging this but should get it into `master` shortly and into the end of month August tag release :bow:

toothbrush:
> Thank you for those changes, I’m going to do some branch deploy testing of my own before merging this but should get it into `master` shortly and into the end of month August tag release 🙇

Awesome, thanks so much for the super quick responses and helping me out with my niche issue! 🙌

keithduncan:
I’ve done some sight verification on how this will change the embedded [plugins](https://github.com/buildkite/elastic-ci-stack-for-aws/tree/master/plugins).

The [secrets](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/368a03a14b51a44a87af733a96cc193c7b10a3eb/packer/linux/conf/buildkite-agent/hooks/environment#L64-L69) plug-in uses the `AWS_DEFAULT_REGION` [variable](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/blob/922b375a124b6620bdac362341b107dc56bdb8f8/s3secrets-helper/s3/s3.go#L30-L39) to [discover](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/blob/922b375a124b6620bdac362341b107dc56bdb8f8/s3secrets-helper/s3/s3.go#L43-L47) which region hosts the secrets bucket. This is a multi-region fix I applied recently. This changeset will change which region is used to discover the bucket region, which I think is fine.

The [ECR Login](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/368a03a14b51a44a87af733a96cc193c7b10a3eb/packer/linux/conf/buildkite-agent/hooks/environment#L71-L81) plug-in also uses the `AWS_DEFAULT_REGION` variable if the more specific plug-in level variable [isn’t set](https://github.com/buildkite-plugins/ecr-buildkite-plugin/blob/838688cee14e8db5b1d1eaaab4316a40b757745a/hooks/environment#L128). With this changeset applied, if you set the `AWS_DEFAULT_REGION` variable in your job env that will change which regional registry is signed in to, where before it would have been the in-region registry. That is reasonably significant, though as before it’s a behaviour you opt-in to by setting the variable in your job env. Additionally, both now and with this change applied the registry region can be configured by setting `BUILDKITE_PLUGIN_ECR_REGION` either in your job env or by appending it to `~/cfn-env` in your bootstrap script.

The docker login plugin appears unaffected.

PR #903, opened 2021-09-03
Install qemu and binfmt hooks

### LINKED ISSUE #799: Multi-arch support
ARM support is becoming available almost everywhere, but the almost is taking a while. It's possible to make adoption easier by adding transparent multi-arch support, so that running foreign binaries on linux works transparently. There's a great guide for how to do this with qemu, and it seems to be what tools like Docker for Mac are doing under the hood:

https://www.stereolabs.com/docs/docker/building-arm-container-on-x86/

These steps require a reboot, which makes them impractical to run in a bootstrap script. Should we ship qemu and binfmt support within the elastic stack? This would be great for the arm64 images at least, where this problem is most painful.

The instructions are, roughly:

```
yum install qemu-user qemu-*-static qemu-kvm-tools qemu-kvm virt-install libvirt libvirt-python libguestfs-tools-c qemu-system-aarch64 glib2-devel pixman-devel -y
wget https://raw.githubusercontent.com/qemu/qemu/master/scripts/qemu-binfmt-conf.sh
wget https://raw.githubusercontent.com/multiarch/qemu-user-static/master/containers/latest/register.sh
chmod +x qemu-binfmt-conf.sh register.sh
sed -i "s/ \\/qemu-binfmt-conf\\.sh/ .\\/qemu-binfmt-conf.sh/" register.sh
```

There's some supporting documentation here:

https://github.com/multiarch/qemu-user-static

keithduncan:
Allow `docker buildx` to run emulation so we can build arm64 images on amd64 machines.

Fixes #799

### DIFF
--- a/packer/linux/scripts/install-docker.sh
+++ b/packer/linux/scripts/install-docker.sh
@@ -82,3 +82,10 @@ esac
 sudo curl --location --fail --silent --output "${DOCKER_CLI_DIR}/docker-buildx" "https://github.com/docker/buildx/releases/download/v${DOCKER_BUILDX_VERSION}/buildx-v${DOCKER_BUILDX_VERSION}.linux-${BUILDX_ARCH}"
 sudo chmod +x "${DOCKER_CLI_DIR}/docker-buildx"
 docker buildx version
+
+echo "Installing qemu..."
+sudo yum install -y qemu qemu-user-static
+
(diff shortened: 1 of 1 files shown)


keithduncan:
This appears to be working nicely, the use of it in https://github.com/buildkite/agent/pull/1502 appears trickier than first expected though.

PR #911, opened 2021-09-08
Enable experimental docker cli in the environment hook

### LINKED ISSUE #798: Bug: Have to add DOCKER_CLI_EXPERIMENTAL to get docker manifests pushed

keithduncan:
Save the value of the `EnableDockerExperimental` parameter in `~/cfn-env` for the environment hook to use. In the environment hook, set `$DOCKER_CONFIG/config.json` `.experimental="enabled"`.

Fixes #798


packer/linux/conf/buildkite-agent/hooks/environment
```diff
@@ -10,6 +10,11 @@ BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY=$(mktemp -d)
 export BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY
 export DOCKER_CONFIG="$BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY"
 
+if [ "${BUILDKITE_DOCKER_EXPERIMENTAL:-false}" = "true" ]
+then
+  cat <<< "$(jq '.experimental="enabled"' "${DOCKER_CONFIG}/config.json")" > "${DOCKER_CONFIG}/config.json"
```
keithduncan, line 15: This pattern is used [elsewhere](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/95d49a50a3d80f24e55eb758281602dc8ef3c418/packer/linux/conf/bin/bk-configure-docker.sh#L13) in the codebase without issue.

Knowing about the truncation issue I was curious whether this would work when writing it. I’ve confirmed that the multiple edits in the above link are preserved, that the former edit isn’t truncated to zero and overwritten by the latter edit.

I’d like to confirm why this is working before moving ahead with a `#shellcheck disable=SC2094`

keithduncan: This pattern seems to come from this long standing issue `https://github.com/stedolan/jq/issues/105#issuecomment-437363987` in jq asking for edit in place.

keithduncan: I reckon this must be fine because it is redirects in a single command which bash processes in order rather than redirects in separate commands in a pipeline which are processed in parallel and race.

PR #913, opened 2021-09-09
Add capabilities and supporting policies for Cloudformation macro

albertywu:
The latest aws-stack.yml uses a macro that requires:

1) `CAPABILITY_AUTO_EXPAND` capability when creating the Cloudformation stack
2) policies that grant `CreateCloudFormationTemplate` and `GetCloudFormationTemplate` to the EC2 instance role

This PR adds documentation for 1), and an inline policy for 2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ aws cloudformation create-stack \
   --output text \
   --stack-name buildkite \
   --template-url "https://s3.amazonaws.com/buildkite-aws-stack/latest/aws-stack.yml" \
-  --capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM \
+  --capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM CAPABILITY_AUTO_EXPAND \
   --parameters "$(cat config.json)"
 ```
 
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -716,6 +716,16 @@ Resources:
                   !Sub
                     - arn:aws:ssm:${AWS::Region}:${AWS::AccountId}:parameter${ParameterPath}
                     - ParameterPath: !If [ UseCustomerManagedParameterPath, !Ref BuildkiteAgentTokenParameterStorePath, !Ref BuildkiteAgentTokenParameter ]
+        - PolicyName: ServerlessCloudformation
+          PolicyDocument:
+            Version: '2021-09-08'
+            Statement:
+              - Effect: Allow
+                Action: serverlessrepo:CreateCloudFormationTemplate
+                Resource: arn:aws:serverlessrepo:us-east-1:172840064832:applications/buildkite-agent-scaler
+              - Effect: Allow
+                Action: serverlessrepo:GetCloudFormationTemplate
+                Resource: arn:aws:serverlessrepo:us-east-1:172840064832:applications/buildkite-agent-scaler      
       AssumeRolePolicyDocument:
         Statement:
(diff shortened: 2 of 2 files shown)


keithduncan:
Hi @albertywu thanks for opening this pull request, and I’m sorry you were bitten by this change 😅  

I agree we should update the readme to include this capability :+1: but it looks like the policy adding `serverlessrepo:CreateCloudFormationTemplate` and `serverlessrepo:GetCloudFormationTemplate` is because you are using an Elastic CI Stack to deploy more Elastic CI Stacks?

I don’t think this part is generally applicable to all customers and that you might be better served by instead adding this policy to _your_ instance roles with the [ManagedPolicyARN](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/2ba56042415187bee363215f5253bc13f1150e29/templates/aws-stack.yml#L297-L300) template parameter. What do you think?

albertywu:
> I don’t think this part is generally applicable to all customers

True, but just be aware that the cloudformation change that introduced macros is a breaking change for users who deploy cloudformation stacks from EC2 instance (Uber, and likely others). It wasn't obvious how to fix the issue either, and we spent a good deal of time figuring out which policies we needed to attach and how to attach them.

> you might be better served by instead adding this policy to your instance roles with the ManagedPolicyARN template parameter. What do you think?

This is exactly what we did at Uber. We're unblocked and have a workaround, but other customers of yours will likely run into the same issue.

---

As a side note, curious why deploying from an EC2 instance requires additional policies to be set, but not when deploying from the AWS CLI. Is this a hole in AWS's security model?

albertywu:
In case you're wondering why we're deploying Cloudformation from EC2 instance instead of a 1-time deploy from CLI (followed by editing the cloudformation template inside the AWS console) -- at Uber, we manage multiple Cloudformation stacks using terraform and the terraform cloudformation [resource](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudformation_stack).

We do this so any config changes to Cloudformation are version-controlled in our terraform configuration. Changes to Cloudformation stacks are deployed on EC2 instances via a Buildkite pipeline (which runs on an EC2 instance).

Hope that makes sense, and helps to clarify why we ran into this issue.

keithduncan:
Thank you for opening this pull request @albertywu :bow: I deeply appreciate the time you took to thoroughly research and write up your findings, and to share them with us.

I’m going to close this in favour of https://github.com/buildkite/elastic-ci-stack-for-aws/pull/926 which instead documents the actions that our template requires in the form of an IAM policy, and steers folks towards creating a CloudFormation service role with those actions included. I have cherry-picked https://github.com/buildkite/elastic-ci-stack-for-aws/pull/913/commits/f2005a737acbe42f9cf9295fe83ec0d8e93fc7b4 into my branch which adds the `CAPABILITY_AUTO_EXPAND` to the docs.

I apologise again for releasing what ended up being a breaking change for you as a SemVer minor update. Up to now I have used "do previously authored pipelines continue to run" as the yard stick for measuring whether a change is backwards compatible. We are currently discussing whether to add changes in CloudFormation capabilities and required IAM actions to our notion of backwards compatibility for future versions.

keithduncan:
I’ve merged #926 that documents the required capability on the README. I’ve also documented all the required IAM actions in the form of an experimental-grade CloudFormation service role that is used to create and delete stacks in our test suite.

Future changes like this should now be caught and surfaced for inclusion in the release notes 🙇

PR #915, opened 2021-09-09
Quote additional sudo permissions

christophermaier:
Commands specified in a sudoers file can be simple commands, or they
can be commands and arguments. Thus,
`BUILDKITE_ADDITIONAL_SUDO_PERMISSIONS` can be expected to contain
spaces, and should be quoted.

Without quotes, anything after the space is taken to be a separate
command to be invoked, which will often give you a non-functional
instance (at least as far as interacting with Buildkite is concerned).

A workaround is to manually escape any spaces.

Signed-off-by: Christopher Maier <chris@graplsecurity.com>


christophermaier:
Withdrawing in favor of my coworker's patch in #916

PR #920, opened 2021-09-14
fix: support MaxInstanceLifetime for rotating out old agents

jessebye:
Adds support for configuring `MaxInstanceLifetime` on the ASG. See also:
- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-maxinstancelifetime
- https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-max-instance-lifetime.html

We would like to use this to ensure no Buildkite agent lives for longer than 24h. This helps prevent some issues we've experienced with long-lived agents, such as disks filling up.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -234,6 +234,11 @@ Parameters:
     Type: Number
     Default: 0
 
+  MaxInstanceLifetime:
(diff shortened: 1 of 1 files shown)


jessebye:
Oops, didn't see that this PR already existed: https://github.com/buildkite/elastic-ci-stack-for-aws/pull/839

PR #921, opened 2021-09-15
add base ami choice to packer json and Makefiles

stvhawes:


Makefile
```diff
@@ -82,13 +82,17 @@ packer-linux-amd64.output: $(PACKER_LINUX_FILES)
...
+		-e SOURCE_AMI_NAME_FILTER \
+		-e SOURCE_AMI_OWNER \
 		-e PACKER_LOG \
 		-v ${HOME}/.aws:/root/.aws \
 		-v "$(PWD):/src" \
 		--rm \
 		-w /src/packer/linux \
 		hashicorp/packer:$(PACKER_VERSION) build -timestamp-ui -var 'region=$(AWS_REGION)' \
 			-var 'arch=x86_64' -var 'goarch=amd64' -var 'instance_type=$(AMD64_INSTANCE_TYPE)' \
+			-var 'source_ami_name_filter=$(SOURCE_AMI_NAME_FILTER)' \
```
keithduncan, line 94: Does this override the default value from the `variables` section with an empty string if the environment variable isn’t set?

keithduncan (commented):
Thank you for opening this pull request @stvhawes!

Customising the stack’s source AMI for the Packer build isn’t a customisation I had considered but I think it’s a nice feature I’m keen to support.

I’ve got one question ☝️ on the ergonomics of using this to ensure existing automation that invokes `make packer` continues to work but otherwise I support it in principle.

PR #922, opened 2021-09-17
Revert "Remove lifecycled"

keithduncan:
Reverts buildkite/elastic-ci-stack-for-aws#908 This is still used and was removed in error, noticed by @pzeballos :bow:

The poll-for-spot-termination handler in lifecycled is still used for the time being, I mistakenly removed this following the SQS and SNS method clean-up.

This may let be re-removed following experimentation with Lambda based approaches to better support ASG initiated termination.

### DIFF
--- a/packer/linux/scripts/install-lifecycled.sh
+++ b/packer/linux/scripts/install-lifecycled.sh
@@ -15,8 +15,8 @@ echo "Installing lifecycled ${LIFECYCLED_VERSION}..."
 
 sudo touch /etc/lifecycled
 sudo curl -Lf -o /usr/bin/lifecycled \
-	https://github.com/lox/lifecycled/releases/download/${LIFECYCLED_VERSION}/lifecycled-linux-${ARCH}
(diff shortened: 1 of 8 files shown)


packer/linux/scripts/install-lifecycled.sh
```diff
@@ -0,0 +1,22 @@
...
+	x86_64)    ARCH=amd64;;
+	aarch64)   ARCH=arm64;;
+	*)         ARCH=unknown;;
+esac
+
+echo "Installing lifecycled ${LIFECYCLED_VERSION}..."
+
+sudo touch /etc/lifecycled
+sudo curl -Lf -o /usr/bin/lifecycled \
+	https://github.com/lox/lifecycled/releases/download/${LIFECYCLED_VERSION}/lifecycled-linux-${ARCH}
```
pzeballos, line 18: although this still works, I think it should point to https://github.com/buildkite/lifecycled (it redirect to this one)

keithduncan: Updated in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/922/commits/0388a19706fe7f7c4de4d3ee59a86979f2782e7c

PR #925, opened 2021-09-22
Bind mount ephemeral storage into main file system

keithduncan:
Follow up to https://github.com/buildkite/elastic-ci-stack-for-aws/pull/557 I realised pointing the buildkite-agent config for the builds location would break [fix-buildkite-agent-builds-permissions](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/e1c5f8b3965fc9b6101c48aca24e91ef778c34a4/packer/linux/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions#L77) on instances with instance storage.

Instead of parameterising the builds path, or reading it from the agent config, I decided to bind mount the ephemeral storage into the main file system. This way the paths stay the same but the underlying block device is the nvme or RAID device.

I think this is also preferable to partitioning the block device and mounting separate partitions into the `git-mirrors` and `builds` directory 🤔 this way we mount the drive(s) once in the [bk-mount-instance-storage.sh](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/packer/linux/conf/bin/bk-mount-instance-storage.sh) script and then additionally bind mount slices of that into different spots of the main file system.

Presently git-mirrors, builds, and the docker data-dir are unconditionally stored on the instance storage, but I could see this expanding to be configurable at some point.

### DIFF
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -134,14 +134,14 @@ if [[ "${BUILDKITE_AGENT_ENABLE_GIT_MIRRORS_EXPERIMENT}" == "true" ]] ; then
     BUILDKITE_AGENT_EXPERIMENTS+=",git-mirrors"
   fi
 
+  BUILDKITE_AGENT_GIT_MIRRORS_PATH="/var/lib/buildkite-agent/git-mirrors"
+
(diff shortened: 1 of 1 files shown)


keithduncan:
I’ve tested these commands on a hand assembled instance. Once this branch has built I’ll deploy a stack from the branch built template and confirm it works there before merging.

keithduncan:
Branch deploy of this template to us-east-1 worked successfully.

Issue #904, opened 2021-09-06
Document 5.6.X CloudFormation required capability


calvinbui:
I upgraded from 5.5.1 to 5.6.1 and noticed the `CAPABILITY_AUTO_EXPAND` capability is now required.

This might have to be documented somewhere.

AWS docs: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-iam-template.html#using-iam-capabilities

via console: 

![image](https://user-images.githubusercontent.com/3604363/132147180-058e1209-636a-48a0-b11f-3aaf7b9c7a42.png)

via terraform (assuming it uses the API):

```
Error: error updating CloudFormation stack (arn:aws:cloudformation:us-east-2:xxxxxxxxxxx:stack/buildkite-elastic-ci/xxxxxxxx): InsufficientCapabilitiesException: Requires capabilities : [CAPABILITY_AUTO_EXPAND]

status code: 400, request id: xxxxxxxxx
```

keithduncan:
Hi @calvinbui :wave: sorry you ran into this, this was added in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/685.

Where would be most useful to mention this? I have added a bullet point to the 5.6.0 release notes. Thankfully that error is fairly informative 😄

calvinbui:
lgtm

PR #926, opened 2021-09-23
Document IAM actions needed to create, update, and delete the template

### LINKED ISSUE #924: Need for CAPABILITY_AUTO_EXPAND is not documented
I recently upgraded to version 5.6.1, and ran into a failure because I had not granted `CAPABILITY_AUTO_EXPAND` when creating the stack.

This appears to have been added recently in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/685, but was not reflected in the changelog or the README.

(This is an issue, rather than a PR, because I'm not 100% familiar with the capability, or the need for it in this stack, or if there are other places that may need updating as well.)

keithduncan:
I gleaned these actions using the IAM Access Analyser to generate a policy for a CloudFormation service role that I granted admin. I then hand tuned the list, removed the admin policy and redeployed a stack using it.

I have scanned all the branches and tried to be thorough but I don’t have high confidence that I will have caught _all_ the possible template parameter combinations that might cause additional IAM actions to be required. I am tempted to put a `BETA` label on this in the docs section to signal that it is likely incomplete.

I’ve cherry-picked f2005a737acbe42f9cf9295fe83ec0d8e93fc7b4 from https://github.com/buildkite/elastic-ci-stack-for-aws/pull/913 which adds `AUTO_EXPAND` to the list of capabilities this requires.

---

TODO

- [x] Add integration test which consumes these policies and deploys and deletes a stack using a service role with these policies

Fixes #924 
cc https://github.com/buildkite/docs/issues/1136
cc #904


keithduncan:
I really wanted to launch a separate stack with the service role so that the step could be marked as soft fail while the new role is in testing 🤔 but the `launch.sh` wasn’t obviously parameterisable.

PR #936, opened 2021-09-30
Fix resource ordering

keithduncan:
Subsequent to https://github.com/buildkite/elastic-ci-stack-for-aws/pull/928 I found more resource ordering issues.

Similar to that change which ensures the IAM control pane continues to allow instances to `autoscaling:TerminateInstanceInAutoScalingGroup` under stack deletion, this change aims to enforce a delete ordering on the VPC resources such that the instances continue to have the _network access_ needed to instigate that terminate action. Previously there was no order dependency on the route table or route table to subnet association. Under stack deletion we could rip the networking out from under the instances again preventing them from ever being able to self terminate. This was fine so long as the instances weren’t protected from scale in because the ASG would initiate termination of them. Now that it can’t it requires a manual intervention to terminate the instance.

This change:

- Moves the conditional and standalone `AWS::IAM::Policy` resources into conditional statements of the existing `IAMPolicy` resource. There already exists a `DependsOn` relationship on this resource as of #928. Because these are conditionally created resources we cannot `DependsOn` them specifically.
- Attempts to create an unconditional `VpcComplete` resource that can be `DependsOn` by the `AutoScalingGroup` that itself establishes a conditional relationship on the leaf nodes of the VPC configuration. I’m not sure if this part is going to work yet.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -670,11 +670,11 @@ Resources:
   # a strict ordering can be established on teardown.
   VpcComplete:
     Type: AWS::CloudFormation::WaitConditionHandle
-    Condition: CreateVpcResources
-    DependsOn:
-      - RouteDefault
-      - Subnet0Routes
-      - Subnet1Routes
+    Metadata:
+      VpcResources: !If
+        - CreateVpcResources
+        - [ !Ref RouteDefault, !Ref Subnet0Routes, !Ref Subnet1Routes ]
+        - !Ref AWS::NoValue
 
   BuildkiteAgentTokenParameter:
     Type: AWS::SSM::Parameter
@@ -1004,6 +1004,7 @@ Resources:
     Type: AWS::AutoScaling::AutoScalingGroup
     DependsOn:
       - IAMPolicies
+      - VpcComplete
     Properties:
       VPCZoneIdentifier: !If [ "CreateVpcResources", [ !Ref Subnet0, !Ref Subnet1 ], !Ref Subnets ]
       MixedInstancesPolicy:
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -752,7 +752,8 @@ Resources:
         Statement:
           - !If
             - HasSecretsBucket
-            - Effect: Allow
+            - Sid: SecretsBucket
+              Effect: Allow
               Action:
                 - s3:Get*
                 - s3:List*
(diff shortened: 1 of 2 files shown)


templates/aws-stack.yml
```diff
@@ -665,6 +666,16 @@ Resources:
       SubnetId: !Ref Subnet1
       RouteTableId: !Ref Routes
 
+  # A resource that depends on the leaf nodes of the VPC configuration so that
+  # a strict ordering can be established on teardown.
+  VpcComplete:
+    Type: AWS::CloudFormation::WaitConditionHandle
```
keithduncan, line 672: We don’t create a wait condition for this handle so it completes immediately, this is just a convenient cloudformation type to use for the 'barrier' resource.

keithduncan:
I have created and deleted a stack from this template and the ordering was as I expected.

templates/aws-stack.yml
```diff
@@ -665,6 +666,16 @@ Resources:
...
+    Metadata:
+      VpcResources: !If
+        - CreateVpcResources
+        - [ !Ref RouteDefault, !Ref Subnet0Routes, !Ref Subnet1Routes ]
```
pda, line 676: I assume these are meaningless as metadata, and only present to make this `VpcComplete` resource implicitly depend on those other resources? And maybe you can't use a regular `DependsOn` because they don't support conditionality like `!If CreateVpcResources`?

templates/aws-stack.yml
```diff
@@ -739,6 +750,36 @@ Resources:
...
+                - s3:GetObjectVersion
+                - s3:GetObjectVersionAcl
+                - s3:ListBucket
+                - s3:PutObject
+                - s3:PutObjectAcl
+                - s3:PutObjectVersionAcl
+              Resource:
+                - !Sub "arn:aws:s3:::${ArtifactsBucket}/*"
+                - !Sub "arn:aws:s3:::${ArtifactsBucket}"
+            - !Ref AWS::NoValue
```
pda, line 782: A small part of me likes that these were separate policies, with names indicating what they're for.
But not a big part.
Maybe we could add descriptive statement IDs (`Sid`) to give them descriptive names at least?

pda (commented):
Looking good.

templates/aws-stack.yml
keithduncan: Exactly correct :+1:

templates/aws-stack.yml
keithduncan: I agree, though there is no way to conditionally `DependsOn` them because they were themselves `Condition`d resources.

We could do the same trick as for the VPC with the policies, but I think it’s less to think about having them conditionally shimmed into the `IAMPolicies` resource that can be unconditionally `DependsOn`d.

Adding an `Sid` is a good idea :+1: our other option would be to make these managed policies which are included in the role directly creating a suitable dependency chain. `ASG -> LaunchTemplate -> InstanceProfile -> IAMRole ->> ManagedPolicies`

PR #949, opened 2021-10-18
Use userns-remap=buildkite-agent unconditionally, conditionally delete config at boot time

### LINKED ISSUE #950: AMI cached docker images are not present on instances with EnableDockerUserNamespaceRemap: true
Ideally we can find a way to produce an AMI which when built upon to cache resources supports both `EnableDockerUserNamespaceRemap: true` and `EnableDockerUserNamespaceRemap: false`.

Presently the best workaround to support pre-cached images is to set `EnableDockerUserNamespaceRemap: false` in you stack’s parameters.

keithduncan:
Needs testing, userns-remap is causing issues with pre-cached Docker images being unavailable once the userns-remap config is applied. We need to test whether images cached with this in place are available to the daemon when running with this turned off.

---

Test procedure:

1. Build off the resulting Linux image and cache some docker images
2. Deploy stack with the built AMI that has `EnableDockerUserNamespaceRemap: true`
    - [ ] Verify cached images are present
3. Deploy stack with the built AMI that has `EnableDockerUserNamespaceRemap: false`
    - [ ] Verify cache images are present

Fixes #950

### DIFF
--- a/packer/linux/scripts/install-buildkite-agent.sh
+++ b/packer/linux/scripts/install-buildkite-agent.sh
@@ -20,6 +20,7 @@ sudo useradd --base-dir /var/lib --uid 2000 buildkite-agent
 sudo usermod -a -G docker buildkite-agent
 
 echo "Configuring docker userns-remap to use the buildkite-agent user"
+#shellcheck disable=SC2094 # Redirections to the same command are processed in order
 cat <<< "$(jq '."userns-remap"="buildkite-agent"' /etc/docker/daemon.json)" > /etc/docker/daemon.json
 
 echo "Downloading buildkite-agent v${AGENT_VERSION} stable..."


keithduncan:
Initial test shows this hasn’t worked.

With the `EnableDockerUserNamespaceRemap: true` left on the cached images are present:

```
$ trap 'kill -- $' INT TERM QUIT; cat /etc/docker/daemon.json
--
  | docker images
  | {
  | "storage-driver": "overlay2",
  | "userns-remap": "buildkite-agent"
  | }
  | REPOSITORY    TAG       IMAGE ID       CREATED       SIZE
  | ruby          latest    090e42145ec2   5 days ago    881MB
  | python        latest    c05c608cfa20   5 days ago    915MB
  | golang        latest    9f8b89ee4475   5 days ago    941MB
  | hello-world   latest    feb5d9fea6a5   3 weeks ago   13.3kB
```

But with `EnableDockerUserNamespaceRemap: false` causing the `userns-remap` configuration to be removed the images are not present:

```
$ trap 'kill -- $' INT TERM QUIT; cat /etc/docker/daemon.json
--
  | docker images
  | {
  | "storage-driver": "overlay2"
  | }
  | REPOSITORY   TAG       IMAGE ID   CREATED   SIZE
```

Will need to dig into this some more.

Issue #148, opened 2016-09-20
Intermittent docker error "could not find an available predefined network"


zain:
I see these on builds occasionally:

```
$ docker-compose -f docker-compose.yml -p buildkite9c0e505a3f5844e4bde33ee36c928627 run node pug-lint app/assets/javascripts
Creating network "buildkite9c0e505a3f5844e4bde33ee36c928627_default" with the default driver
ERROR: failed to parse pool request for address space "LocalDefault" pool "" subpool "": could not find an available predefined network
```

Here's the Dockerfile I'm using:

```
FROM node:4.5.0

RUN mkdir -p /usr/src/app
WORKDIR /usr/src/app

COPY package.json /usr/src/app/

RUN npm install
RUN npm install -g eslint babel-eslint pug-lint

COPY . /usr/src/app
```

and the pipeline.yml:

```
steps:
  - name: ":docker:"
    plugins:
      docker-compose:
        build: node

  - wait

  - name: ":eslint:"
    command: "eslint --cache app/assets/javascripts spec/javascripts"
    plugins:
      docker-compose:
        run: node

  - name: ":dog:"
    command: "pug-lint app/assets/javascripts"
    plugins:
      docker-compose:
        run: node
```

lox:
I'd suggest you change those docker registry credentials now ;)

lox:
Perhaps related to this docker issue? https://github.com/docker/docker/issues/23971

keithpitt:
@lox already done :)

lox:
On reading the above issue a bit more, looks like there is a limit on the number of networks created, I wonder if we are not correctly cleaning up after builds with `docker-compose down`

zain:
Running this fixes the issue, but it's not really a solution because I can't just constantly be running this. `docker network ls | awk '{print $1}' | xargs docker network rm`

lox:
Which version of the stack are you seeing this in @zain? I don't have much experience with the docker-compose plugin, as it's not in the stable release. It does look to me like the networks aren't being cleaned up after each command.

lox:
Perhaps @sj26 might know.

sj26:
There is docker-compose support in the stable version using environment variables. :-)

We are running `docker-compose down` at the end of builds which should also remove any created networks:

https://github.com/buildkite/agent/blob/2-2-stable/templates/bootstrap.sh#L498-L499

We also run it in plugin (used by the beta agent):

[...]

PR #968, opened 2021-11-29
Allow Docker address pool customisation

ouranos:
# What

Adds the `EnableDockerClassCSubnets` flag which configure the Docker address pool to allocate `/24` subnet instead of the default `/16` and`/20`.

# Why

We're using large instance with nvme storage to leverage the docker cache in our build (`c5d.12xlarge` with 12 agents), we often get the following error:

```
ERROR: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network
```

This is because the default docker address pool configuration is as follow:

Type | Default Size | Default Pool
-- | -- | --
global | /24 | 10.0.0.0/8
local | /16 | 172.17.0.0/12
local* | /20 | 192.168.0.0/16

\* _Local networks are allocated from 172.17.0.0/12, and then 192.168.0.0/16 once 172.17.0.0/12 is exhausted._

This means you'll get the error once 32 networks are created, which can happen with large instances.

Using a `/24` allows up to 4,352 networks with 256 addresses each which should be a good compromise.

# Test

With the default settings:
```
$ sudo docker network inspect buildkitexxx_default | jq ".[].IPAM.Config"
[
  {
    "Subnet": "172.22.0.0/16",
    "Gateway": "172.22.0.1"
  }
]
```

With the option enabled:
```
$ sudo docker network inspect buildkitexxx_default | jq ".[].IPAM.Config"
[
  {
    "Subnet": "172.16.1.0/24",
    "Gateway": "172.16.1.1"
  }
]
```

# Notes

This is related to #148

Happy to rework the PR to allow for more customisation of the address pools but I thought this was a good balance between flexibility and ease of use.

Another option would be to reduce the `DOCKER_PRUNE_UNTIL` variable (which defaults to 4h) to clear more networks but this will also clear the image cache more often which is not ideal. And it still might cause issues if running lots of agents in parallel on the same instance.

### DIFF
--- a/packer/linux/conf/bin/bk-configure-docker.sh
+++ b/packer/linux/conf/bin/bk-configure-docker.sh
@@ -22,3 +22,6 @@ fi
 if [[ "${BUILDKITE_ENABLE_INSTANCE_STORAGE:-false}" == "true" ]] ; then
   cat <<< "$(jq '."data-root"="/mnt/ephemeral/docker"' /etc/docker/daemon.json)" > /etc/docker/daemon.json
 fi
+
+# Customise address pools
+cat <<<"$(jq '."default-address-pools"=[{"base":"172.17.0.0/12","size":20},{"base":"192.168.0.0/16","size":24}]' /etc/docker/daemon.json)" >/etc/docker/daemon.json


ouranos:
@keithduncan any thoughts on this?

keithduncan:
👋 @ouranos I’m no longer with Buildkite or working on this project, I’d suggest contacting your account manager

DanielHeath:
I'd really like to see this implemented; large build agents are a huge performance win, but fall over with `ERROR: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network` when too many builds are run concurrently.

eleanorakh:
Thanks so much @ouranos. This seems reasonable, we'll find the time to internally review this soon!

pda:
Our current thinking is to make the `/24` Class C network size the default, without a new parameter to choose otherwise.
Can anybody see a problem with that?
Is anybody using more than ~250 IPs within each Docker network?

DanielHeath:
That would solve my problem neatly, and I can't see how it'd introduce any new issues.

ouranos:
Yes, that would work for us. 

I don't think anybody would use that many IPs in a docker network unless they have really specific requirements. 
Maybe leave the option to override for that edge case?

Otherwise happy with the new default 👌

ouranos:
> Our current thinking is to make the /24 Class C network size the default, without a new parameter to choose otherwise.

@pda would you like me to open a PR for this?

ouranos:
@eleanorakh @pda any updates on this?

pda:
Hey @ouranos, sorry for the silence.

> would you like me to open a PR for this?

That would be wonderful, if you can either update this PR or create a new PR.

You might have already seen this, but for anybody else following along, here's a great write-up on the topic:

> The definitive guide to docker's default-address-pools option
> https://straz.to/2021-09-08-docker-address-pools/

In that write-up, the author suggests the following configuration:

```json
{
  "default-address-pools" : [
    {
      "base" : "172.17.0.0/12",
      "size" : 20
    },
    {
      "base" : "192.168.0.0/16",
      "size" : 24
    }
  ]
}
```

Or, expressed as a diff against the default configuration:

```diff
--- daemon-old.json	2021-09-12 17:56:03.344023200 +1000
+++ daemon-new.json	2021-09-12 17:52:49.274194000 +1000
@@ -2,11 +2,11 @@
   "default-address-pools" : [
     {
       "base" : "172.17.0.0/12",
-      "size" : 16
+      "size" : 20
     },
     {
       "base" : "192.168.0.0/16",
-      "size" : 20
+      "size" : 24
     }
   ]
 }
```


I like the idea of expanding each pool mask by 4 bits (making each subnet 16x larger), and preserving the two different sizes. That way, the first 256 networks are large `/20` (4096 IPs) within `172.17.0.0/12`, and once those are used up the following 256 networks are smaller `/24` (256 IPs) within `192.168.0.0/16 ` so that a total of 512 networks are possible.

That means most people get large networks, but people that use heaps of networks end up with smaller ones.

So — I guess I'm proposing size:20 for 172.17.0.0/12, rather than size:24 for both :)

I love your use of `jq` to update the otherwise-awkward-to-update JSON config file :)

packer/linux/conf/bin/bk-configure-docker.sh
```diff
@@ -22,3 +22,8 @@ fi
 if [[ "${BUILDKITE_ENABLE_INSTANCE_STORAGE:-false}" == "true" ]] ; then
   cat <<< "$(jq '."data-root"="/mnt/ephemeral/docker"' /etc/docker/daemon.json)" > /etc/docker/daemon.json
 fi
+
+# Customise address pools
+if [[ "${DOCKER_CLASS_C_SUBNETS:-false}" == "true" ]] ; then
+  cat <<< "$(jq '."default-address-pools"=[{"base":"172.17.0.0/12","size":24},{"base":"192.168.0.0/16","size":24}]' /etc/docker/daemon.json)" > /etc/docker/daemon.json
```
pda, line 28: I think this is the one line we want to keep, making it default and non-optional for now.

pda: Hmm, it just occurred to me: often with shell pipelines it's problematic to write to the file that you're reading from. Is `jq` guaranteed to fully read `/etc/docker/daemon.json` prior to opening it for truncating-write via `>` output redirection?

ouranos: Sounds good, I'll update this PR now.
Yes, it works on our deployment and it's the same pattern as for the instance storage.

ouranos:
@pda Rebased and updated with the new size

pda:
Excellent, thanks @ouranos

PR #976, opened 2021-12-17
add CpuCredits parameter

TheDen:
PR Adds [CreditSpecification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata.html#cfn-ec2-launchtemplate-launchtemplatedata-creditspecification) in the [AWS::EC2::LaunchTemplate LaunchTemplateData](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata.html) specification.

Defaulting to `unlimited`, but can be `standard` if it's a better default.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -60,6 +60,7 @@ Metadata:
         - ImageId
         - ImageIdParameter
         - InstanceType
+        - CpuCredits
         - EnableInstanceStorage
         - AgentsPerInstance
         - KeyName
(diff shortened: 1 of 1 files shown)


templates/aws-stack.yml
```diff
@@ -931,6 +939,8 @@ Resources:
           BlockDeviceMappings:
             - DeviceName: !If [ UseDefaultRootVolumeName, !If [ UseWindowsAgents, /dev/sda1, /dev/xvda ], !Ref RootVolumeName ]
               Ebs: { VolumeSize: !Ref RootVolumeSize, VolumeType: !Ref RootVolumeType }
+          CreditSpecification:
+            CpuCredits: !Ref CpuCredits
```
pda, line 943: I notice the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-launchtemplatedata.html#cfn-ec2-launchtemplate-launchtemplatedata-creditspecification) say:

> The credit option for CPU usage of the instance. Valid for T2, T3, or T3a instances only.

Will the presence of this option break the LaunchTemplate for other instance types?
Or are they happy [as long as it's set to `unlimited`?]?

TheDen: Hmm I'm not certain about this either. I tried to test it locally but couldn't get it working.

TheDen: Perhaps a cleaner solution would be to conditionally add the `CreditSpecification` YAML key+value if `InstanceType` matches one of the supported EC2 types?

PR #979, opened 2022-01-12
Recognise c7g instances as ARM

pda:
Recognise [newish](https://aws.amazon.com/blogs/aws/join-the-preview-amazon-ec2-c7g-instances-powered-by-new-aws-graviton3-processors/) `c7g.*` instance types as ARM so they use the `linuxarm64` AMI.

This PR is a subset of https://github.com/buildkite/elastic-ci-stack-for-aws/pull/969 …  there are other new ARM instance types, but this brings us to the limit of 10 conditions within `Fn::Or` in CloudFormation. Of the new instance type, `c7g` sounds most generally useful for CI workloads, maybe, and has been [specifically requested](https://github.com/buildkite/elastic-ci-stack-for-aws/pull/969#issuecomment-1010509706).

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -582,15 +582,16 @@ Conditions:
 
     UsingArmInstances:
       !Or
-        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "m6g" ]
-        - !Equals [ !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "m6gd" ]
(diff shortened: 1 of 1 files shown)


pda:
Oh and:

> While I was adding to the list I also re-orderde it to be alphabetical. Mainly so it's easier to scan and check for missing types.
— https://github.com/buildkite/elastic-ci-stack-for-aws/pull/969

… this PR does that too, which is why it doesn't appear a `+1/-0` diff.

eleanorakh:
👍🏼💃

PR #980, opened 2022-01-13
InstanceFamilies mappings for exhaustive arm64 instance families.

pda:
This is an attempt to address the problems described in #969 by introducing an exhaustive mapping of instance families to their architecture (`x86_64` or `arm64`) automatically updatable via the DescribeInstanceTypes API.

However, it seem this approach might be a dead-end because of this [FindInMap limitation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-findinmap.html#w2ab1c31c28c26c13):

> E1011 FindInMap only supports [Fn::FindInMap, Ref] functions at Conditions/UsingArmInstances/Fn::Equals/0/Fn::FindInMap/1

I discovered that _after_ sinking way too much time into `jq` and `awk`.

The error comes from the `FindInMap` here:

```
Conditions:
    UsingArmInstances:
      !Equals
        - !FindInMap [ InstanceFamilies, !Select [ 0, !Split [ ".", !Ref InstanceType ] ], "Arch"]
        - arm64
```

Seems we can't use the result of `!Select` to grab the (first) instance family from the `InstanceType` (comma separated list), because `FindInMap` doesn't support that. Frustrating.

I'll open this PR as a draft anyway, in case it inspires any other creative ideas :)

e.g. is there some way to put the result of `!Select [ 0, !Split [ ".", !Ref InstanceType ] ]` somewhere that can be accessed by `!Ref`? Probably not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,30 @@ build/aws-stack.yml:
 	}' templates/aws-stack.yml | sed "s/%v/$(VERSION)/" > $@
 
 # -----------------------------------------
+# Instance family mapping update.
+# Unlike AMI mapping, these change infrequently so the results are committed.
+
+# requires AWS authentication
+build/InstanceFamilies.yaml:
+	echo "  InstanceFamilies:" > $@
+	aws ec2 describe-instance-types | jq -r '.InstanceTypes | sort_by(.InstanceType) | .[] | \
+	    (.InstanceType | split(".")[0]) as $$family | \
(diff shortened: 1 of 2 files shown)


pda:
We could easily build a much larger mapping of all instance _types_ (not just families, but sizes too) to architectures. I'm not sure if that would end up exceeding a static limit. But even then we don't have a static `FindInMap` lookup key because our `InstanceType` parameter is actually a comma-separated list since #710 added support for Mixed Instance ASG.

PR #981, opened 2022-01-25
Add support for additional ARM/Graviton instance types, reprise

toothbrush:
This PR is a retry of https://github.com/buildkite/elastic-ci-stack-for-aws/pull/969 and supersedes it – the problem encountered there is that the CloudFormation [intrinsic function `!Or`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-conditions.html#intrinsic-function-reference-conditions-or) can only take between minimum 2 and maximum 10 arguments.

To address that issue i've perpetrated a bit of a hack, but it works.  Basically, split the conditional into a sub-clause for each instance family.

I don't recall if it's possible to switch base branch in already-created PRs, so here's a fresh one.  Apologies to the authors of the previous one – feel free to cherry-pick my changes if you prefer to keep the original PR open.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -580,17 +580,24 @@ Conditions:
     UseLinuxAgents:
       !Equals [ !Ref InstanceOperatingSystem, "linux" ]
 
+    # Unfortunately, Cloudformation's !Or intrinsic function only accepts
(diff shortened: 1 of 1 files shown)


pda:
🎉

PR #984, opened 2022-02-10
Permissions boundary for Autoscaling application

kwong-chong-lfs:
Attempts to deploy a Buildkite elastic-ci stack via a role with permission boundary currently fail.  It looks like the autoscaling application doesn't apply the boundary condition ARN passed to the parent stack.  
These two PRs:
https://github.com/buildkite/buildkite-agent-scaler/pull/62
https://github.com/buildkite/elastic-ci-stack-for-aws/pull/984
address this.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -1186,7 +1186,7 @@ Resources:
     Properties:
       Location:
         ApplicationId: arn:aws:serverlessrepo:us-east-1:172840064832:applications/buildkite-agent-scaler
-        SemanticVersion: '1.1.3'
+        SemanticVersion: '1.3.1'
       Parameters:
         BuildkiteAgentTokenParameter: !If [ UseCustomerManagedParameterPath, !Ref BuildkiteAgentTokenParameterStorePath, !Ref BuildkiteAgentTokenParameter ]
         BuildkiteAgentTokenParameterStoreKMSKey: !If [ UseCustomerManagedKeyForParameterStore, !Ref BuildkiteAgentTokenParameterStoreKMSKey, "" ]
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -1190,6 +1190,7 @@ Resources:
       Parameters:
         BuildkiteAgentTokenParameter: !If [ UseCustomerManagedParameterPath, !Ref BuildkiteAgentTokenParameterStorePath, !Ref BuildkiteAgentTokenParameter ]
         BuildkiteAgentTokenParameterStoreKMSKey: !If [ UseCustomerManagedKeyForParameterStore, !Ref BuildkiteAgentTokenParameterStoreKMSKey, "" ]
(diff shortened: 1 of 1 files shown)


templates/aws-stack.yml
```diff
@@ -1190,6 +1190,7 @@ Resources:
       Parameters:
         BuildkiteAgentTokenParameter: !If [ UseCustomerManagedParameterPath, !Ref BuildkiteAgentTokenParameterStorePath, !Ref BuildkiteAgentTokenParameter ]
         BuildkiteAgentTokenParameterStoreKMSKey: !If [ UseCustomerManagedKeyForParameterStore, !Ref BuildkiteAgentTokenParameterStoreKMSKey, "" ]
+        RolePermissionsBoundaryARN: !If [ SetInstanceRolePermissionsBoundaryARN, !Ref InstanceRolePermissionsBoundaryARN, "" ]
```
moskyb, line 1193: would it be worth having a separate permissions boundary for the scaling lambda and and the EC2 instance? it seems like they have separate concerns in terms of what kind of other AWS services they interact with

pda: I was wondering that too.
But, I suspect in most/all organizations, a single Permission Boundary will be assigned to a CI stack (or even to the team operating that stack), so maybe nobody would make use of a separate one.

I think we could ship it like this now, and potentially add separate parameters/boundaries for different components if/when asked for. For backwards-compatibility, we could use conditionals to fall back to the existing shared `InstanceRolePermissionsBoundaryARN` if it's present and a specific one isn't. So — we're not painting ourselves into a corner by merging this as-is, I think.

moskyb: ☝️this isn't a "please do this" review comment btw, i'm genuinely interested as we don't tend use permissions boundaries much internally.

kwong-chong-lfs: We would only have one boundary condition per deployment.  Having a second condition as an option should be fine - just need a default and conditions for those who opt to only use one boundary condition.

moskyb:
@kwong-chong-lfs i've merged + released https://github.com/buildkite/buildkite-agent-scaler/pull/62, would you be able to update this PR to use v1.3 of the SAR repo for the agent scaler?

moskyb:
also! apologies for letting this PR rot a little bit! i got it in my head that i'd merged this, but that's obviously not the case 😅 we'll work to get this merged ASAP.

kwong-chong-lfs:
> 

@moskyb Updated to v1.3 of buildkite-agent-scaler as requested.

moskyb:
Awesome, thanks heaps @kwong-chong-lfs! We're hoping to get the repo released later today - after that's done i'll merge this PR in :)

moskyb:
Hey @kwong-chong-lfs - sorry this has taken so long to sort out, there were some shenanigans with the deploy process with the agent scaler. They're all sorted out now though - i've updated this PR a little bit, but i'll merge it once the build is done.

kwong-chong-lfs:
All good @moskyb.  Thank you for the update.

PR #988, opened 2022-02-25
Add default metrics to CloudWatch agent

ouranos:
Following #811, we've removed our custom patch to install the CloudWatch agent.

The default config (downloaded with `amazon-cloudwatch-agent-ctl -m ec2 -a fetch-config -c default -s`) had an extra config block to reports on memory and disk usage.

This is a nice feature to monitor buildkite instances utilisation (and assist with dimensioning).

For example, see the bottom right graph:
image

We could potentially add other dimensions to make it easier to differentiate by queue name. I'm currently using the ASG name which might change when updating the stack.


packer/linux/conf/cloudwatch-agent/config.json
```diff
@@ -2,6 +2,29 @@
...
+				"resources": [
+					"*"
+				]
+			}
+		},
+		"append_dimensions": {
+			"ImageId": "${aws:ImageId}",
+			"InstanceId": "${aws:InstanceId}",
+			"InstanceType": "${aws:InstanceType}",
+			"AutoScalingGroupName": "${aws:AutoScalingGroupName}"
```
freewil, line 25: for purposes of cloudwatch metric billing, each unique dimension value is considered 1 metric, so including granular dimensions like `InstanceId` or even `InstanceType` could really blow up costs for those of us with a large number of ec2 instances.

ouranos: Good point. How about having just the `AutoScalingGroupName` dimension then?
Or would there be a way to have the queue as a dimension?

freewil: yeah i think that makes sense. not sure if it's possible to add the queue, but i think that'd be nice as well. It should be available as a tag on the instance, `BuildkiteQueue`.

freewil: I looked into this a bit, and one way to add `BuildkiteQueue` as a dimension would be to modify the cloudwatch config file on boot using jq ([similar to docker](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/19278b86f3e08cce38f144811f9c9c0c253589fd/packer/linux/conf/bin/bk-configure-docker.sh#L18)), as part of the `bk-install-elastic-stack.sh` script. You should have access to [BUILDKITE_QUEUE](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/19278b86f3e08cce38f144811f9c9c0c253589fd/packer/linux/conf/bin/bk-install-elastic-stack.sh#L117) from there and then just enable and start the service on boot instead of [at AMI creation time](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/d2230c5163ecdf4cc87f46c8c72af2ea02f5b11e/packer/linux/scripts/install-cloudwatch-agent.sh#L13).

ouranos:
@freewil Ok updated the PR after much testing (lost a bit of time with the issue mentioned in #995)

I've done the following changes:
- Added a new stack parameter (`EnableCloudWatchMetrics`) to opt-in - by default the behavior stays the same (ie, no metrics)
- Deferred the start of the CloudWatch agent to boot hook `bk-configure-cloudwatch-agent.sh` (so we can generate the configuration at run time)
- Tweaked the `append_dimensions`

As per the [documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html#CloudWatch-Agent-Configuration-File-Metricssection):

> The only supported key-value pairs for append_dimensions are shown in the following list. Any other key-value pairs are ignored. ["ImageID", "InstanceId", "InstanceType", "AutoScalingGroupName"]
> If you want to append dimensions to metrics with arbitrary key-value pairs, use the append_dimensions parameter in the field for that particular type of metric.

After this change I get the following metrics in CloudWatch (ignore the pollution due to various tests):
image

image
image

I reckon we don't need both the ASG name and the Queue, so we could drop one of them.

Let me know what you think. Maybe we can change the `namespace` too? Or make it more customisable?
I don't really care about the disk metrics (the auto clean up works fine for us) so we could drop it too.

packer/linux/conf/bin/bk-configure-cloudwatch-agent.sh
```diff
@@ -0,0 +1,24 @@
...
+            metrics: {
+                metrics_collected: {
+                    mem: {measurement: ["mem_used_percent"], append_dimensions: {BuildkiteQueue: $queue}},
+                    disk: {measurement: ["used_percent"], resources: ["*"], append_dimensions: {BuildkiteQueue: $queue}}
+                },
+                append_dimensions: {
+                    AutoScalingGroupName: "${aws:AutoScalingGroupName}"
+                }
+            }
+        }' $cw_config)" >$cw_config
```
freewil, line 19: so looks like this will append to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/509802db39718c8e0623535804237fed35f6f516/packer/linux/conf/cloudwatch-agent/config.json so that the existing file doesn't get clobbered?

ouranos: Yes, it actually rewrites it. It's the same pattern used to [configure docker](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/19278b86f3e08cce38f144811f9c9c0c253589fd/packer/linux/conf/bin/bk-configure-docker.sh#L18), albeit more complicated.

* The `jq` command will read the `$cw_config` file, append the metrics configuration block and output it (`jq [options] <jq filter> [file...]`). Where the jq filter is:
  - `.` identity filter
  - `{}`: object construction
  - `+`: addition operator between two objects. So takes the original configuration and merge the new object created:
     > Objects are added by merging, that is, inserting all the key-value pairs from both objects into a single combined object.
* This output is passed to the standard input cat with a `here-string` (`<<<`) using command substitution (`$(..)`). So the jq command runs in a subshell.
* `cat` is then overwriting the original config file

It's equivalent to:

```bash
new_config=$(jq <filter> $cw_config)
echo $new_config | cat > $cw_config
```
 
From what I understand, the combination of here-string and command substitution will use a temp file or memory to avoid any race condition while modifying the file in place.

freewil (approved):
cool - LGTM. I don't work for buildkite btw, but would love to see this get merged.

packer/linux/conf/bin/bk-configure-cloudwatch-agent.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [[ "${CLOUDWATCH_ENABLE_METRICS:-false}" == "true" ]]; then
+    cw_config="/opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.json"
+    cat <<<"$(jq \
+        --arg queue "$BUILDKITE_QUEUE" \
+        '. + {
+            metrics: {
+                metrics_collected: {
+                    mem: {measurement: ["mem_used_percent"], append_dimensions: {BuildkiteQueue: $queue}},
+                    disk: {measurement: ["used_percent"], resources: ["*"], append_dimensions: {BuildkiteQueue: $queue}}
```
moskyb, line 13: if we're adding the queue as dimension for memory and disk, should we do it for CPU too?

moskyb (commented):
this is awesome @ouranos! thanks for submitting. 

> Maybe we can change the namespace too? Or make it more customisable?

i think adding a namespace is probably a good idea, it'll make filtering in the console significantly less of a pain. maybe something like `BuildkiteElasticStack/Agent` or something similar? very open to other ideas though, and agree that making it customizable (but with a sensible default) is a good idea.

PR #989, opened 2022-02-28
autoscaling:DescribeAutoScalingInstances can only be applied to all resources

jeremiahsnapp:


pda:
Thanks — this is supported by the docs for [Actions, resources, and condition keys for Amazon EC2 Auto Scaling](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonec2autoscaling.html#amazonec2autoscaling-actions-as-permissions):

image

PR #601, opened 2019-06-23
Add support for loading BuildkiteAgentTokenPath from Parameter Store

jradtilbrook:
This provides changes to allow using a Buildkite agent token that is set in AWS SSM.

It should help with #422 and #597.

I'm assuming the lambda code is coming from [buildkite-agent-scaler](https://github.com/buildkite/buildkite-agent-scaler) since that's the name of the bucket the lambda binary is taken from. I've looked through that code and found that it already supports retrieving the token from SSM by setting `BUILDKITE_AGENT_TOKEN_SSM_KEY`. The only remaining thing I can think of is adding the IAM permissions to the lambda to allow it to read and/or decrypt the parameter in SSM.

Before proceeding with this further, can I get the maintainers (and community I suppose) opinion on this change? I could make further changes so that it's hopefully backwards compatible, but currently it is not.

PR #993, opened 2022-03-01
BuildkiteAgentToken vs BuildkiteAgentTokenParameterStorePath clarity.

pda:
Clarify `BuildkiteAgentToken` vs `BuildkiteAgentTokenParameterStorePath` in the stack parameter descriptions.

I don't think `BuildkiteAgentToken` is _clearly_ deprecated (yet?), so don't mention that here just now.

I think we should first revisit the dynamics/security of [CloudFormation `NoEcho` parameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties), and the ability to create encrypted SSM Parameter Store parameters from CloudFormation.

When `BuildkiteAgentToken` is provided, we currently store it in SSM Parameter Store anyway:

```yaml
  BuildkiteAgentTokenParameter:
    Type: AWS::SSM::Parameter
    Condition: CreateAgentTokenParameter
    Properties:
      Name: !Sub "/${AWS::StackName}/buildkite/agent-token"
      Type: String
      Value: !Ref BuildkiteAgentToken
```

But it's stored unencrypted, which I suspect was a limitation of CloudFormation : SSM integration. Maybe that's improved?

Also, one of the reasons `BuildkiteAgentTokenParameterStorePath` was introduced (https://github.com/buildkite/elastic-ci-stack-for-aws/pull/601) was because managing a `NoEcho` `BuildkiteAgentToken` parameter in this stack via Terraform resulted in problems which sound more like a Terraform bug, which may or may not have been fixed since.

So — I propose we clarify the descriptions a little and un-deprecate `BuildkiteAgentToken` until we're sure of its fate.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -127,13 +127,13 @@ Parameters:
     Default: "stable"
 
   BuildkiteAgentToken:
-    Description: Buildkite agent registration token. Deprecated, use BuildkiteAgentTokenParameterStorePath instead.
+    Description: Buildkite agent registration token. Or, preload it into SSM Parameter Store and use BuildkiteAgentTokenParameterStorePath for secure environments.
     Type: String
     NoEcho: true
     Default: ""
 
   BuildkiteAgentTokenParameterStorePath:
-    Description: AWS SSM path to the Buildkite agent registration token (this takes precedence over BuildkiteAgentToken). Expects a leading slash ('/').
+    Description: Existing SSM Parameter Store path to the Buildkite agent registration token (takes precedence over BuildkiteAgentToken). Expects a leading slash ('/').
     Type: String
     Default: ""
     AllowedPattern: "^$|^/[a-zA-Z0-9_.\\-/]+$"


templates/aws-stack.yml
```diff
@@ -127,13 +127,13 @@ Parameters:
     Default: "stable"
 
   BuildkiteAgentToken:
-    Description: Buildkite agent registration token. Deprecated, use BuildkiteAgentTokenParameterStorePath instead.
+    Description: Buildkite agent registration token. Or, preload it into SSM Parameter Store and use BuildkiteAgentTokenParameterStorePath for secure environments.
```
moskyb, line 130: i think we could probably use some slightly stronger wording here - maybe something like 

```suggestion
    Description: Buildkite agent registration token. For production workloads, we recommend preloading the token into SSM Parameter Store and using BuildkiteAgentTokenParameterStorePath instead.
```

i think we want to make it clear that the SSM is the golden path for prod workloads, and minimise confusion between the two parameters.

pda: I'm not sure that we're sure that we SSM is the golden path.
I think step one is making it clear that both work, and step two (out of scope for this PR) is to lay a roadmap for consolidating them… possibly even to _neither_ of the existing options. e.g. Secrets Manager.

moskyb: right! makes sense. in that case, LGTM :)

pda:
The CI build is stuck on due to an issue we're investigating re. latest Amazon Linux 2 AMI running on Nitro EC2 instances, but I'm confident updating these CloudFormation parameter descriptions is safe, so 🚀

PR #995, opened 2022-03-02
Use Amazon Linux 2 AMI with working cloud init

moskyb:
There's an issue confirmed by AWS support with the more recent 20220218 release where instances using AWS Nitro to fail to run cloud-init, so this PR adjusts our Packer AMI build process to use the 20220207 AL2 AMI as the base image, which doesn't have the issue.

We'll revert this PR when AWS support have confirmed that the issue is resolved.

### DIFF
--- a/packer/linux/buildkite-ami.json
+++ b/packer/linux/buildkite-ami.json
@@ -12,7 +12,7 @@
       "region": "{{user `region`}}",
       "source_ami_filter": {
         "filters": {
-          "name": "amzn2-ami-hvm-2.0.*-gp2",
+          "name": "amzn2-ami-hvm-2.0.20220207.1-gp2",
           "architecture": "{{user `arch`}}",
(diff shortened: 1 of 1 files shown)


packer/linux/buildkite-ami.json
```diff
@@ -12,7 +12,7 @@
       "region": "{{user `region`}}",
       "source_ami_filter": {
         "filters": {
-          "name": "amzn2-ami-hvm-2.0.*-gp2",
+          "name": "amzn2-ami-hvm-2.0.20220207.1-{{user `arch`}}-gp2",
```
pda, line 15: Assuming that `2.0.20220207.1` segment is the same for all `arch` that we support (which I'm fairly happy to assume) then that looks good as a temporary solve 👍🏼

pda (approved):

PR #999, opened 2022-03-25
Bump buildkite-agent to v3.35.0

moskyb:
We released Agent v3.35 earlier today; we should bump the elastic stack to use this version and include it in Elastic Stack v5.8.


pda (approved):
👍🏼 assuming CI builds end up happy.

PR #1001, opened 2022-03-27
Revert "Use Amazon Linux 2 AMI with working cloud init"

moskyb:
Reverts buildkite/elastic-ci-stack-for-aws#995

New AMIs have been released that fix this issue, so we can safely go back to using the latest AMIs :)

### DIFF
--- a/packer/linux/buildkite-ami.json
+++ b/packer/linux/buildkite-ami.json
@@ -12,7 +12,7 @@
       "region": "{{user `region`}}",
       "source_ami_filter": {
         "filters": {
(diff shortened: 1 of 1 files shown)


pda:
Did we end up trying this advice from AWS?

>  "Customer to run the following command cloud-init clean --logs to clean up cloud-init data as a best practice before creating a new image:". Hence, on your source instance, please run the below commands before creating the AMI:

`cloud-init clean --logs` immediately prior to shutdown.

It sounds like that would leave us more resilient to potential future issues also related to stale data in cloud-init cache.

pda (approved):

moskyb:
i tried adding that to our packer steps, but it didn't work 🤷

PR #1008, opened 2022-04-12
Add quotes around AWS variables

ctgardner:
All references to AWS variables (eg, AWS::Region) have been enclosed
within quotations.

Omitting the quotes can cause some YAML parsers to fail.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -716,7 +716,7 @@ Resources:
       VpcResources: !If
         - CreateVpcResources
         - [ !Ref RouteDefault, !Ref Subnet0Routes, !Ref Subnet1Routes ]
-        - !Ref AWS::NoValue
+        - !Ref "AWS::NoValue"
 
   BuildkiteAgentTokenParameter:
     Type: AWS::SSM::Parameter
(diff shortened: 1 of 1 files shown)


ctgardner:
When upgrading from `v5.7.0` to `v5.8.0`, one of our tools was failing with a YAML parsing error. We discovered this was caused by the addition of this line to the template:

```yaml
LocalSecretsBucketRegion: !If [ CreateSecretsBucket, !Ref AWS::Region, !Ref SecretsBucketRegion ],
```

Specifically, the failure was caused by the second argument. Enclosing `AWS::Region` within quotes fixes this issue. I've proactively corrected all occurrences of this within the template.

I'm not an expert, but without the quotes this might technically be invalid YAML, which would explain the parsing error.

moskyb (approved):
ah, the joys of YAML parsing 🙃

this LGTM, it shouldn't break anything. i'll merge it in and release it as part of the next patch release, which will be pretty soon. thanks so much for your contribution!

out of curiosity, what was the tool that was failing to parse the YAML? there's quite a bit of variability between yaml parsers unfortunately, but i'd be keen to do some more testing on my end to dig into the problem a bit more.

ctgardner:
@moskyb We're using https://github.com/cultureamp/cfparams. This uses https://github.com/sanathkr/go-yaml internally.

PR #1011, opened 2022-04-22
Update docker

moskyb:
Linux: 20.10.9 -> 20.10.14
Windows: 20.10.7 -> 20.10.9

We can't use the same version of docker across both linux and windows because our method of installing docker on windows [is no longer supported](https://github.com/microsoft/Windows-Containers/issues/186#issuecomment-989749953), and isn't being updated with new versions. We'll need to properly fix this in time, but for now, we also need to update docker versions for the elastic stack

Fixes PIP-434


benesch:
Would be swell to get this merged! We're running into a bug in Docker v20.10.9 that's fixed in v20.10.10: https://github.com/MaterializeInc/materialize/pull/12017.

Anything we can do to help?

jordandcarter (approved):
🙌

moskyb:
@benesch i'll try to get this out in a patch release later today :)

benesch:
Awesome, thanks very much @moskyb!

moskyb:
@benesch this is in the process of being released as [v5.8.2](https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/2730). It should finish building in half an hour or so :)

benesch:
Thanks very much, @moskyb! Worked like a charm.

PR #1016, opened 2022-04-29
Allow accessing tags via instance metadata

jchanam:
With this new parameter, we can enable access from the instance to its tags via metadata.

This change does not break the current configuration, as defaults to the `disabled` value, with is the default one.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -928,6 +928,7 @@ Resources:
             HttpTokens: !Ref IMDSv2Tokens
             # Allow containers using a Docker network on the host to receive IDMSv2 responses
             HttpPutResponseHopLimit: 2
+            InstanceMetadataTags: enabled
           Monitoring:
             Enabled: !Ref EnableDetailedMonitoring
           ImageId: !If


moskyb (commented):
this change looks good to me at first blush - would it require agent changes to take advantage of tags being in instance metadata? afaict we're already looking in the instance metadata for some agent metadata.

additionally, would there be any drawback to enabling this by default?

jchanam:
Hi @moskyb !

I don't think there's any need to change the agent. This is used to access the tags of the EC2 instances from within them only (not another kind of metadata).

I'm not sure about any drawback of enabling it by default, but as the AWS default is having it disabled, I think it's less risky and more continuist not to change the default and actual behavior.

pda:
In general we prefer fewer configuration parameters, when possible.

I think EC2 disables IMDS `InstanceMetadataTags` by default to protect against private tag information (e.g. organization / team / individual / infrastructure information) being accessed by processes that weren't authorised to do so. In our case, the instance role already has `ec2:DescribeTags` IAM access, so any process on the machines can access that anyway.

So, I think it should be enabled for everyone, without a configuration parameter.

jchanam:
@pda Is this agreed on your side? I can change the PR to remove the parameter and enable it by default

pda:
Yep, @moskyb and I both agree we can enable this for everyone without a parameter 👍🏼

jchanam:
@pda @moskyb It's now updated to enable metadata tags by default without the parameter.

moskyb (approved):
LGTM @jchanam! I did a quick lil rebase to get rid of what i presume is an internal bug tracker ID for your org.

Thanks so much for your contribution!

PR #1019, opened 2022-05-19
Fix mktemp issue on ubuntu(linux)

jradtilbrook:
This fixes an issue when running the log-collector script on linux based environments (or perhaps GNU).

It works on Mac/Unix, but linux requires the template passed to `mktemp` to contain at least 3 `X`'s in order to work.

### DIFF
--- a/utils/log-collector
+++ b/utils/log-collector
@@ -55,7 +55,7 @@ outputFormat="${outputFormat:=text}"
 
 # Create a temporary directory to store our logs
 pwd=$(pwd)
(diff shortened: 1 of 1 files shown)


jeremybumsted (commented):
@jradtilbrook I wonder if this is fixed in the changes I made in #1020 ?

jradtilbrook:
@jeremybumsted I don't think so. The error is with the usage of mktemp and it needs literal `X`'s in the template string. But may as well add it to yours and do it all in one

PR #1020, opened 2022-05-19
Update log collector

jeremybumsted:
Some minor changes to use tar instead of zip when creating the archive and some refactoring for readability

### DIFF
--- a/utils/log-collector
+++ b/utils/log-collector
@@ -16,7 +16,7 @@ set -euo pipefail
 
 usage() {
 cat << EOF
-Usage: ./bk-log-collector -s <stack-name> -i <instance-id>
+Usage: ./log-collector -s <stack-name> -i <instance-id>
 Collect Buildkite Elastic CI stack logs from AWS CloudWatch
 -s      Auto Scaling Group Name, can be found at console.aws.amazon.com/ec2autoscaling/
 -i      Instance ID, the AWS instance ID for the instance you want to collect logs for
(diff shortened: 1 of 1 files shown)


utils/log-collector
```diff
@@ -16,7 +16,7 @@ set -euo pipefail
 
 usage() {
 cat << EOF
-Usage: ./bk-log-collector -s <stack-name> -i <instance-id>
+Usage: ./log-collector -s <stack-name> -i <instance-id>
```
jradtilbrook, line 19: Could this use `$0` or something like that so that it's dynamic and always correct?

utils/log-collector
```diff
@@ -54,48 +54,55 @@ now=$(date +"%F")
 outputFormat="${outputFormat:=text}"
 
 # Create a temporary directory to store our logs
-pwd=$(pwd)
-dir=$(mktemp -d "$pwd"/buildkite-logs-"$now")
+dir=$(mktemp -d buildkite-logs-"$now")
```
jradtilbrook, line 57: I've got a separate PR since this breaks on GNU I think. Maybe just roll it into this one. See #1019

jradtilbrook (commented):
Makes sense but haven't tested it. Also got some points below

utils/log-collector
jeremybumsted: good suggestion!

utils/log-collector
```diff
@@ -54,48 +54,55 @@ now=$(date +"%F")
...
 
 # Make an archive with the logs
-zip -jq buildkite-logs-"$now".zip "$dir"/*.log
+tar -czvf buildkite-logs-"$now".tar.gz "$dir"/*.log
 
 # Clean up the unneeded files
 rm -r "$dir"
 
-echo "Finished collecting logs, you can find them here: $pwd/buildkite-logs-$now.zip"
+pwd=$(pwd)
```
pzeballos, line 104: Maybe to keep it simple, remove this line and use the environment variable $PWD instead of creating a new one

utils/log-collector
```diff
@@ -54,48 +54,55 @@ now=$(date +"%F")
...
+
+echo "Finished collecting logs, you can find them here: ${pwd}/buildkite-logs-${now}.tar.gz"
```
pzeballos, line 106:
```suggestion
echo "Finished collecting logs, you can find them here: $PWD/buildkite-logs-${now}.tar.gz"
```

pzeballos (approved):
LGTM! just a small suggestion

moskyb (approved):
Looks great!

PR #1030, opened 2022-06-09
Make tracing easier to configure

moskyb:
**🤔 Problem:** We recently released the [opentelemetry-tracing experiment](https://github.com/buildkite/agent/pull/1632) for the agent, but currently, it's tricky to use through the elastic stack for a couple of reasons:
- We don't currently expose the `--tracing-backend` argument as a cfn parameter
- At current, there's no easy way to pack arbitrary environment variables into the agents launched by the elastic stack. To use opentelemetry tracing, we need to supply an envar like `OTEL_EXPORTER_OTLP_ENDPOINT`, which tells the agent where to send its traces, and it's tricky/impossible to set this envar via the elastic stack.

**✅ Solution:** Fix those two things! 
- Make the tracing-backend config element settable through cloudformation, defaulting to `""`
- ~~Add a new cfn parameter called `BuildkiteAgentExtraEnvironmentVariables`, a comma-separated string of envars. each of these envars will be added to the environment in which the agent runs, across both windows and linux.~~
- Add a cloudformation param to the stack for `EnvFileUrl`, working in a similar way to the `BootstrapScriptUrl` param.

### DIFF
--- a/packer/linux/conf/buildkite-agent/systemd/buildkite-agent.service
+++ b/packer/linux/conf/buildkite-agent/systemd/buildkite-agent.service
@@ -11,6 +11,8 @@ User=buildkite-agent
 Environment="HOME=/var/lib/buildkite-agent"
 Environment="PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin"
 Environment="USER=buildkite-agent"
+# The =- (rather than just = ) in the line below means that systemd won't complain if it can't find the file
+EnvironmentFile=-/var/lib/buildkite-agent/env
 ExecStart=/usr/bin/buildkite-agent start
 ExecStopPost=/usr/local/bin/terminate-instance
 RestartSec=5
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -70,6 +70,7 @@ Metadata:
         - ArtifactsBucket
         - AuthorizedUsersUrl
         - BootstrapScriptUrl
+        - AgentEnvFileUrl
         - RootVolumeSize
         - RootVolumeName
         - RootVolumeType
@@ -222,7 +223,12 @@ Parameters:
     Default: ""
 
   BootstrapScriptUrl:
-    Description: Optional - HTTPS or S3 URL to run on each instance during boot
+    Description: Optional - HTTPS or S3 URL for a script to run on each instance during boot
+    Type: String
+    Default: ""
+
+  AgentEnvFileUrl:
+    Description: Optional - HTTPS or S3 URL for a list of environment variables to propagate into the agent's execution environment. Note that these environment variables **will not** be passed into builds running on the agent, only to the agent process itself.
     Type: String
     Default: ""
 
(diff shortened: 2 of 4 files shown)


pda:
I wonder if we could point at a collection of environment variables stored elsewhere on AWS?

It could be like `BootstrapScriptUrl` which can be hosted on S3 or any arbitrary HTTPS endpoint.

Or it could be an SSM Parameter Store prefix; see https://github.com/cultureamp/parameter-store-exec for what I'm imagining on that front.

Externalising it sidesteps the questions around syntax/delimiters, and the janky bash/python needed to handle them, and also makes it less likely people will pass secrets to the agent via an insecure stack parameter.

🤔

moskyb:
@pda given the prior art of the bootstrap script, i think doing an environment file in a similar way is pretty reasonable. I'll update this PR so it does something similar to what we do for the bootstrap script.

packer/linux/conf/buildkite-agent/systemd/buildkite-agent.service
```diff
@@ -11,6 +11,8 @@ User=buildkite-agent
 Environment="HOME=/var/lib/buildkite-agent"
 Environment="PATH=/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin"
 Environment="USER=buildkite-agent"
+# The =- (rather than just = ) in the line below means that systemd won't complain if it can't find the file
```
pda, line 14: 👌🏼

sj26: Nice

templates/aws-stack.yml
```diff
@@ -222,7 +223,12 @@ Parameters:
     Default: ""
 
   BootstrapScriptUrl:
-    Description: Optional - HTTPS or S3 URL to run on each instance during boot
+    Description: Optional - HTTPS or S3 URL for a script to run on each instance during boot
+    Type: String
+    Default: ""
+
+  AgentEnvFileUrl:
+    Description: Optional - HTTPS or S3 URL for a list of environment variables to propagate into the agent's execution environment. Note that these environment variables **will not** be passed into builds running on the agent, only to the agent process itself.
```
sj26, line 231: Nice

sj26 (commented):
Wonderful!

templates/aws-stack.yml
```diff
@@ -1041,10 +1059,12 @@ Resources:
                   BUILDKITE_AGENT_TAGS="${BuildkiteAgentTags}" \
                   BUILDKITE_AGENT_TIMESTAMP_LINES="${BuildkiteAgentTimestampLines}" \
                   BUILDKITE_AGENT_EXPERIMENTS="${BuildkiteAgentExperiments}" \
+                  BUILDKITE_AGENT_TRACING_BACKEND="${BuildkiteAgentTracingBackend}" \
                   BUILDKITE_AGENT_RELEASE="${BuildkiteAgentRelease}" \
                   BUILDKITE_QUEUE="${BuildkiteQueue}" \
                   BUILDKITE_AGENT_ENABLE_GIT_MIRRORS_EXPERIMENT="${EnableAgentGitMirrorsExperiment}" \
                   BUILDKITE_ELASTIC_BOOTSTRAP_SCRIPT="${BootstrapScriptUrl}" \
+                  BUILDKITE_ENV_FILE_URL=${AgentEnvFileUrl} \
```
pda, line 1067: Could argue that these should be `BUILDKITE_AGENT_ENV_FILE_URL` but I don't think it matters too much 😁

PR #1031, opened 2022-06-23
Make cloudwatch log stream names explicit

moskyb:
When log_stream_name is omitted from cloudwatch agent config, it defaults to the id of the instance sending the logs. This PR makes this choice explicit, bringing the windows config in line with the linux cloudwatch config


DrJosh9000 (approved):
🥳

PR #1032, opened 2022-07-04
Don't accidentally send the token to logs on startup

moskyb:
The elastic stack sends its startup logs to the `/buildkite/elastic-stack/{instance_id}` log group in cloudwatch, which is basically a bash script with `set -x` enabled. As part of this process, we fetch the token from SSM Parameter store, but this is included in the output, meaning that the token gets leaked into cloudwatch logs, which are potentially much less privileged than the agent tokens themselves.

CleanShot 2022-07-05 at 10 46 49@2x

This PR disables set -x while we do secret things, and then turns it back on afterwards

### DIFF
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -161,7 +161,10 @@ then
 fi
 chown buildkite-agent: "${BUILDKITE_AGENT_BUILD_PATH}"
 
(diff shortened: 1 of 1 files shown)


pda (approved):
As discussed, I'm only partially convinced that we should keep the `set -x` in the first place.
I see `-x` mode as “active debugging” not to be left enabled “in production”.
But there's a lot of observability/debugging upside to leaving it on all the time (and carefully redacting secrets), so 🚀

PR #1033, opened 2022-07-06
Update docker version 20.10.17

hari2192:
Our team is experiencing issue https://github.com/moby/moby/pull/43354 which is apparently fixed in docker version 20.10.17 (latest available)


packer/linux/scripts/install-docker.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -eu -o pipefail
 
-DOCKER_VERSION=20.10.14
+DOCKER_VERSION=20.10.15
```
moskyb, line 4: While we're here, we might as well update it to [latest](https://docs.docker.com/engine/release-notes/#201017), i reckon!
```suggestion
DOCKER_VERSION=20.10.17
```

moskyb (changes requested):

moskyb (approved):

PR #1050, opened 2022-09-20
enable default encryption and enforce https

ckornacker:
Following the AWS Foundational Security Best Practices by enforcing encryption a rest and in transit. This change sets the default bucket encryption for the secrets bucket to AES256 and denies access to the bucket over HTTP to bring it in line with the security best practices and other compliance frameworks.

Happy to put this behind parameters as well instead of enabling it by default.


triarius:
Thanks for the PR @ckornacker. We think this is a good idea to add to the elastic stack, albeit behind a parameter.

Unfortunately, it looks like some things have broken in with our CI when deploying the encrypted bucket. We will need to look into it before merging this.

In the meantime, if you could add the parameter with encryption disabled by default, that help us.

triarius:
Hi @ckornacker, our CI should allow this to build now, so if you can rebase onto master and put it behind a parameter, we can get this merge.

ckornacker:
> Hi @ckornacker, our CI should allow this to build now, so if you can rebase onto master and put it behind a parameter, we can get this merge.

Hi @triarius, will do, thanks!

ckornacker:
@triarius I've rebased and added a parameter for the S3 encryption.

triarius (approved):
Thanks @ckornacker, LGTM!

PR #1057, opened 2022-11-04
Add property to indicate if the EBS volume is encrypted

pzeballos:
As per https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-launchtemplate-blockdevicemapping-ebs.html#cfn-ec2-launchtemplate-blockdevicemapping-ebs-encrypted, add the option to indicate if the EBS volume is encrypted or not. Default value: `false`

image

(requested by a customer 🙂)


triarius (commented):
Thanks! LGTM

PR #1061, opened 2022-11-10
WIP: Support for nix native inside pipelines

csakshaug:
As a long running user and customer of buildkite I am glad to perhaps contribute something back, especially to [elastic-ci-stack-for-aws](https://github.com/buildkite/elastic-ci-stack-for-aws) that we use a lot. We (@afreakk and me) have recently been discussing it would be amazing to have native nix support in the agents in the same way as aws-cli, docker etc.

The proof-of-concept this pull-request shows quite basically how we can add nix support. This solution requires you to add run nix-daemon before using it. Right now we have just tested with just adding it to .env

```
if [ -e '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh' ]; then
 . '/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
fi
```

We have been testing with building a custom image and changing cloudformation ImageId with this image and it works like charm.

If you guys think this is a good idea to add this feature we probably wanted to look into some optimizing, here is some point I have would have considered:

* Have this also a optional like docker support with parameters in the cloudformation
* Perhaps consider to have validation of the downloaded source
* Load nix-daemon more smoother then we do now
* Consider more examples of uses like with nix-shell and packages
* Squash on merge ;-)
* Tests

### DIFF
--- a/packer/linux/conf/cloudwatch-agent/config.json
+++ b/packer/linux/conf/cloudwatch-agent/config.json
@@ -47,6 +47,12 @@
 						"log_group_name": "/buildkite/system",
 						"log_stream_name": "{instance_id}",
 						"timestamp_format": "%b %d %H:%M:%S"
+					},
+					{
+						"file_path": "/var/log/secure",
+						"log_group_name": "/buildkite/auth",
+						"log_stream_name": "{instance_id}",
+						"timestamp_format": "%Y-%m-%d %H:%M:%S,%f"
 					}
 				]
 			}


csakshaug:
Side note, we have also been playing to build/debug with [nix and qemu in our fork](https://github.com/fivetentech/elastic-ci-stack-for-aws/tree/qemu-test).

triarius:
Hi @csakshaug. Thanks for contributing this PR!

Firstly, I think nix is a great tool, and it would be very useful to use in a CI system. Unfortunately, we can't include absolutely every such tool in the Elastic CI Stack or the AMI would balloon in size. We have tried to include what we believe there is strong demand for, and leave it up to the user to install other tools as needed.

However, if it can be demonstrated that there is a strong demand for nix, then we will be more likely to merge this PR. One way to do that is to create a corresponding issue for visibility, and encouraging other developers that want nix to be installed by default in the Elastic CI Stack to vote for it with a 👍.

PS: is there any reason why this PR is still a draft? I also encourage you to squash your commits into a single one and force push to your branch, so we can merge it cleanly.

csakshaug:
Hi @triarius

Thanks for the nice feedback :-)

I was probably a little excited that we made it work and wanted to share the idea, but I look up in the branch and create a issue  like you suggest.

PR #1067, opened 2022-11-17
Add permissions to service role to allow deployment and deletion of encrypted s3 buckets with Bucket Policy

triarius:
This is necessary prior to merging https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1050.
Ideally we should rebase that PR over master and run a build on it again, however it's a contributed PR, so we may not expect that.
I have cherry-picked those commits and run a build on with the commits in this PR applied, and it seems to work: https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/2884.


dabarrell (approved):
Reviewed by committee

PR #1068, opened 2022-11-28
Make a parameter for buildkite-agent-scaler version

triarius:
Combined with https://github.com/buildkite/buildkite-agent-scaler/pull/74, this will enable launching stacks with custom versions of the scaler.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -46,6 +46,8 @@ Metadata:
         - BuildkiteTerminateInstanceAfterJob
         - BuildkiteAdditionalSudoPermissions
         - BuildkiteWindowsAdministrator
+        - BuildkiteAgentScalerServerlessARN
+        - BuildkiteAgentScalerVersion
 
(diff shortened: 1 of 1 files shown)


templates/aws-stack.yml
```diff
@@ -1232,7 +1237,7 @@ Resources:
     Properties:
       Location:
         ApplicationId: arn:aws:serverlessrepo:us-east-1:172840064832:applications/buildkite-agent-scaler
```
moskyb, line 1239: Is it worth parameterising `ApplicationId` as well if people want to use forks of the scaler?

(no is a perfectly acceptable answer)

triarius: Good call. I believe it's a bit complicated to set up their CI to publish to serverless repo, but I guess if they fork our repo they should be able to figure it out.

triarius:
Closing this, I have not been able to make buildkite-agent scaler work with an alternative repo

PR #1080, opened 2023-01-19
Clean up some install scripts

triarius:
These were things I noticed while working on another PR, but did not want them to get in the way of that more urgent work.

There are a few types of changes here:
- Version bumps
- Version consistency between linux and windows
- Script refactors
- Chmods

The git-lfs install script was rewritten as it seemed like of the steps were failing, but as this was being silently ignored because the script had not set `errexit`, I could not be sure. There is now a test that git-lfs can run, so the new version of the script should work if that test passes.


moskyb (approved):
🎉

PR #1087, opened 2023-01-30
correct invalid SSM policy action

jsleeio:
`ec2messages:SendRepl` isn't a valid policy action. We've been getting error messages about invalid IAM policy in some of our accounts. I don't have the inside story from AWS here but I wonder if they have started being more strict on these

https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-setting-up-messageAPIs.html

Note this has not been tested by me yet, but looks more correct than the current state :-)

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -913,7 +913,7 @@ Resources:
               - ec2messages:FailMessage
               - ec2messages:GetEndpoint
               - ec2messages:GetMessages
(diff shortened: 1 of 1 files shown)


moskyb (approved):
😬 whoops!

thanks so much for your contribution @jsleeio! I'll get this merged

robertogomez80 (approved):

Issue #1028, opened 2022-06-07
Customise EC2 instance name


ChrisLane:
**Is your feature request related to a problem? Please describe.**
I have several buildkite nodes representing several different queues but all of them are named "buildkite-agent" on the AWS EC2 page, this makes it hard to distinguish between them at a glance.

**Describe the solution you'd like**
I would like to be able to customise the EC2 instance name to something other than "buildkite-agent".

**Describe alternatives you've considered**
I am able to filter based on resource tags but that's more effort than having a more unique instance name.

**Additional context**
This seems to have been suggested in the past https://github.com/buildkite/elastic-ci-stack-for-aws/pull/237#issuecomment-273953273
I couldn't find any current solutions.

moskyb:
kia ora @ChrisLane! i feel your pain on this one, we have a good couple of stacks in our account and it's difficult and annoying to find the one you want sometimes.

We've been talking about this internally, and what we'd ideally like to do is just add the stack name to the instance name - ie `buildkite-agent-my-cool-elastic-stack-name-goes-here`. Unfortunately though, the past shackles us, and that's probably a pretty breaking change - other people could be monitoring all the instances called `buildkite-agent`, for instance.

A happy middle ground would be to add a param to customise the instance name per stack, which we can do without too much trouble. We'll put this on our backlog and get to it at some point, hopefully soon.

PR #1088, opened 2023-01-30
Add InstanceName parameter

fd-jonathanlinn:
Adds the ability to customise the Name tag on the EC2 instances spawned by the ASG, defaults to "buildkite-agent" for backwards compatibility.

Might want to add a MinLength/MaxLength to this, happy to do so based on preferences.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -83,6 +83,7 @@ Metadata:
         - InstanceRolePermissionsBoundaryARN
         - IMDSv2Tokens
         - EnableDetailedMonitoring
+        - InstanceName
 
       - Label:
           default: Auto-scaling Configuration
@@ -488,6 +489,11 @@ Parameters:
       - "true"
       - "false"
     Default: "false"
+
+  InstanceName:
+    Type: String
+    Description: Optional - Customise the EC2 instance Name tag
+    Default: "buildkite-agent"
 
 Rules:
   HasToken:
@@ -1061,7 +1067,7 @@ Resources:
                 - Key: Role
                   Value: buildkite-agent
                 - Key: Name
-                  Value: buildkite-agent
+                  Value: !Ref InstanceName
                 - Key: BuildkiteAgentRelease
                   Value: !Ref BuildkiteAgentRelease
                 - Key: BuildkiteQueue


templates/aws-stack.yml
```diff
@@ -489,6 +490,11 @@ Parameters:
       - "false"
     Default: "false"
 
+  InstanceName:
+    Type: String
+    Description: Optional - Customise the EC2 instance Name tag
+    Default: "buildkite-agent"
```
moskyb, line 496: 🤔 i wonder if we should abandon backwards compatibility here and have the default be 
```suggestion
    Default: !Ref 'AWS::StackName'
```

instead... given that instance names are non-unique already, they aren't going to be useful for addressing individual instances.

this would solve an issue that i have when i'm digging through the aws console sometimes where i look at this screen:
CleanShot 2023-01-31 at 10 10 07@2x

and it's impossible to know which instance is from which elastic stack (though it's pretty clear what `Pat's Scratch Box` is 😅)

this isn't a "fix this before we merge" comment; i'm very interested to hear your thoughts on this

moskyb:
re: min and max length, i think we're fine to leave it as is and have that end of things enforced by EC2 - i'm sure that there are reasonable limits in place there. Given that the failure dynamics are largely the same - the cfn stack won't launch in either case - i don't think there's much point in enforcing it on our end 🤷

again though, keen to hear your thoughts on this :)

templates/aws-stack.yml
fd-jonathanlinn: My only concern with not providing a default of the old value is what monitoring/automation others might have set up - doing it based on instance name isn't best practice but that doesn't mean it's not done, mostly taking guidance from https://github.com/buildkite/elastic-ci-stack-for-aws/issues/1028.  Happy to make it reference the StackName for the reasons you've given, also saves adding Parameter 😃

fd-jonathanlinn:
Yeah agreed on length we could remove those, was mostly trying to play it safe.  If we end up referencing StackName it's a moot point since it won't launch like you said

templates/aws-stack.yml
moskyb: oh man, contradicted by my past self 😅 on reflection i do think they're right though, we are indeed shackled by the past.

at some point in the medium term, we'll probably be releasing an elastic stack v6, that might be the right time to make the change, rather than shooting from the hip here.

moskyb (approved):

ChrisLane:
Thanks @fd-jonathanlinn! Somehow I didn't see your mention of my issue but I'm glad to see that I can do this now.

@moskyb could https://buildkite.com/docs/agent/v3/elastic-ci-aws/parameters be updated please?
I notice this option isn't mentioned in the docs.

Issue #1085, opened 2023-01-23
warning: unable to access '/etc/gitconfig': Is a directory


jdub:
**Describe the bug**

I get an error with git (even if I run it in a minimal command step) like this with v5.16.1:

```
$ git clone -v --recursive -- https://github.com/buildkite-plugins/docker-buildkite-plugin .
  | Cloning into '.'...
  | warning: unable to access '/etc/gitconfig': Is a directory
  | warning: unable to access '/etc/gitconfig': Is a directory
  | fatal: unknown error occurred while reading the configuration files
```

I have worked around this by removing the directory and replacing it with an empty file during bootstrap, e.g.

```bash
rm -r /etc/gitconfig || true
touch /etc/gitconfig
```

(The non-`-f` and `true` is just defensive, in case there's something in the directory, but so far there hasn't been.)

**Steps To Reproduce**

Run a git clone on the AMI provided by Elastic Stack v5.16.1

**Expected behavior**

A successful git clone. :-)

**Actual behaviour**

A failed git clone. :-)

**Stack parameters (please complete the following information):**
 - AWS Region: ap-southeast-2
 - Version: v5.16.1

triarius:
Hi @jdub. Thanks for using the Elastic CI Stack!

I'm not sure that either a file or directory `/etc/gitconfig` exists in the Elastic CI Stack AMI file system. I just booted up stacks for both v5.16.1 and the master branch and ran:
```
sh-4.2$ ls -lA /etc/gitconfig
ls: cannot access /etc/gitconfig: No such file or directory
sh-4.2$ sudo ls -lA /etc/gitconfig
ls: cannot access /etc/gitconfig: No such file or directory
sh-4.2$ sudo su buildkite-agent - -c 'ls -lA /etc/gitconfig'
ls: cannot access /etc/gitconfig: No such file or directory
```
in an AWS SSM session on an instance in each stack.

I am also able to git clone into an empty directory:
```
sh-4.2$ cd /tmp
sh-4.2$ mkdir test
sh-4.2$ cd test
sh-4.2$ git clone -v --recursive -- https://github.com/buildkite-plugins/docker-buildkite-plugin .
Cloning into '.'...
POST git-upload-pack (175 bytes)
POST git-upload-pack (gzip 2352 to 1199 bytes)
remote: Enumerating objects: 1712, done.
remote: Counting objects: 100% (188/188), done.
remote: Compressing objects: 100% (87/87), done.
[...]

PR #1089, opened 2023-02-02
Build git 2.39.1 with rpm

triarius:
We started to build git from source to update it to a version that was not vulrenable to [CVE-2022-23521](https://nvd.nist.gov/vuln/detail/CVE-2022-23521), as the version in the Amazon Linux 2 core repo was not (and is still not) updated to a non vulrenable version.

However, even though we followed the [official instructions to build it from source](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git#_installing_from_source), there were was a discripency with the expected path for the system git config was [surfaced](https://github.com/buildkite/elastic-ci-stack-for-aws/issues/1085#issuecomment-1408595131) to us. Essentially, configuring the build with the `/usr` prefix moves the expected path for the system config from `/etc/gitconfig` to `/usr/etc/gitconfig`.

While we tried to build it with a `/` prefix, that build failed: https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/2982 with the message `git: 'remote-https' is not a git command. See 'git --help'.` when trying to checkout a repo in tests.

Rather than debugging this, in this PR we download the RPM source for the git package from the Amazon Linux 2 repos and bump the version ourselves. This has been confirmed to preserve the system gitconfig path. There were a few wrinkles along the way:

1. Building with docs pulled in some x11 packages. We think its not appropriate to have these on a essentially headless system like the elastic-ci-stack. The usefulness of docs on such a system is also questionable, so we elected to build without docs.
2. While `rpmbuild` has a system to make parts of the build, like docs optional, this did not extend to the tool we use to install build dependencies, yum-builddep. So we were forced to remove the optional docs build dependencies. We did this with `sed` in a way that removes all optional docs blocks from the rpm. It should not be a problem as the would be ignored by `rpmbuild --without docs` anyway.
3. On arm64 instances, the build dep of `openssl-devel` (which resolved to `1:openssl-devel-1.0.2k-24.amzn2.0.4.aarch64`) would conflict with a previously installed `openssl11-devel` package. Since git should be able to compile when built with either version of openssl, we replace the `openssl-devel` build dep with `openssl11-devel`.

### DIFF
--- a/packer/linux/scripts/install-utils.sh
+++ b/packer/linux/scripts/install-utils.sh
@@ -22,20 +22,32 @@ sudo mv /tmp/conf/bin/bk-* /usr/local/bin
 echo "Configuring awscli to use v4 signatures..."
 sudo aws configure set s3.signature_version s3v4
 
-# https://git-scm.com/book/en/v2/Getting-Started-Installing-Git#_installing_from_source
 GIT_VERSION=2.39.1
-echo "Installing git $GIT_VERSION from source"
-sudo yum -y groupinstall "Development Tools"
+echo "Installing git $GIT_VERSION"
+sudo yum install -y rpm-build yum-utils
+sudo amazon-linux-extras install epel -y
+
 pushd "$(mktemp -d)"
-curl -sSL "https://mirrors.edge.kernel.org/pub/software/scm/git/git-${GIT_VERSION}.tar.xz" | tar xJ
-pushd git-"$GIT_VERSION"
-make configure
-./configure --prefix=/usr
-make all
-sudo make install
-git --version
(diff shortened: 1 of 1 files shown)


packer/linux/scripts/install-utils.sh
```diff
@@ -12,7 +12,6 @@ sudo pip install future
 sudo pip3 install future
 
 echo "Installing zip utils..."
-sudo yum update -y -q
```
triarius, line 15: There is a `yum update -y` earlier in this script (line 5).

DrJosh9000 (approved):
I hope this isn't needed for long!

PR #1090, opened 2023-02-02
Remove redundant yum update and make a script executable

triarius:
Some clean up I wanted to do while working on another PR.


packer/linux/scripts/install-utils.sh
```diff
@@ -12,7 +12,6 @@ sudo pip install future
 sudo pip3 install future
 
 echo "Installing zip utils..."
-sudo yum update -y -q
```
triarius, line 15: There is a yum update -y earlier in this script (line 5).

DrJosh9000 (approved):

PR #1095, opened 2023-02-21
Add support for c7gn, m7g, and r7g instance type classes to use the arm64 AMI

triarius:
Otherwise they will launched with the amd64 ami and the stack will fail to
launch and be rolled back.


DrJosh9000 (approved):
LGTM ✅

PR #1099, opened 2023-03-07
Explicitly Disable Public Access For Managed Secrets Buckets

triarius:
We don't encourage anyone to create objects in these buckets that have public ACLs.
However, I don't think there is a use case for public access for the entire bucket though, so let's disable it.


moskyb (approved):
LGTM! 🎉

Issue #998, opened 2022-03-23
FR: Update to Amazon Linux 2022


dhalperi:
**Is your feature request related to a problem? Please describe.**
Amazon Linux 2 is pretty old; defaults to Python 2.7, etc

**Describe the solution you'd like**
Upgrade to a newer Amazon Linux base. (Draft in preview ; support when it's ready?)

**Describe alternatives you've considered**

**Additional context**

edify42:
It's now amazonlinux2023 and it's in general availability: https://aws.amazon.com/blogs/aws/amazon-linux-2023-a-cloud-optimized-linux-distribution-with-long-term-support/

Issue #1096, opened 2023-02-21
Amazon Linux 2022


razbomi:
**Is your feature request related to a problem? Please describe.**
I am having trouble with using node 18 with the default buildkite agent AMI on elastic stack version v5.16.1. 

```
node: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by node)
node: /lib64/libc.so.6: version `GLIBC_2.28' not found (required by node)
```

This is because amazon linux 2 doesn't support the latest LTS for nodejs.
https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-up-node-on-ec2-instance.html

Amazon recommend using node 16 instead.

**Describe the solution you'd like**
Do you have any plans to upgrade to a newer version of [amazon linux](https://docs.aws.amazon.com/linux/al2022/release-notes/relnotes-20230118.html)?
I'm happy to raise a PR if it would be helpful.

**Describe alternatives you've considered**
I am running docker in docker which is causing slow builds and build failures for various reasons.
I have considered installing `glibc` in the the host bootstrap script.
I am happy to try any alternative suggestions.

**Additional context**
Build failure using the following [asdf plugin](https://github.com/byerobot/asdf-buildkite-plugin):

<img width="1593" alt="Screenshot 2023-02-21 at 8 59 06 pm" src="https://user-images.githubusercontent.com/1096424/220313126-f37c86a1-f4c8-4ea7-8167-d33c28cebc7b.png">

This is my `.tool-versions`:

```
terraform 1.0.10
terragrunt 0.35.5
nodejs 18.14.1
```

toothbrush:
We actually _just_ got bitten by this, and i was thinking the same (i could open a PR).  However, looking at the Packer config:

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/8022cb599c3693fa741d0f5d9d11cf3fbe153b91/packer/linux/buildkite-ami.json#L14

And searching for AMIs in AWS Console, it looks as if it will in fact select the newest version? 🤔 I'm a bit confused, i'd have expected to see a 5.15 kernel available at least, since that's what you get when you `docker run --rm -it amazonlinux`:

```shellsession
bash-4.2# uname -a
Linux 5c241319d407 5.15.49-linuxkit #1 SMP PREEMPT Tue Sep 13 07:51:32 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
```

[...]

PR #1103, opened 2023-04-01
Take a stab at upgrading to AL2023

toothbrush:
As folks have noted (https://github.com/buildkite/elastic-ci-stack-for-aws/issues/998, https://github.com/buildkite/elastic-ci-stack-for-aws/issues/1096), Amazon Linux 2023 is now generally available.  I thought i'd have a look and see what it'd take to use that as the base for Buildkite's AMI.  I haven't tested extremely thoroughly, but these modifications allow ARM64 and Intel AMI builds to run to completion successfully.

Gotchas:
* Had to pin more Python pip packages to get `docker-compose` (the old one) to successfully install on ARM64.  Could be brittle if the versions of those packages shipped in RPM change.
* QEMU isn't available on AL2023...  Could be a blocker?  I didn't bother trying to install from source yet.  If i get extremely bored i might.

Closes https://github.com/buildkite/elastic-ci-stack-for-aws/issues/998.


moskyb:
kia ora @toothbrush! cheers heaps for getting started on this - we already have an internal project for updating the stack to AL2023, and we hope to release it soonish.

unfortunately, having qemu on board is a hard requirement because we need it to be able to do cross-platform docker builds, which means that we'll probably end up having to install it from source :(

thanks so much for all the work you've already done - it'll probably make it into our project in some form or other (with attribution of course!) :)

PR #1104, opened 2023-04-03
Make a parameter for buildkite-agent-scaler edititon and version

triarius:
This will allow using a development version of the buildkite-agent-scaler with the elastic-ci-stack. To allow customers to use in all our supported regions, the pipeline needs to be run with the environment variable `COPY_TO_ALL_REGIONS=true`.


DrJosh9000 (approved):
LGTM

PR #1106, opened 2023-04-21
Fix error with docker experimental CLI

moskyb:
The elastic stack assumed that when docker experimental CLI was enabled, that there was already a docker config.json file. This meant that if one wasn't present, the elastic stack environment hook would fail every time

### DIFF
--- a/packer/linux/conf/buildkite-agent/hooks/environment
+++ b/packer/linux/conf/buildkite-agent/hooks/environment
@@ -10,8 +10,11 @@ BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY=$(mktemp -d)
 export BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY
 export DOCKER_CONFIG="$BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY"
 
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Nice

packer/linux/conf/buildkite-agent/hooks/environment
```diff
@@ -10,8 +10,11 @@ BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY=$(mktemp -d)
 export BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY
 export DOCKER_CONFIG="$BUILDKITE_DOCKER_CONFIG_TEMP_DIRECTORY"
 
-if [ "${BUILDKITE_DOCKER_EXPERIMENTAL:-false}" = "true" ]
-then
+if [ "${BUILDKITE_DOCKER_EXPERIMENTAL:-false}" = "true" ]; then
+  if [ ! -f "${DOCKER_CONFIG}/config.json" ]; then
```
pda, line 14: I prefer `[[` over `[` because it's less bonkers, but I see `[` was already used here so 🤷‍♂️

Issue #1108, opened 2023-04-21
Cloudformation stopped working - BucketOwnerEnforced


saviogl:
**Describe the bug**
As of April AWS changed their default values for newly created S3 buckets, and started disabling ACL (Access List Control) for them, in favor of Object Ownership method 

https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html 

This change breaks the existing stack as it is applying a pre-canned Access Control on the `ManagedSecretsLoggingBucket`

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/aws-stack.yml#L947

which then breaks the creation of the stack with the following messages:

```
The following resource(s) failed to create: [ManagedSecretsLoggingBucket, BuildkiteAgentTokenParameter, Vpc, Gateway, AsgProcessSuspenderRole]. Rollback requested by user.  
```

```
Bucket cannot have ACLs set with ObjectOwnership's BucketOwnerEnforced setting (Service: Amazon S3; Status Code: 400; Error Code: InvalidBucketAclWithObjectOwnership; Request ID: RCP1S0NK9DC0H3KA; S3 Extended Request ID: A+7biLiZTiYU9vl0O7+QcPHwX4lGt/FVhX6kv44WI1vjPyKIlDLomQClpxCnoc60EK9uj4mPA9o=; Proxy: null)
```

**Steps To Reproduce**
1. Apply the latest stack against an AWS account with the default configurations

```
aws cloudformation create-stack \
  --output text \
  --stack-name buildkite \
  --template-url "https://s3.amazonaws.com/buildkite-aws-stack/latest/aws-stack.yml" \
  --capabilities CAPABILITY_IAM CAPABILITY_NAMED_IAM CAPABILITY_AUTO_EXPAND \
  --parameters "$(cat config.json)"
```

**Expected behavior**
The stack and of its resources should be created without issues

**Actual behaviour**
The stack was unable to be created due to the new BucketOwnerEnforced default setting for AWS S3 and the pre-canned ACL configured in the stack

**Stack parameters (please complete the following information):**
 - AWS Region: [e.g. us-east-1]
 - Version [e.g. v5.6.1]

**Additional context**

PR #1109, opened 2023-04-21
fix: remove ACL from logging bucket

saviogl:
With the new defaults from AWS stated in https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html we shouldn't rely on ACLs nor there's a need to manually update public access configuration as these will now be **_ON_** by default. 

Removing these two configurations should leave the buckets with the recommended AWS security settings.

**This issue is blocking new clients to quickly bootstrap Buildkite in their account**

Fixes: #1108

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -944,6 +944,11 @@ Resources:
     Condition: CreateSecretsBucket
     DeletionPolicy: Retain
     Properties:
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
       BucketEncryption:
         !If
         - EnforceSecretsBucketEncryption
@@ -1013,6 +1018,11 @@ Resources:
         DestinationBucketName: !Ref ManagedSecretsLoggingBucket
       VersioningConfiguration:
         Status: Enabled
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
       Tags:
         - !If
           - UseCostAllocationTags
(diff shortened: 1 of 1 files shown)


triarius (changes requested):
Thanks for finding and fixing this @saviogl! 🫡

Unfortunately, new stacks won't be able to launch due to an "Invalid policy syntax. (Service: Amazon S3; Status Code: 400; Error Code: MalformedPolicy..."

I think the suggestion should fix it.

However, as we should get this fix out ASAP, I've made another PR (#1110) that incorporates your commit, and we will most likely merge that instead.

templates/aws-stack.yml
```diff
@@ -944,12 +944,6 @@ Resources:
     Condition: CreateSecretsBucket
     DeletionPolicy: Retain
     Properties:
-      AccessControl: LogDeliveryWrite
-      PublicAccessBlockConfiguration:
-        BlockPublicAcls: true
-        BlockPublicPolicy: true
-        IgnorePublicAcls: true
-        RestrictPublicBuckets: true
```
triarius, line 952: It also looks like the UI still shows a scary warning message if you don't have this:
2023-04-24T01:23:23,118602840+10:00

So I think we should undelete the `PublicAccessBlockConfiguration`

templates/aws-stack.yml
```diff
@@ -967,23 +961,39 @@ Resources:
...
         Id: ManagedSecretsLoggingBucketPolicy
         Version: "2012-10-17"
         Statement:
-        - Sid: AllowSSLRequestsOnly # AWS Foundational Security Best Practices v1.0.0 S3.5
-          Effect: Deny
-          Principal: '*'
-          Action: 's3:*'
+        - Sid: S3ServerAccessLogsPolicy # Grant permissions to the logging service principal using a bucket policy (https://docs.aws.amazon.com/AmazonS3/latest/userguide/enable-server-access-logging.html)
+          Effect: Allow
+          Principal: 'logging.s3.amazonaws.com'
```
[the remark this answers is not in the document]
saviogl: Shoot, I patched this locally and forgot to push upstream 👍

templates/aws-stack.yml
saviogl: @triarius Yeah, it seems that not all defaults have been applied yet! It's likely still rolling out, as the expectation was to have these being rolled out as default

saviogl:
@triarius Added back the Public Bucket policies block 👍 - With that said you've got the other PR up there too so up to you 😄

triarius (approved):
Thanks for getting back quickly @saviogl. So quick that my colleague's work day hadn't started for them to review #1110. But I can merge this instead now.

PR #1110, opened 2023-04-23
Fix ACLs can no longer be created for Managed Secrets Logging Bucket

triarius:
This reimplements #1109 with a fix to the `Principal` value. Thank you to @saviogl for finding and (very nearly) fixing the original issue!

Fixes: #1108


triarius:
Closing because #1109 was rectified before this was reviewed, entirely because of timezone differences.

PR #1117, opened 2023-05-10
Bumping python from 3.7 to 3.10

123sarahj123:
An old Python version was being used (3.7) for the AZ Rebalancing Lambda

This PR in particular is looking at [AZ Rebalancing Autoscaling Lambda](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/aws-stack.yml#L1295) - it was picked up that 3.7 is set as the Runtime

https://3.basecamp.com/3453178/buckets/23112918/card_tables/cards/6126980695#__recording_6127034819

Line 1295 to 1317:
https://github.com/buildkite/elastic-ci-stack-for-aws/blob/master/templates/aws-stack.yml#L1295


triarius (approved):
LGTM @123sarahj123 🎉. Can you undraft and merge this?

PR #1122, opened 2023-06-06
Upgrade Base Image to Amazon Linux 2023

triarius:
Thanks for bearing with us everyone, we have finally gotten around to baking Amazon Linux 2023 base images and getting them to run builds!

Although @toothbrush made a valiant effort in \#1103, their, and our previous attempts were DoA as the user data startup scripts were failing for various reasons. Thus, the instance would be marked as unhealthy soon after they booted and the ASG would continuously be booting new instances to replace them.

While fixing this, I took the opportunity to upgrade a few packages, and I've preferred to install these through the AL2023 repos as much as possible. Some older packages have been replaced where their functionality was provided by other packages that are available by default.

Consider adding these to the changelog at the time of release.

### Changed
- many packages have been added, upgraded or been removed since Amazon Linux 2. Refer to https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2.html for more details

### Upgraded
- aws-cli v2 (follows version in repo)
- openssl 3 (follows version in repo)
- qemu-user-static to v7.0.0
- amazon-ssm-agent (follows version in repo)
- python3 to 3.9 (follows version in repo)
- docker compose v2 to v2.18.3
- docker cli to 24.0.2 (follows version in repo)
   - Note: docker daemon remains at 20.10.23 (follows version in repo)
   - Note: the docker group has gid 993 now

### Removed
- python2
- cronie (replaced with systemd timers)
- docker-compose v1
- aws session-manager plugin

### CI
- packer installation scripts have been consolidated to be fewer
- goss upgraded to 0.3.23
- many more goss assertions, including the ability to run arm64 and amd64 docker images (through qemu-user-static or otherwise)


packer/linux/buildkite-ami.json
```diff
@@ -58,29 +54,11 @@
...
-      "type": "shell",
-      "script": "scripts/install-session-manager-plugin.sh"
-    },
-    {
-      "type": "shell",
-      "script": "scripts/install-nvme-cli.sh"
-    },
-    {
-      "type": "shell",
-      "script": "scripts/upgrade-kernel.sh"
```
triarius, line 77: Removed

packer/linux/buildkite-ami.json
```diff
@@ -58,29 +54,11 @@
...
-      "type": "shell",
-      "script": "scripts/install-git-lfs.sh"
-    },
-    {
-      "type": "shell",
-      "script": "scripts/install-session-manager-plugin.sh"
-    },
-    {
-      "type": "shell",
-      "script": "scripts/install-nvme-cli.sh"
```
triarius, line 73: Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/8752ac0ff6e4125b8231b83fc683939ef58bfc17/packer/linux/scripts/install-utils.sh#L20

packer/linux/scripts/install-nvme-cli.sh
triarius, line 1: Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/8752ac0ff6e4125b8231b83fc683939ef58bfc17/packer/linux/scripts/install-utils.sh#L20

packer/linux/scripts/install-s3secrets-helper.sh
triarius, line 1: Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/8752ac0ff6e4125b8231b83fc683939ef58bfc17/packer/linux/scripts/install-buildkite-utils.sh#L14-L18

packer/linux/scripts/install-lifecycled.sh
triarius, line 1: Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/8752ac0ff6e4125b8231b83fc683939ef58bfc17/packer/linux/scripts/install-buildkite-utils.sh#L20-L28

packer/linux/scripts/install-git-lfs.sh
triarius, line 1: Moved to Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/8752ac0ff6e4125b8231b83fc683939ef58bfc17/packer/linux/scripts/install-utils.sh#L30-L35

packer/linux/scripts/install-session-manager-plugin.sh
triarius, line 1: This isn't necessary, the session manager plugin is to allow logging into managed nodes with the aws-cli. A buildkite-agent typically (arguably ever) does not need to do that. It DOES need to be logged into by ssm, but the ssm-agent is for that. We install the ssm agent on https://github.com/buildkite/elastic-ci-stack-for-aws/blob/8752ac0ff6e4125b8231b83fc683939ef58bfc17/packer/linux/scripts/install-utils.sh#L15 and start it here: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/8752ac0ff6e4125b8231b83fc683939ef58bfc17/packer/linux/scripts/install-utils.sh#L15

packer/linux/conf/docker/subgid
```diff
@@ -1,2 +1,2 @@
-buildkite-agent:1001:1
+buildkite-agent:993:1
```
triarius, line 1: The docker rpm installed from the repos automatically creates a `docker` group with this gid.

packer/linux/conf/buildkite-agent/systemd/buildkite-agent.service
```diff
@@ -24,4 +24,3 @@ KillMode=process
 
 [Install]
 WantedBy=multi-user.target
-DefaultInstance=1
```
triarius, line 27: This is not needed as this unit is not used as a template and spits out a warning in journal

packer/linux/conf/docker/scripts/docker-gc
triarius, line 1: These are now systemd timers

packer/linux/conf/docker/scripts/docker-low-disk-gc
triarius, line 1: These are now systemd timers

packer/linux/conf/bin/bk-configure-docker.sh
```diff
@@ -1,27 +1,29 @@
...
 # Move docker root to the ephemeral device
-if [[ "${BUILDKITE_ENABLE_INSTANCE_STORAGE:-false}" == "true" ]] ; then
+if [[ "${BUILDKITE_ENABLE_INSTANCE_STORAGE:-false}" == "true" ]]; then
   cat <<< "$(jq '."data-root"="/mnt/ephemeral/docker"' /etc/docker/daemon.json)" > /etc/docker/daemon.json
 fi
 
 # Customise address pools
 cat <<<"$(jq '."default-address-pools"=[{"base":"172.17.0.0/12","size":20},{"base":"192.168.0.0/16","size":24}]' /etc/docker/daemon.json)" >/etc/docker/daemon.json
+
+systemctl restart docker
```
triarius, line 29: Need to restart docker to get the new settings.

packer/linux/conf/docker/scripts/docker-gc
triarius, line 1: These are now systemd timers

packer/linux/conf/docker/scripts/docker-low-disk-gc
triarius, line 1: These are now systemd timers

packer/linux/scripts/install-git-lfs.sh
triarius, line 1: Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/00e7a8b0dc6116b83fbe44b3330648e2f71a5afd/packer/linux/scripts/install-utils.sh#L30-L35

packer/linux/scripts/install-nvme-cli.sh
triarius, line 1: Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/00e7a8b0dc6116b83fbe44b3330648e2f71a5afd/packer/linux/scripts/install-utils.sh#L20

packer/linux/scripts/install-s3secrets-helper.sh
triarius, line 1: Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/00e7a8b0dc6116b83fbe44b3330648e2f71a5afd/packer/linux/scripts/install-buildkite-utils.sh#L14-L18

packer/linux/scripts/install-session-manager-plugin.sh
triarius, line 1: This isn't necessary, the session manager plugin is to allow logging into managed nodes with the aws-cli. A buildkite-agent typically (arguably ever) does not need to do that. It DOES need to be logged into by ssm, but the ssm-agent is for that. We install the ssm agent on https://github.com/buildkite/elastic-ci-stack-for-aws/blob/00e7a8b0dc6116b83fbe44b3330648e2f71a5afd/packer/linux/scripts/install-utils.sh#L15 and start it here: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/00e7a8b0dc6116b83fbe44b3330648e2f71a5afd/packer/linux/scripts/install-utils.sh#L28

packer/linux/scripts/install-docker.sh
```diff
@@ -33,31 +32,6 @@ sudo curl -Lfs -o /etc/systemd/system/docker.socket "https://raw.githubuserconte
...
-  echo 'cryptography<3.4' >"$CONSTRAINT_FILE"
-  echo 'urllib3<2' >"$CONSTRAINT_FILE"
-  sudo pip3 install --constraint "$CONSTRAINT_FILE" "docker-compose==${DOCKER_COMPOSE_VERSION}"
-
-	docker-compose version
-else
-  echo "No docker compose option configured for arch ${MACHINE}"
-  exit 1
-fi
-
```
moskyb, line 60: chefskiss

packer/linux/conf/docker/subgid
pda: My spidey senses tingle a bit with this change.
I vaguely remember some fragile things hanging off these `subgid` mappings.
But, it's probably fine.
As long as we test it with a rootless container? Maybe via Buildkite's docker plugin or something?

pda (commented):
I haven't read this thoroughly enough to add a green tick, but it's looking most excellent 👍🏼

packer/linux/buildkite-ami.json
triarius, line 73: Most of these have been folded into other scripts.

goss.yaml
```diff
@@ -100,34 +104,62 @@ process:
...
+    stdout:
+    - success
+
+  # Check docker plugins are installed
+  docker info --format=$'{{ `{{range .ClientInfo.Plugins}}{{if eq .Name "buildx"}}success\n{{end}}{{if eq .Name "compose" }}success{{end}}{{end}}` }}':
+    exit-status: 0
+    timeout: 30000
+    stdout:
+    - success
+    - success
```
toothbrush, line 141:
```suggestion
  # Check docker plugins are installed
  docker info --format=$'{{range .ClientInfo.Plugins}}{{if or (eq .Name "buildx") (eq .Name "compose")}}{{ .Name }},{{end}}{{end}}':
    exit-status: 0
    timeout: 30000
    stdout:
    - buildx,compose,
```

I don't know if this is any better, but i found the original to be a bit of a head-scratcher.  Also, this will tell you _which_ plugin is missing, if any.

WDYT?

goss.yaml
```diff
@@ -100,34 +104,62 @@ process:
...
-  # Checks that docker containers can run
-  "docker run --rm -v /var/run/docker.sock:/var/run/docker.sock docker:latest version":
+  /usr/local/bin/docker-low-disk-gc:
     exit-status: 0
-    timeout: 30000 # it can take some time to download the image
 
-  # Checks that permissions
-  'sh -c "docker run --rm -v \"$PWD:/pwd\" alpine:latest touch /pwd/test && stat -c %U:%G test"':
+  # Check docker userns is enabled
+  docker info --format='{{ `{{range .SecurityOptions}}{{if eq . "name=userns"}}{{"success"}}{{end}}{{end}}` }}':
```
toothbrush, line 129: Weirdly, on my machine, the nested ``{{ `{{ range`` constructions don't work.  How about something like this, perhaps a bit more intuitive?

```suggestion
  # Check docker userns is enabled
  docker info --format='{{range .SecurityOptions}}{{ if eq . "name=cgroupns" }}{{.}}{{end}}{{end}}':
```

goss.yaml
```diff
@@ -100,34 +104,62 @@ process:
...
     exit-status: 0
-    timeout: 30000 # it can take some time to download the image
+    timeout: 30000
+    stdout:
+    - success
```
toothbrush, line 133:
```suggestion
    stdout:
    - name=userns
```

Apologies for splitting up the suggestion – Github wouldn't let me put it on the whole block due to deleted lines 🤷

packer/linux/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -92,20 +90,20 @@ set_unless_present "AWS_DEFAULT_REGION" "$AWS_REGION"
...
-	curl -Lsf -o /usr/bin/buildkite-agent-edge \
-		"https://download.buildkite.com/agent/experimental/latest/buildkite-agent-linux-${AGENT_ARCH}"
-	chmod +x /usr/bin/buildkite-agent-edge
-	buildkite-agent-edge --version
+if [[ "${BUILDKITE_AGENT_RELEASE}" == "edge" ]]; then
+  if [[ "$(uname -m)" == "aarch64" ]]; then
+    AGENT_ARCH="arm64"
+  else
+    AGENT_ARCH="amd64"
+  fi
```
toothbrush, line 98: Elsewhere you've got a nice case statement, how about doing that consistently here?

```suggestion
  case $(uname -m) in
    x86_64)    ARCH=amd64;;
    aarch64)   ARCH=arm64;;
    *)         ARCH=unknown;;
  esac
```

(although this is very nitpicky, feel free to ignore me 😅)

packer/linux/scripts/install-lifecycled.sh
triarius, line 1: Moved to https://github.com/buildkite/elastic-ci-stack-for-aws/blob/90e22e4ea076dff965d2a02dd63e043afe00ba0a/packer/linux/scripts/install-buildkite-utils.sh#L20-L27

goss.yaml
triarius: Yeah, that's a great idea. I'll do that.

goss.yaml
triarius: So the nested templates are necessary because goss treats its yaml file as a template, so it would try to evaluate `.SecurityOptions` in the `--format` argument when it is meant to pass it through to docker.

I'll add a comment about this.

goss.yaml
triarius: Nw, I think the way GitHub expects it to work is that I batch them when committing.

packer/linux/conf/bin/bk-install-elastic-stack.sh
triarius: There's many more examples of sprinkled throughout the code base. I can't accept this as is thought, the variable was `$AGENT_ARCH`, but the suggestion has it as `$ARCH`. It's probably nicer to standardise on $ARCH if it does not exist in this script, though.

goss.yaml
toothbrush: Ah gotcha, so you're saying my suggestion works in my Terminal but wouldn't work in the GOSS yaml file?  Thanks for the explanation 👌

packer/linux/conf/bin/bk-install-elastic-stack.sh
toothbrush: Ah whoops, i missed that `$AGENT_ARCH`!  Glad someone is paying attention 😅

goss.yaml
triarius: Yeah, and mine has the opposite issue.

I could put the whole line in a nested template, that might be the best of both words.

goss.yaml
triarius: Hmm, I haven't actually been able to get goss to show the failing output, so you won't be able to tell which plugin is missing. There seems to be a fix in the works for goss v4 though: https://github.com/goss-org/goss/pull/576/files

But I found out that the elements of `stdout` are regexes and not expected lines. I think I can rewrite it with this in mind.

toothbrush:
Ah interesting, and yes i imagine that would be clearest - allowing a wholesale copy-pastable command to be easy to see. Thanks for indulging my nits, i am learning things about goss too 👍 On 7 Jun 2023, at 10:03, Narthana Epa ***@***.***> wrote:﻿
@triarius commented on this pull request.



In goss.yaml:
> +  # Check docker userns is enabled
+  docker info --format='{{ `{{range .SecurityOptions}}{{if eq . "name=userns"}}{{"success"}}{{end}}{{end}}` }}':

Yeah, and mine has the opposite issue.
I could put the whole line in a nested template, that might be the best of both words.

Issue #1121, opened 2023-06-06
git-mirrors only works as an experment


jim-barber-he:
**Describe the bug**
We have been using the git-mirrors expermient for ages now by passing the following parameter to the Buildkite Elastic CI Stack for AWS
```
EnableAgentGitMirrorsExperiment = true
```

As of version `5.21.0` of the stack, it installs agent version `3.47.0` which has promoted the `git-mirrors` experiment to stable.
Now our builds have the following warning:
```
The git-mirrors experiment has been promoted to a stable feature. You can safely remove the `--experiment git-mirrors`flag to silence this message and continue using the feature
```

So I removed the parameter above, but now the git-mirror is no longer being used.
The script that sets things up for git-mirrors only seems to do the work if the git-mirrors expeirment was enabled.
https://github.com/buildkite/elastic-ci-stack-for-aws/blob/90e22e4ea076dff965d2a02dd63e043afe00ba0a/packer/linux/conf/bin/bk-install-elastic-stack.sh#L129-L151

It seems the stack and script need updating to support new parameter(s) to enable the git-mirrors feature maybe?

**Stack parameters (please complete the following information):**
 - AWS Region: ap-southeast-2
 - Version 5.21.0

**Additional context**

Note that we have a bootstrap script that pre-creates the git-mirrors area on our RAM disk like so:
```
# Configure git-mirrors to use the RAM disk.                                                                             
BUILDKITE_GIT_MIRRORS_PATH="/var/lib/buildkite-agent/git-mirrors"                                                        
EPHEMERAL_GIT_MIRRORS_PATH="/mnt/ephemeral/git-mirrors"                                                                  
mkdir -p "$BUILDKITE_GIT_MIRRORS_PATH" "$EPHEMERAL_GIT_MIRRORS_PATH"                                                     
mount -o bind "$EPHEMERAL_GIT_MIRRORS_PATH" "$BUILDKITE_GIT_MIRRORS_PATH"                                                
echo "$EPHEMERAL_GIT_MIRRORS_PATH $BUILDKITE_GIT_MIRRORS_PATH none defaults,bind 0 0" >> /etc/fstab                      
chown buildkite-agent: "$BUILDKITE_GIT_MIRRORS_PATH" 
```
It basically copies what is in your `bk-install-elastic-stack.sh` script.
But the fact we're using that shouldn't really affect anything.

triarius:
[...]

PR #1123, opened 2023-06-06
 Rename the parameter `EnableGitMirrorsExperiment` to `BuildkiteAgentEnableGitMirrors`

triarius:
And no longer set the experiment in the agent config. Now that the
experiment has been mainlined, this feature is toggled by the non
emptiness of the `git-mirrors-path` config value. So all we need to do is
ensure that if the new parameter is true, it is set to a non-empty value
(and the path exists and is writable by the agent).

Fixes: #1121


DrJosh9000 (approved):
Good stuff

PR #1129, opened 2023-06-08
Publish to both `main` and `master` for the time being

triarius:
With v6, we will be renaming the default branch to main using the GitHub UI. When this change is made, the next build on the main branch won't push a stack template to the same location. So let's push to both locations for now.

### DIFF
--- a/.buildkite/steps/publish.sh
+++ b/.buildkite/steps/publish.sh
@@ -43,12 +43,28 @@ else
   echo "Skipping publishing latest, '$BUILDKITE_TAG' doesn't match '$(git describe origin/master --tags --match='v*')'"
 fi
 
-# Publish the most recent commit from each branch
-s3_upload_templates "${BUILDKITE_BRANCH}/"
+publish_for_branch() {
+  local branch="$1"
 
-# Publish each build to a unique URL, to let people roll back to old versions
-s3_upload_templates "${BUILDKITE_BRANCH}/${BUILDKITE_COMMIT}."
+  # Publish the most recent commit from each branch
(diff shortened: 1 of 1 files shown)


.buildkite/steps/publish.sh
```diff
@@ -43,12 +43,28 @@ else
...
 EOF
+}
+
+if [[ "$BUILDKITE_BRANCH" != "$BUILDKITE_PIPELINE_DEFAULT_BRANCH" ]]; then
+  publish_for_branch "$BUILDKITE_BRANCH"
+
+  exit 0
+fi
+
+# TODO: remove "master" from this list
```
DrJosh9000, line 66: This is largely a hypothetical, but if we all forget what this was about, when would the right time to remove "master" be? 6 months from now?

moskyb: i'm a big fan of time bomb comments for this - eg `# If you're reading this and the date is after YYYY-MM-DD the you should delete master from this list`, because it's a concrete call to action, and avoids both the bystander effect and the murky who/when/how of a TODO

moskyb: i remember ages ago finding a rust crate that was able to parse these and automatically fail CI if it found any hanging time bombs... lost to the sands of the internet now though

triarius: I did make a ticket about it: https://linear.app/buildkite/issue/PDP-1110/remove-publishing-elastic-stack-to-master

I put it in the backlog. I think moving to PDP triage might make it more visible.

PR #1130, opened 2023-06-09
Remove `SpotPrice` parameter and add parameter `SpotAllocationStrategy`

triarius:
We think using % of on demand and a spot allocation strategy is the more modern way of controlling the proportion of spot instances used. We have wanted to deprecate `SpotPrice` for some time now: https://github.com/buildkite/elastic-ci-stack-for-aws/issues/792

I've chosen to keep the default SpotAllocationStrategy to be `capacity-optimized` as it was in v5.

Make `capacity-optimized` the default spot allocation strategy

Amazon's recommended strategy is `price-capacity-optimized`: https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_InstancesDistribution.html but the default strategy for a InstancesDistribution is `lowest-price`.
In the elastic ci stack v5, we overrode the default and set it to `capacity-optimized`.

For a CI system, I think the strategy with the lowest risk of interruption is preferable.
There are two distinct reasons for this:
- Interrupting a CI job has a cost in developer time. Usually, developers are
  more expensive than instances.
- The instances carry state, e.g. the checkout dir and docker caches.
  Loosing this state costs time on subsequent jobs, which means those
  instance could be running for longer, and developers could be waiting
  for longer.

This is clearly a tricky balancing act, so giving users the freedom to
select a different strategy is a good idea, which is why I've decided to expose it as a parameter.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -92,6 +92,7 @@ Metadata:
         - MinSize
         - MaxSize
         - OnDemandPercentage
+        - SpotAllocationStrategy
         - ScaleOutFactor
         - ScaleInIdlePeriod
         - ScaleOutForWaitingJobs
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
Makes sense to me! 🥒

PR #1136, opened 2023-06-14
Don't (re)install docker (on Windows)

triarius:
The way we were installing Docker EE has stopped working. But, it seems to already be installed in the base AMI.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -51,6 +51,7 @@ steps:
     command:
       - git --version
       - docker info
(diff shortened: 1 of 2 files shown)


DrJosh9000 (approved):
Huzzah! ❇️

.buildkite/pipeline.yml
```diff
@@ -51,6 +51,7 @@ steps:
     command:
       - git --version
       - docker info
+      - docker run --rm mcr.microsoft.com/windows/nanoserver:ltsc2019 cmd.exe /c echo hello
```
sj26, line 54: Nice

sj26 (approved):
Nice work

PR #1139, opened 2023-06-20
Pull docker image used for qemu binfmt in packer

triarius:
We use a method documented by docker to install some tools necessary for building (and running) multiarch docker images. See https://docs.docker.com/build/building/multi-platform/

Previously we did this entirely in the startup script `packer/linux/scripts/install-docker.sh`. However, as was revealed in dogfooding, launching many instances at the same time could cause the startup script to fail due to docker hub rate limits. Instead, we can pull the image when the AMI is being built by packer, and then run the installation in the startup script.

Why not pull and run when building the AMI? Running multiarch images in CI failed when I tried to do that: https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/3275

### DIFF
--- a/packer/linux/scripts/install-docker.sh
+++ b/packer/linux/scripts/install-docker.sh
@@ -53,7 +53,9 @@ docker-compose version
 
 # See https://docs.docker.com/build/building/multi-platform/
 QEMU_BINFMT_VERSION=7.0.0-28
+QEMU_BINFMT_DIGEST=sha256:66e11bea77a5ea9d6f0fe79b57cd2b189b5d15b93a2bdb925be22949232e4e55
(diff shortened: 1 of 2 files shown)


packer/linux/scripts/install-docker.sh
```diff
@@ -50,3 +50,10 @@ docker compose version
 
 sudo ln -s "${DOCKER_CLI_DIR}/docker-compose" /usr/bin/docker-compose
 docker-compose version
+
+# See https://docs.docker.com/build/building/multi-platform/
+QEMU_BINFMT_VERSION=7.0.0-28
+sudo mkdir -p /usr/local/lib
+echo "QEMU_BINFMT_VERSION=${QEMU_BINFMT_VERSION}" | sudo tee -a /usr/local/lib/bk-install-elastic-stack.sh
```
triarius, line 57: We do this so that the version that is pullled is later run.

PR #1140, opened 2023-06-21
Allow the MaxSize to be 0

triarius:
This is useful for users that are doing blue-green deployments and want the
(eg blue) stack to be able to be scaled in to 0 instances while the green
stack is fully scaled out.


patrobinson (approved):
ok-i-like-it-picasso

PR #1148, opened 2023-07-02
Create script to prepend `--compatibility` flag to docker-compose

triarius:
Making /usr/bin/docker-compose a symlink or copy of the docker compose plugin binary is something that is increasingly popular. Indeed, this is now the only way to install a standalone `docker-compose` executable on Linux that's documented by docker: https://docs.docker.com/compose/install/standalone/#on-linux

So we did this in Elastic CI Stack for AWS v6 as well.

Unfortunately, this breaks certain workflows in the docker-compose plugin, which assumes that the `docker-compose` in the $PATH is docker-compose v1.

Arguably, the docker-compose plugin should default to v2 as v1 is EOL, but that is a breaking change. In the interim, we can prepend `--compatibility` to the docker-compose command to ensure that the plugin works as expected.

While I was in the area, I bumped the versions of docker compose and buildx.

### DIFF
--- a/packer/linux/conf/bin/docker-compose
+++ b/packer/linux/conf/bin/docker-compose
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-# Prepends --compatibility flag to arguments to docker-compose if not present
+# Prepends the --compatibility flag to arguments to docker-compose if not present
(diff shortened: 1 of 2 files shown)


DrJosh9000 (approved):
Makes sense to me!

PR #1149, opened 2023-07-03
Speculative fixes for docker image building hanging

triarius:
A lot of the PR is cleanup, but there are some speculative fixes for the failure of building of a docker image that was revealed during dogfooding.

I've added the building of that Docker image to the tests, and it clearly passes, but there may be some very rare flake involved.

The changes that may affect this are:

Installing multiarch qemu (via a docker container) has been moved from after the docker daemon has been configured to before, then all trace of the image that was used to do this is removed.

The files (`/etc/subuid`, `/etc/subgid`) needed for docker usernamespaceing are generated on first boot as I noticed that they seem to be modified dynamically after we copy them in. They should be dynamic as the values of the `buildkite-agent` user's UID and the `docker` group's GID are dynamic. It was a coincidence that the hard-coded values did not change between builds of the AMI before.


packer/linux/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -237,15 +233,7 @@ if ! docker ps; then
...
 fi
 
-# See https://docs.docker.com/build/building/multi-platform/
-echo Installing qemu binfmt for multiarch...
-docker run \
-  --privileged \
-  --userns=host \
-  --rm \
-  "tonistiigi/binfmt:${QEMU_BINFMT_TAG}" \
-    --install all
```
triarius, line 247: Now, after this is done, we restart the docker daemon.

triarius:
We've dogfooded this version for a bit, and we have a way forward. Let's merge it into v6 and tag v6.0.0-beta2

goss.yaml
```diff
@@ -140,6 +140,10 @@ command:
   /usr/local/bin/docker-low-disk-gc:
     exit-status: 0
 
+  docker buildx build -f tests/Dockerfile --progress=plain -t buildkite-postgres:latest tests:
+    exit-status: 0
+    timeout: 30000
+
```
triarius, line 146: This tests building a problematic container image

DrJosh9000: Worth adding this as a comment in the file?

DrJosh9000 (approved):

triarius:
@DrJosh9000 I'm going to beef up the tests based on what I've learnt since making this PR. Watch this space for another PR where I'll explain it better.

PR #1155, opened 2023-07-19
Install docker from binaries

triarius:
We previously installed docker from the Amazon Linux 2023 repos. However, that version is was 20.23.0. This installs 24.0.4 using binaries from the docker releases.


triarius:
We might keep this up ourselves, but relying on the docker rpm in the official repos seems like a better bet for long term stability.

PR #1159, opened 2023-07-21
Set `allow_dependency_failure: true` on stack cleanup jobs

triarius:
This will clean up the test stacks, even if the test fails.


sj26 (approved):
Nice

PR #1178, opened 2023-08-01
Bump CHANGELOG for v6.1.0

DrJosh9000:


DrJosh9000:
I'm thinking we should get #1173 into a release, but I reckon that makes it v6.1.0 - sound good?

triarius (approved):
LGTM

PR #1179, opened 2023-08-08
Fix `set -e` fails from env hooks

triarius:
In POSIX shells, `set -e` is not respected in certain contexts. One of these is: if it is in a function that is called on the left of a `||`.

See https://stackoverflow.com/questions/19789102/why-is-bash-errexit-not-behaving-as-expected-in-function-calls

We previously attempted to run this script in a function and exit status 2 if there was an error, but this was not working due to the above. Since it was not working, I took the liberty to change the exit code. Exit code 2 is too overloaded in my opinion.

I chose 53. This number looks like "S3", failure to network with which, could be the source of some transient failure of this hook.

I recommend review with whitespace changes hidden.

### DIFF
--- a/packer/linux/conf/buildkite-agent/hooks/environment
+++ b/packer/linux/conf/buildkite-agent/hooks/environment
@@ -1,144 +1,143 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
+set -Eeuo pipefail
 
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
LGTM 🥦

PR #1184, opened 2023-08-09
Add missing authorized keys systemd units

### LINKED ISSUE #1183: authorized keys causing termination loop in v6.1.0
**Describe the bug**
I have an authorized keys file configured and it looks to be downloading correctly but failing to configure a systemd timer, causing a crash loop:

image

**Stack parameters (please complete the following information):**
 - AWS Region: us-west-1
 - Version: v6.1.0
 - AuthorizedUsersUrl: s3://...

sj26:
Fixes #1183 — I don't think the systemd units responsible for refreshing authorized keys are being copied into the right place when the AMI is baked. This copies the approach of install-docker:

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/804061df748b42d3d40d5378892d84b2ff8e71ec/packer/linux/scripts/install-docker.sh#L19-L21

### DIFF
--- a/goss.yaml
+++ b/goss.yaml
@@ -8,6 +8,12 @@ file:
     owner: buildkite-agent
     group: buildkite-agent
 
+  /etc/systemd/system/refresh_authorized_keys.service:
+    exists: true
+
+  /etc/systemd/system/refresh_authorized_keys.timer:
+    exists: true
+
   /var/lib/buildkite-agent/builds:
     filetype: directory
     exists: true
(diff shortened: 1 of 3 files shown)


triarius (approved):
Thanks for finding and fixing this @sj26. LGTM

sj26:
I've added some tests. I think without some config changes the service might be run all the time, instead of only once enabled. I'd like to see if the goss spec passes.

sj26:
Nice, the tests pass:

https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/3869#0189d939-7c8c-4a2d-81df-8e7425d8d716/191-206

I'll install this version of the stack and verify it refreshes authorized keys.

sj26:
Confirmed — I plugged this AMI into our stack and it configures the keys correctly and enables the timer. I haven't waited to see if it fires yet, but this seems enough of an improvement to merge with confidence 🙌

PR #1192, opened 2023-08-14
fix: Add Missing Service Role Permissions

philnielsen:
We required these extra permissions when we made the switch to using the service role. I'm not sure what permutation of parameters added these extra requirements, but I don't think it would be unique for our org.

### DIFF
--- a/templates/service-role.yml
+++ b/templates/service-role.yml
@@ -181,7 +181,8 @@ Resources:
                 "lambda:RemovePermission",
                 "lambda:DeleteFunction",
                 "lambda:InvokeFunction",
-                "lambda:TagResource"
+                "lambda:TagResource",
+                "lambda:UpdateFunctionConfiguration"
               ],
               "Resource": "arn:aws:lambda:*:*:function:*"
             },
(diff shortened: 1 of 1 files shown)


triarius (commented):
Thanks, @philnielsen. We do test the creating stacks with service role on CI, but only on the default parameters.

I'm not sure what combination of parameters would cause this either, but just looking at the new permissions, it looks like they might have been used when upgrading an existing stack with (maybe with different parameters). Can you confirm that that's the case to narrow down my search?

This fine good to me, but I think we will have to verify this ourselves before merging, so anything you can do to narrow down the search space will speed this up.

philnielsen:
That is correct. the two logical IDs it was failing to update were `AutoscalingFunction` in the autoscaling nested stack and `AgentAutoScaleGroup` in the main stack when going from [v5.21.0](https://github.com/buildkite/elastic-ci-stack-for-aws/releases/tag/v5.21.0) to [v5.22.2](https://github.com/buildkite/elastic-ci-stack-for-aws/releases/tag/v5.22.2). I don't think any other parameters changed

triarius (approved):
Apologies for the delay, I verified each of the permissions added now. I also observed we needed to add:

```
"lambda:UpdateFunctionCode",
```

I'll make a PR for that too.

PR #1197, opened 2023-08-17
Prevent tag builds from publishing a latest template when they are not "on the main branch"

triarius:
While previously, the policy may have been different. I think it makes
sense to codify this policy:

Only builds for the latest stable tag on the main branch are published to:
- s3://buildkite-aws-stack/latest/aws-stack.yml
- s3://buildkite-aws-stack/aws-stack.yml

### DIFF
--- a/.buildkite/steps/publish.sh
+++ b/.buildkite/steps/publish.sh
@@ -24,17 +24,17 @@ echo "--- Building :cloudformation: CloudFormation templates"
 make build/aws-stack.yml
 
 echo "--- Uploading :cloudformation: CloudFormation templates"
-trunk="origin/$BUILDKITE_PIPELINE_DEFAULT_BRANCH"
-latest_tag=$(git describe --tags --abbrev=0 --match='v*' "$trunk")
+trunk="origin/${BUILDKITE_PIPELINE_DEFAULT_BRANCH}"
+latest_tag="$(git describe --tags --abbrev=0 --match='v*' "$trunk")"
 # Pre-release tags are those that have a hyphen in them, e.g. v1.0.0-rc1
-latest_stable_tag=$(git describe --tags --abbrev=0 --match='v*' --exclude='*-*' "$trunk")
+latest_stable_tag="$(git describe --tags --abbrev=0 --match='v*' --exclude='*-*' "$trunk")"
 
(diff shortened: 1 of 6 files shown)


.buildkite/steps/publish.sh
```diff
@@ -32,15 +24,19 @@ echo "--- Building :cloudformation: CloudFormation templates"
...
+latest_tag=$(git describe --tags --abbrev=0 --match='v*' "$trunk")
+# Pre-release tags are those that have a hyphen in them, e.g. v1.0.0-rc1
+latest_stable_tag=$(git describe --tags --abbrev=0 --match='v*' --exclude='*-*' "$trunk")
+
+# Only publish to 'latest' (and the empty prefix) if this tag is the latest stable tag.
+if [[ $BUILDKITE_TAG == "$latest_stable_tag" ]]; then
+  s3_upload_templates "latest/"
   s3_upload_templates
+elif [[ $BUILDKITE_TAG == "$latest_tag" ]]; then
+  echo "Skipping publishing latest, although $BUILDKITE_TAG matchs $latest_tag it does not doesn't match $latest_stable_tag"
```
DrJosh9000, line 37: Just the things I remember from https://google.github.io/styleguide/shellguide.html (which I should probably re-read)

(there's also a spelling fix in here)

```suggestion
trunk="origin/${BUILDKITE_PIPELINE_DEFAULT_BRANCH}"
latest_tag="$(git describe --tags --abbrev=0 --match='v*' "$trunk")"
# Pre-release tags are those that have a hyphen in them, e.g. v1.0.0-rc1
latest_stable_tag="$(git describe --tags --abbrev=0 --match='v*' --exclude='*-*' "$trunk")"

# Only publish to 'latest' (and the empty prefix) if this tag is the latest stable tag.
if [[ "${BUILDKITE_TAG}" == "${latest_stable_tag}" ]]; then
  s3_upload_templates "latest/"
  s3_upload_templates
elif [[ "${BUILDKITE_TAG}" == "${latest_tag}" ]]; then
  echo "Skipping publishing latest, although ${BUILDKITE_TAG} matches ${latest_tag} it does not doesn't match ${latest_stable_tag}"
```

DrJosh9000 (approved):

PR #1199, opened 2023-08-23
Change the Community Slack links to Forum ones

mcncl:
We're moving to the forum rather than Slack, so I'm adjusting the links before we do.


triarius (approved):
LGTM Thanks!

PR #1206, opened 2023-09-01
Fix instance storage mount script fails when instance storage not available

triarius:
It's understandable for it to fail, but the result is that users that have misconfigured their stacks to use instance storage on an instance type that doesn't support it will experiance flapping instances.

Now, we will allow them to misconfigure their stack. Maybe later, we can feed a warning into the job logs, but there is presently no mechanism to do so in the agent.


DrJosh9000 (approved):
Nice!

PR #1209, opened 2023-09-11
Disable idle timeout for scale-in protected instances

DrJosh9000:
Setting MinSize prevents AWS from scaling in below a number of agents. This doesn't prevent the agent terminating, which causes the terminate-instance script to run. That's fine, but... the agent can terminate for a variety of reasons, the most common being the idle timeout.

If MinSize > 0, and there are few jobs, then after ScaleInIdlePeriod (which becomes agent's disconnect-after-idle-timeout) the agent terminates (taking the instance with it). Then a new instance is spun up, it times out, and terminates. Lather, rinse, repeat. 

Arguably the intention of MinSize is to ensure some minimum capacity is always ready. So the idle timeout shouldn't apply to MinSize of the instances running.

### DIFF
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -65,6 +65,14 @@ PLUGINS_ENABLED=()
 [[ $ECR_PLUGIN_ENABLED == "true" ]] && PLUGINS_ENABLED+=("ecr")
 [[ $DOCKER_LOGIN_PLUGIN_ENABLED == "true" ]] && PLUGINS_ENABLED+=("docker-login")
 
+# Check whether this instance is protected from scale-in
+SCALE_IN_PROTECTION="$(aws autoscaling describe-auto-scaling-instances --instance-id "${INSTANCE_ID}" --query 'AutoScalingInstances[0].ProtectedFromScaleIn')"
+
+# If so, disable the idle timeout for the agent.
+if [[ "${SCALE_IN_PROTECTION}" = 'true' ]]; then
+  export BUILDKITE_SCALE_IN_IDLE_PERIOD=0
+fi
+
 # cfn-env is sourced by the environment hook in builds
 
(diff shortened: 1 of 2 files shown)


triarius (commented):
The condition `SCALE_IN_PROTECTION = "true"` will always be met because new instances are always protected from scale in: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/9c7dd0b86ad25c79778748a451f33753c2da671b/templates/aws-stack.yml#L1297

Thus, this PR will cause the scale in idle period to always be 0.

However, currently, the only mechanism by which the ASG scales in relies on BUILDKITE_SCALE_IN_IDLE_PERIOD > 0.

The rough sequence is
- ScaleInIdle period is exhausted
- Agent process terminates
- systemd runs the script specified in [ExecStopPost](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/9c7dd0b86ad25c79778748a451f33753c2da671b/packer/linux/conf/buildkite-agent/systemd/buildkite-agent.service#L17)
- this script terminates the instance and decrements the ASG desired capacity
- if the ASG is already at the min size, the decrement will be rejected, but the termination will still occur and the ASG launches a new instance.

Because this is the only way a stack can scale in, we can't disable this without first finding a different way to terminate instances.

But, if the behaviour you want is just that instances don't terminate when the stack is at Min capacity, then I recommend you modify the script in ExecPostStop to detect whether it should terminate the instance or not. If not, it might be able to restart the systemd unit.

This new behaviour should be set by a parameter too. Only users that want it should be made to use it. Some users may be unwittingly relying on the fact that their instances are cycling event when they have a non-zero min capacity.

triarius:
For future travellers, this bit is not quite correct:
> -   if the ASG is already at the min size, the decrement will be rejected, but the termination will still occur and the ASG launches a new instance.

Instead, the `ExecStopPost` script will error and systemd will restart the agent. The instance will survive this restart.

PR #1212, opened 2023-09-12
Prevent permission script acting on symlinks

DrJosh9000:
`chown -R` by default does not traverse symlinks, but we should prevent any of the path it is given from being a symlink.

Other changes:

- The target should not just exist, but only be a directory, so that's a new error.
- Docker container `bats/bats` has been updated in the last 7 years, `lucor/bats` not so much.


packer/linux/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions
```diff
@@ -71,15 +71,30 @@ exit_if_blank "${AGENT_DIR}"
...
-if [[ -e "${PIPELINE_PATH}" ]]; then
-	/bin/chown -R buildkite-agent:buildkite-agent "${PIPELINE_PATH}"
+# If it doesn't exist, then we won't do anything.
+if [[ ! -e "${PIPELINE_PATH}" ]]; then
+	exit 0
+fi
+
+
+# Check for symlink shenanigans
+if [[ "$(realpath "${PIPELINE_PATH}")" != "${PIPELINE_PATH}" ]]; then
```
triarius, line 88: Why not
```suggestion
if [[ -L "${PIPELINE_PATH}" ]]; then
```

triarius (approved):
LGTM

packer/linux/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions
DrJosh9000: That catches the final path segment being a symlink, but not the intermediate segments.

PR #1225, opened 2023-09-25
Correctly terminate single-job instances

moskyb:
There's a situation that can occur when using the TerminateAfterJob mode of the agent, where when the ASG gets down to its MinSize, after running a job, the agent will attempt to self-terminate using the `aws autoscaling terminate-instance-in-autoscaling-group` command, but becuase that action would cause the ASG to go below its minsize, the call fails. This fails the entire systemd `ExecPostStop` hook, which fails the entire unit, which causes systemd to restart the unit.

This means that in certain circumstances (when the instance attempted to self-terminate but the call failed), single-job instances can retain state from previous jobs, which is the opposite of what we want for these instances.

This PR changes the autoterminate script so that if the instance is in TerminateAfterJob mode, if the call to terminate fails, the instance will call `shutdown` on itself, causing the instance to be removed from the ASG no matter what EC2 has to say about it

### DIFF
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -131,6 +131,10 @@ echo Writing Phase 2/2 for /var/lib/buildkite-agent/cfn-env helper function...
 cat <<EOF >>/var/lib/buildkite-agent/cfn-env
 
 set_always         "BUILDKITE_AGENTS_PER_INSTANCE" "$BUILDKITE_AGENTS_PER_INSTANCE"
+
+# also set via /etc/systemd/system/buildkite-agent.service.d/environment.conf
+set_always         "BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB" "$BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB"
+
 set_always         "BUILDKITE_ECR_POLICY" "${BUILDKITE_ECR_POLICY:-none}"
 set_always         "BUILDKITE_SECRETS_BUCKET" "$BUILDKITE_SECRETS_BUCKET"
 set_always         "BUILDKITE_SECRETS_BUCKET_REGION" "$BUILDKITE_SECRETS_BUCKET_REGION"
@@ -357,6 +361,7 @@ done
 echo "Waited $next_wait_time times for docker to start. We will exit if it still has not started."
 check_docker
 
+# also set in /var/lib/buildkite-agent/cfn-env so that its shown in the job logs
 systemctl set-environment "BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB=${BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB}"
 echo Starting buildkite-agent...
 systemctl enable --now buildkite-agent
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -361,8 +361,17 @@ done
 echo "Waited $next_wait_time times for docker to start. We will exit if it still has not started."
 check_docker
 
-# also set in /var/lib/buildkite-agent/cfn-env so that its shown in the job logs
-systemctl set-environment "BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB=${BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB}"
+echo Writing buildkite-agent systemd environment override...
+# also set in /var/lib/buildkite-agent/cfn-env so that it's shown in the job logs
+mkdir -p /etc/systemd/system/buildkite-agent.service.d
+cat <<EOF | tee /etc/systemd/system/buildkite-agent.service.d/environment.conf
+[Service]
+Environment="BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB=${BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB}"
+EOF
+
+echo Reloading systemctl services...
+systemctl daemon-reload
+
 echo Starting buildkite-agent...
 systemctl enable --now buildkite-agent
 
--- a/packer/linux/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/conf/bin/bk-install-elastic-stack.sh
@@ -357,6 +357,7 @@ done
 echo "Waited $next_wait_time times for docker to start. We will exit if it still has not started."
 check_docker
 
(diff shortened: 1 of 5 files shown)


DrJosh9000 (commented):
Should we do a similar thing on the Windows side?

moskyb:
> Should we do a similar thing on the Windows side?

ooh, good call, have added now. i'm gonna do some smoke testing before merging, though

packer/windows/conf/bin/bk-install-elastic-stack.ps1
```diff
@@ -227,6 +228,9 @@ If ((![string]::IsNullOrEmpty($Env:BUILDKITE_ENV_FILE_URL)) -And (Test-Path -Pat
   }
 }
 
+nssm set buildkite-agent AppEnvironmentExtra BUILDKITE_TERMINATE_INSTANCE_AFTER_JOBUILDKITE_TERMINATE_INSTANCE_AFTER_JOB=$Env:BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB
```
DrJosh9000, line 231: Should it be this?

```suggestion
nssm set buildkite-agent AppEnvironmentExtra BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB=$Env:BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB
```

packer/windows/conf/bin/bk-install-elastic-stack.ps1
```diff
@@ -246,9 +250,9 @@ cfn-signal `
   --stack "$Env:BUILDKITE_STACK_NAME" `
   --resource "AgentAutoScaleGroup" `
   --exit-code 0 ; if (-not $?) {
-    # This will fail if the stack has already completed, for instance if there is a min size
-    # of 1 and this is the 2nd instance. This is ok, so we just ignore the erro
-    Write-Output "Signal failed"
-  }
+  # This will fail if the stack has already completed, for instance if there is a min size
+  # of 1 and this is the 2nd instance. This is ok, so we just ignore the erro
+  sm   Write-Output "Signal failed"
```
DrJosh9000, line 255: `sm`?

```suggestion
  Write-Output "Signal failed"
```

packer/windows/conf/bin/bk-install-elastic-stack.ps1
moskyb: no idea what happened here... fixed now though, as with `sm`

packer/linux/conf/buildkite-agent/scripts/terminate-instance
```diff
@@ -12,4 +16,15 @@ region=$(curl -H "X-aws-ec2-metadata-token: $token" --fail --silent --show-error
...
 
-aws autoscaling terminate-instance-in-auto-scaling-group --region "$region" --instance-id "$instance_id" "--should-decrement-desired-capacity"
+if [[ $BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB == "true" ]]; then
+  # If we're the final before the ASG's min size, the call to terminate-instance-in-autoscaling-group will fail, as AWS
+  # won't allow the ASG to go below its min size. In this case, we need to call shutdown instead and force the issue -
+  # the ASG will then spin up a new instance to replace the one we're shutting down, leaving all well in the world.
+  #
+  # We need to do this because if the call to terminate fails, the systemd unit will start up a new buildkite-agent process
+  # on this machine, with state left over from the last agent, which is the opposite of what we want when $BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB is true.
+  terminate "$region" "$instance_id" || shutdown
```
DrJosh9000, line 26: I think the default time for `shutdown` is 1 minute. `shutdown now`?

packer/windows/conf/bin/bk-install-elastic-stack.ps1
```diff
@@ -227,6 +228,9 @@ If ((![string]::IsNullOrEmpty($Env:BUILDKITE_ENV_FILE_URL)) -And (Test-Path -Pat
   }
 }
 
+nssm set buildkite-agent AppEnvironmentExtra BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB=$Env:BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB
+If ($lastexitcode -ne 0) { Exit $lastexitcode }
```
DrJosh9000, line 232: This line repeats the next one - I'm guessing that the existing one was checking the exit code from the `Get-Content` in the loop above. So maybe the lines should be swapped.

DrJosh9000 (approved):
Some more comments, but I think this is good 🥑

packer/windows/conf/bin/bk-install-elastic-stack.ps1
DrJosh9000, line 1: I commend your effort!

hckuo-anyscale:
when can we merge this?

packer/linux/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -260,6 +260,7 @@ check_docker() {
 check_docker
 
 # start buildkite-agent
+systemctl set-environment "BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB=${BUILDKITE_TERMINATE_INSTANCE_AFTER_JOB}"
```
triarius, line 263: This is going to make it into the job's env yeah? It seems like a useful thing for a job to know, so I'm not opposed to exposing it. But typically, env variables set by the elastic stack are disclosed in an environment hook that sources `/var/lib/buildkite-agent/cfn-env`.

Indeed, that file is the way for configuration set by template parameters to effect the behaviour of various scripts that run on the elastic stack.

You're using a different way of doing that here, which I generally think is better. But it only works for scripts run through systemd. 🤷‍♂️

I'll approve it if it's just addded to `cfn-env` though. That file is generated in https://github.com/buildkite/elastic-ci-stack-for-aws/blob/ee348e4976c9eb4696dc44c3b4f08e59f073059f/packer/linux/conf/bin/bk-install-elastic-stack.sh#L99-L100

templates/aws-stack.yml
```diff
@@ -1092,6 +1092,7 @@ Resources:
           IamInstanceProfile:
             Arn: !GetAtt "IAMInstanceProfile.Arn"
           InstanceType: !Select [ "0", !Split [ ",", !Join [ ",", [ !Ref InstanceTypes, "", "", "" ] ] ] ]
+          InstanceInitiatedShutdownBehavior: "terminate"
```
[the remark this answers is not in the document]
triarius: I think if we just make it conditional on the `TerminateInstanceAfterJob` parameter, it's good enough for now.

triarius (approved):

triarius (changes requested):
I've tested this out with by launching a real stack, and it does not work because the buildkite-agent user does not have permission to shut down the instance.

I'm exploring other options.

PR #1226, opened 2023-09-27
Add test stack remover script

moskyb:


utils/test-stack-remover.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+# The CI for this repo can sometimes leave some stacks lying around - every couple of months we need to clean them up.
+# This script will delete any stacks that are in DELETE_FAILED state, that contain the word _test_ in their name, and
+# that are not autoscaling stacks (these are child stacks that will be deleted when their parent is deleted)
+
+# no -u, as we rely on the $ROLE_ARN variable being passed in
```
triarius, line 7: Shouldn't this be the ideal use case for `set -u`? If the `ROLE_ARN` not in the env, it will error, if it is not, it will not. Then you're prevented from unintentionally invoking the aws cloudformation command with an empty argument.

I ran shellcheck locally on this with `set-u` and it didn't find any problems.

utils/test-stack-remover.sh
```diff
@@ -0,0 +1,17 @@
...
+set -eo pipefail
+
+aws cloudformation list-stacks \
+| jq -r ".StackSummaries[] | {StackName, StackStatus} | select(.StackStatus == \"DELETE_FAILED\").StackName" \
+| rg test \
+| rg -vi autoscaling \
+| sort | uniq \
+| xargs -t -I {} aws cloudformation delete-stack \
+  --stack-name {} \
+  --role-arn "$ROLE_ARN"
```
triarius, line 17: Unfortunately, shfmt will now ask that it look like:
```suggestion
  | jq -r ".StackSummaries[] | {StackName, StackStatus} | select(.StackStatus == \"DELETE_FAILED\").StackName" \
  | rg test \
  | rg -vi autoscaling \
  | sort | uniq \
  | xargs -t -I {} aws cloudformation delete-stack \
    --stack-name {} \
    --role-arn "$ROLE_ARN"
```

triarius: But a rebase is a good idea too.

utils/test-stack-remover.sh
moskyb: oh duh, onbiously 🤦

utils/test-stack-remover.sh
moskyb: fixed! i also changed the `rg` calls to `grep` cals - it's more likely to be installed

triarius (approved):
Thanks 💖

PR #1230, opened 2023-10-02
Improve logging for startup scripts on linux

triarius:
Prior to this PR, the startup scripts had tracing enabled (set -x) which made them difficult to read. This PR disables tracing and adds manual logging to the startup scripts.

I've tried to log before performing an action, and log if it fails. If a file is written to, I also try to log what was written. If you compare the output of these logs (the cloudwatch log group `/buildkite/elastic-stack/<instance-id>`) before an after this PR, the after has more information and is easier to follow (IMO).

#### Before
2023-10-02T22:30:17,887170592+11:00

#### After
2023-10-02T22:27:33,246787488+11:00

I've also removed some use of `>&2`, see https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1230/commits/e8184d8ff077abdce8cdd0ee2336105196762aed.

### DIFF
--- a/packer/linux/conf/bin/bk-configure-docker.sh
+++ b/packer/linux/conf/bin/bk-configure-docker.sh
@@ -17,7 +17,7 @@ on_exit() {
   echo "${BASH_SOURCE[0]} completed successfully."
 }
 
-trap on_exit EXIT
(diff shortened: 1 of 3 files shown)


moskyb (approved):
this is way better!

PR #1231, opened 2023-10-03
Fix qemu binfmt image is pulled during instance startup

triarius:
The intention was always for it to be pulled when the AMI is built, and just run when then instance is started, but it turns out the garbage collector was running before the docker setup script, so the image was being pulled on every instance start.

This PR fixes this by only enabling the timers after docker has been configured and the qemu binfmt image has run.

I've also added more comments around the mechanism by which the digest of the qemu binfmt image is passed from the packer build scripts to the startup script, as it was the source of some confusion. As well as more logging in general.

Closes: #1229

### DIFF
--- a/packer/linux/conf/bin/bk-configure-docker.sh
+++ b/packer/linux/conf/bin/bk-configure-docker.sh
@@ -81,7 +81,7 @@ if ! docker run \
   --userns=host \
   --rm \
   --pull=never \
-  "tonistiigi/binfmt:${QEMU_BINFMT_TAG}" \
+  "tonistiigi/binfmt@${QEMU_BINFMT_DIGEST}" \
     --install all; then
   echo Failed to install binfmt
   docker image ls
--- a/packer/linux/conf/bin/bk-configure-docker.sh
+++ b/packer/linux/conf/bin/bk-configure-docker.sh
@@ -42,7 +42,8 @@ if ! docker run \
   --pull=never \
   --rm \
   "tonistiigi/binfmt@${QEMU_BINFMT_DIGEST}" \
-    --install all; then
+    --install all
+then
   echo Failed to install binfmt.
   echo Avaliable docker images:
   docker image ls
--- a/packer/linux/conf/bin/bk-configure-docker.sh
+++ b/packer/linux/conf/bin/bk-configure-docker.sh
@@ -78,6 +78,7 @@ docker run \
   --privileged \
   --userns=host \
   --rm \
+  --pull=never \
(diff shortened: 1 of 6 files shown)


packer/linux/scripts/install-docker.sh
```diff
@@ -47,11 +45,22 @@ sudo cp /tmp/conf/bin/docker-compose /usr/local/bin/docker-compose
...
+echo Contents of /usr/local/lib/bk-configure-docker.sh:
+cat <<'EOF' | sudo tee -a /usr/local/lib/bk-configure-docker.sh
 QEMU_BINFMT_VERSION=7.0.0-28
 QEMU_BINFMT_DIGEST=sha256:66e11bea77a5ea9d6f0fe79b57cd2b189b5d15b93a2bdb925be22949232e4e55
 QEMU_BINFMT_TAG="qemu-v${QEMU_BINFMT_VERSION}@${QEMU_BINFMT_DIGEST}"
+EOF
+# shellcheck disable=SC1091
+source /usr/local/lib/bk-configure-docker.sh
 sudo mkdir -p /usr/local/lib
-echo "QEMU_BINFMT_TAG=\"$QEMU_BINFMT_TAG\"" | sudo tee -a /usr/local/lib/bk-configure-docker.sh
```
triarius, line 55: When images are pulled with a tag and digest, docker does not seem to keep the tag. Try it with 
```
$ docker pull alpine:3.18.4@sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978
$ docker image ls
REPOSITORY                   TAG             IMAGE ID       CREATED         SIZE
alpine                       <none>          8ca4688f4f35   4 days ago      7.33MB
```

I don't think it affects `docker run` though 🤷‍♂️.

packer/linux/conf/bin/bk-configure-docker.sh
```diff
@@ -25,11 +25,30 @@ trap '[[ $? = 0 ]] && on_exit' EXIT
...
 
+echo Installing qemu binfmt for multiarch...
+if ! docker run \
+  --privileged \
+  --userns=host \
+  --pull=never \
+  --rm \
+  "tonistiigi/binfmt@${QEMU_BINFMT_DIGEST}" \
+    --install all; then
+  echo Failed to install binfmt.
```
triarius, line 46:
```suggestion
    --install all
then
  echo Failed to install binfmt.
```

moskyb (approved):
awesome work 🎉

PR #1232, opened 2023-10-08
Add a step to CI to check files have been formatted with shfmt

triarius:
To reduce loss of productivity due to debates about the formatting of shell scripts, I've added a formatting check to the CI that uses [shfmt](https://github.com/mvdan/sh#shfmt) to detect script that are not formatted in accordance with it. This complements the existing checking of lints using [shellcheck](https://github.com/koalaman/shellcheck).

The options I've supplied to `shfmt` are:
- `--indent 2` to use 2 spaces for indentation. This matches our conventions for languages other than go.
- `--binary-next-line` which allows binary operators onto the next line
```
# good
echo hello world \
  | grep hello
echo hello world | grep hello

# bad
echo hello world |
  grep hello
```
If we didn't specify this, the first good example will be formatted to be like the first bad example, which I consider to be less readable as it requires reading to the end of the first line to see what the binary operator is, and is confusing as it does not use a line continuation at all.


packer/linux/conf/bin/bk-configure-docker.sh
```diff
@@ -42,8 +42,7 @@ if ! docker run \
   --pull=never \
   --rm \
   "tonistiigi/binfmt@${QEMU_BINFMT_DIGEST}" \
-    --install all
-then
+  --install all; then
```
triarius, line 45: This is a bit unfortunate, I would prefer i the formatter did not do this, but I don't consider it important enough to be fatal to the goal of having a formatter.

Luckily, to mitigate potential confusion between the end of the conditional and start of the body of the if statement, most text editors will highlight the `then` with coloured crayon technology.

DrJosh9000 (approved):
🥬

moskyb (approved):
🥳

PR #1234, opened 2023-10-12
Bump docker-compose to v2.22.0

jkburges:
Version v2.21.0 includes a "fix for incorrect proxy variables during build" - https://github.com/docker/compose/pull/10908. Thought I may as well go for the latest at time of writing.


triarius (approved):
Thanks! LGTM

PR #1235, opened 2023-10-12
Copy AMIs to ap-southeast-2 in branches

triarius:
It's easier to test on EC2 instances in ap-southeast-2 than in us-east-1 as most of our developers live in APAC.


triarius:
Looks like the AMIs are copied to ap-southeast-2 even if the same AMI was previously copied. We will need to find a solution to this before we merge, or it will lead to the # of ap-southeast-2 AMIs growing faster than the # of us-east-1 AMIs.

PR #1238, opened 2023-10-13
fix: wrap quotes around AWS::StackName

n-tucker:
## Purpose 🎯 
- Fixes an error encountered when we attempt to unmarshal the YAML template via our internal tooling
- After experimentation, I discovered that one I wrapped quotes around `AWS::StackName` our tooling was able to parse the template successfully

## Context 🧠 
- We perform some wrangling of CFN parameters via an internal CLI tool. As part of this process, the tool parses a YAML template to figure out which parameters are required as input
- Discovered this yaml parsing error when I was attempting to smoke test our upgrade to the `v6` elastic stack
- My best guess is that the yaml unmarshalling process doesn't like `::` unless it's quoted and/or escaped? 🤷

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -1130,7 +1130,7 @@ Resources:
                 - Key: Role
                   Value: buildkite-agent
                 - Key: Name
-                  Value: !If [ UseStackNameForInstanceName, !Ref AWS::StackName, !Ref InstanceName ]
+                  Value: !If [ UseStackNameForInstanceName, !Ref "AWS::StackName", !Ref InstanceName ]
                 - Key: BuildkiteAgentRelease
                   Value: !Ref BuildkiteAgentRelease
                 - Key: BuildkiteQueue


triarius (commented):
> My best guess is that the yaml unmarshalling process doesn't like :: unless it's quoted and/or escaped? 🤷

Maybe. I've tried out various parsers, and they DON'T seem to have issues with unquoted `::` in a yaml string:
javascript (js-yaml):
https://stackblitz.com/edit/node-qeqetm?file=index.js

ruby:
```shell
❯ ruby -ryaml -rjson -e 'puts YAML.safe_load("test: AWS::Stack").to_json'
{"test":"AWS::Stack"}
```

yq (go)
```shell
❯ echo 'test: AWS::Stack' | yq -o json
{
  "test": "AWS::Stack"
}
```

go (gopkg.in/yaml.v3):
https://goplay.tools/snippet/Mnv_QECRtZB

Can you let us know how you're parsing it? If it's common enough, we can add it to the CI pipeline to prevent future change from reintroducing the same problem elsewhere.

triarius (approved):
We can merge it and you can back to us about the tool @n-tucker.

PR #1239, opened 2023-10-17
Fix Windows AMI build failed

triarius:
The base AMI seems to have shifted a bit, it was failing to build. After debugging, it was found that the docker installation step required a restart, and the next part of the script (`docker --version`) errored out because of that. Even, then, the installation would not work until the restart was made. Eventually, we found this was because the docker install script check is checking if the windows feature called "Containers" is enabled and otherwise running
```powershell
Add-WindowsFeature Containers
```
This necessitates a restart, and the install script would not install docker until that happened.

So, in this PR, we use the [windows-restart](https://developer.hashicorp.com/packer/docs/provisioners/windows-restart) provisioner in packer to restart the builder instance after enabling containers and before installing docker. This did not seem to correctly wait for the previous provisioner, so I also added a 10s pause. The image building job takes ~21 min, so a 10s pause should not have a significant impact. The greater concern is if
```powershell
Add-WindowsFeature Containers
```
takes longer than 10s. I think this is unlikely, as it just toggles some windows configuration. If it starts becoming a source of build flake, we can look into debugging why the `windows-restart` provisioner is not waiting correctly.

Also, I:
- converted the packer json to hcl. This allows us to put comments in the packer file and hcl is easier to read than json (if you know it)
- added some tags to built AMIs


moskyb (approved):
fuck yes, this is so good 👏

.buildkite/pipeline.yaml
```diff
@@ -26,8 +26,6 @@ steps:
     name: ":go: fixperms build"
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
-    depends_on:
-      - "fixperms-tests"
```
triarius, line 30: It's faster if you build and test at the same time

PR #1240, opened 2023-10-18
fix: re-add rsyslog

peter-svensson:
Rsys is no longer installed on Amazon Linux 2023, see: [changelog](https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2.html#journald)

This re-adds the rsyslog daemon and starts it.


triarius (approved):
Thanks @peter-svensson

PR #1241, opened 2023-10-19
Tag Images with Agent and Elastic CI Stack Version

triarius:
The format of the tags will be:
```
Key=AgentVersion,Value=x.y.x
Key=Version:x.y.z,Value=true
```
Since the same AMI can be used for multiple Elastic CI Stack Versions, we can't just have one `Version` tag key and need to have a separate tag key for each version of the Elastic CI Stack.

The Agent Version is not like that and is known when the image is built by Packer. Note, however, that if users elect to use the "beta" or "edge" channels for agents, the version of the agent may differ from that of the tag.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -212,8 +212,8 @@ steps:
     artifact_paths: "build/*.yml"
     depends_on: "copy-ami"
 
-  - id: "cleanup"
-    name: "Cleanup"
+  - id: cleanup
+    name: ":broom: Cleanup"
     command: .buildkite/steps/cleanup.sh
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
--- a/Makefile
+++ b/Makefile
@@ -121,6 +121,7 @@ print-agent-versions:
 
 # Build linuxarm64 packer image
 packer-linux-arm64.output: $(PACKER_LINUX_FILES)
+	@echo Agent Version: $(CURRENT_AGENT_VERSION_LINUX)
 	docker run \
 		-e AWS_DEFAULT_REGION  \
 		-e AWS_PROFILE \
@@ -147,6 +148,7 @@ build/windows-amd64-ami.txt: packer-windows-amd64.output env-AWS_REGION
 
 # Build windows packer image
 packer-windows-amd64.output: $(PACKER_WINDOWS_FILES)
+	@echo Agent Version: $(CURRENT_AGENT_VERSION_WINDOWS)
 	docker run \
 		-e AWS_DEFAULT_REGION  \
 		-e AWS_PROFILE \
--- a/Makefile
+++ b/Makefile
@@ -111,6 +111,9 @@ build/linux-arm64-ami.txt: packer-linux-arm64.output env-AWS_REGION
 	mkdir -p build
 	grep -Eo "$(AWS_REGION): (ami-.+)" $< | cut -d' ' -f2 | xargs echo -n > $@
 
+CURRENT_AGENT_VERSION_LINUX ?= $(shell sed -En 's/^AGENT_VERSION="?(.+?)"?$/\1/p' packer/linux/scripts/install-buildkite-agent.sh)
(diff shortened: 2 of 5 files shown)


.buildkite/pipeline.yaml
```diff
@@ -212,8 +212,8 @@ steps:
     artifact_paths: "build/*.yml"
     depends_on: "copy-ami"
 
-  - id: "cleanup"
-    name: "Cleanup"
+  - id: cleanup
+    name: ":broom: Cleanup"
```
triarius, line 216: Life is better with emoji

Makefile
```diff
@@ -193,12 +205,9 @@ update-stack: build/aws-stack.yml env-STACK_NAME
...
 bump-agent-version:
-	sed -i.bak -E "s/\[Buildkite Agent v.*\]/[Buildkite Agent v$(AGENT_VERSION)]/g" README.md
-	sed -i.bak -E "s/AGENT_VERSION=.+/AGENT_VERSION=$(AGENT_VERSION)/g" packer/linux/scripts/install-buildkite-agent.sh
-	sed -i.bak -E "s/\\\$$AGENT_VERSION = \".+\"/\$$AGENT_VERSION = \"$(AGENT_VERSION)\"/g" packer/windows/scripts/install-buildkite-agent.ps1
-	rm README.md.bak packer/linux/scripts/install-buildkite-agent.sh.bak packer/windows/scripts/install-buildkite-agent.ps1.bak
-	git add README.md packer/linux/scripts/install-buildkite-agent.sh packer/windows/scripts/install-buildkite-agent.ps1
-	git commit -m "Bump buildkite-agent to v$(AGENT_VERSION)"
+	sed -iE "s/\[Buildkite Agent v.*\]/[Buildkite Agent v$(AGENT_VERSION)]/g" README.md
+	sed -iE "s/AGENT_VERSION=.+/AGENT_VERSION=$(AGENT_VERSION)/g" packer/linux/scripts/install-buildkite-agent.sh
+	sed -iE "s/\\\$$AGENT_VERSION = \".+\"/\$$AGENT_VERSION = \"$(AGENT_VERSION)\"/g" packer/windows/scripts/install-buildkite-agent.ps1
```
triarius, line 210: I removed the backups because they were deleted immediately anyway. 🤷‍♂️

I removed the automatic commit because I am against bots making commits on behalf of humans, especially locally. In practice, I often run this recipe when I'm unwittingly on the main branch. I then have to undo the commit, and apply it again on a branch.

DrJosh9000: I do that too!

moskyb: same! send it.

Makefile
```diff
@@ -193,12 +205,9 @@ update-stack: build/aws-stack.yml env-STACK_NAME
 AGENT_VERSION ?= $(shell curl -Lfs "https://buildkite.com/agent/releases/latest?platform=linux&arch=amd64" | grep version | cut -d= -f2)
 
 bump-agent-version:
-	sed -i.bak -E "s/\[Buildkite Agent v.*\]/[Buildkite Agent v$(AGENT_VERSION)]/g" README.md
-	sed -i.bak -E "s/AGENT_VERSION=.+/AGENT_VERSION=$(AGENT_VERSION)/g" packer/linux/scripts/install-buildkite-agent.sh
-	sed -i.bak -E "s/\\\$$AGENT_VERSION = \".+\"/\$$AGENT_VERSION = \"$(AGENT_VERSION)\"/g" packer/windows/scripts/install-buildkite-agent.ps1
-	rm README.md.bak packer/linux/scripts/install-buildkite-agent.sh.bak packer/windows/scripts/install-buildkite-agent.ps1.bak
-	git add README.md packer/linux/scripts/install-buildkite-agent.sh packer/windows/scripts/install-buildkite-agent.ps1
-	git commit -m "Bump buildkite-agent to v$(AGENT_VERSION)"
+	sed -iE "s/\[Buildkite Agent v.*\]/[Buildkite Agent v$(AGENT_VERSION)]/g" README.md
```
DrJosh9000, line 208: I think `sed -iE` still creates the backup (it will be named `README.mdE`)

triarius: Oh, good point.

triarius: Should be fixed now. I tested it by changing line 205 in the Makefile to have a fixed agent version.

DrJosh9000 (approved):
Sweet!

PR #1245, opened 2023-10-22
Mark last instance unhealthy if `BuildkiteTerminateInstanceAfterJob` is enabled

triarius:
This is a different implementation of the same goal as #1225, but we found when testing that PR that the instances would NOT terminate on linux because the buildkite-agent user does not have permission to shutdown the instance. Rather than granting it this permisison, in this PR we mark the instance as unhealthy if
1. The idle period has expired and the agent has shutdown.
1. The call to the AWS API to terminate the instance and decrement the desired capacity fails.
1. `BuildkiteTerminateInstanceAfterJob` is enabled.

Closes: #1225

### DIFF
--- a/packer/linux/conf/buildkite-agent/scripts/terminate-instance
+++ b/packer/linux/conf/buildkite-agent/scripts/terminate-instance
@@ -3,28 +3,50 @@
 set -euo pipefail
 
 terminate() {
-  aws autoscaling terminate-instance-in-auto-scaling-group --region "$1" --instance-id "$2" "--should-decrement-desired-capacity"
+  aws autoscaling terminate-instance-in-auto-scaling-group \
+    --region "$1" \
+    --instance-id "$2" \
+    --should-decrement-desired-capacity
+}
+
+mark_as_unhealthy() {
+  aws autoscaling set-instance-health \
+    --region "$1" \
+    --instance-id "$2" \
+    --health-status Unhealthy
 }
 
 echo "sleeping for 10 seconds before terminating instance to allow agent logs to drain to cloudwatch..."
-
 sleep 10
 
-token=$(curl -X PUT -H "X-aws-ec2-metadata-token-ttl-seconds: 60" --fail --silent --show-error --location "http://169.254.169.254/latest/api/token")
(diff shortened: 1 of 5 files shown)


packer/linux/conf/buildkite-agent/scripts/terminate-instance
```diff
@@ -2,14 +2,51 @@
...
+    --region "$1" \
+    --instance-id "$2" \
+    --should-decrement-desired-capacity
+}
+
+mark_as_unhealthy() {
+  aws autoscaling set-instance-health \
+    --region "$1" \
+    --instance-id "$2" \
+    --health-status Unhealthy
```
DrJosh9000, line 16: The ps1 includes `--no-should-respect-grace-period` - I think this should have it too, if the goal is to emulate `shutdown` (and where someone has configured a grace period).

DrJosh9000 (approved):

packer/linux/conf/buildkite-agent/scripts/terminate-instance
triarius: Looks like I copy-pasted that from your PR 😅.

The grace period is set by the [`HealthCheckGracePeriod`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-autoscaling-autoscalinggroup.html#cfn-autoscaling-autoscalinggroup-healthcheckgraceperiod) field of an `AWS::AutoScaling::AutoScalingGroup`, but that is not set here: https://github.com/buildkite/elastic-ci-stack-for-aws/blob/5b02d9550e3f8274028edecbfe504269718912ae/templates/aws-stack.yml#L1253-L1304

So the default of `0s` is what all stacks have by default. Given this, I think if the user has gone out of their way to set a grace period, we should not interfere.

DrJosh9000: I'm split on this. On the one hand, that's a fairly good argument for respecting the grace period. But on the other, why would they configure `BuildkiteTerminateInstanceAfterJob` at all (if they also want a grace period, in which to... recover the agent?)

triarius: IDK, but if explicitly did both, they may have thought of a reason we haven't.

PR #1252, opened 2023-10-26
Enable optionally changing EC2 Instance Types used for AMI Creation

tomowatt:
Not all EC2 Instance Types are available in each AWS Region, for example, in `eu-west-2` the `m7` EC2 Instance Types are not available - yet - so running the Makefile commands using the defaults will result in failing to launching the EC2 Instances for AMI Creation with the following error message:
```
Error launching source instance: Unsupported: The requested configuration is currently not supported. Please check the documentation for supported configurations.
```

Potential future usuablity improvement could be add some kind of check(s)?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,9 +9,9 @@ PACKER_WINDOWS_FILES = $(exec find packer/windows)
 
 AWS_REGION ?= us-east-1
 
-ARM64_INSTANCE_TYPE = m7g.xlarge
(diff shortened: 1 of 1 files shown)


triarius (approved):
LGTM @tomowatt,

My understanding is that customers don't need to run this Makefile. We build AMIs for them and make them publicly available for them to use. These are templated into cloudformation template that we release as well.

If they want to customise the AMI further, they would typically fork the repo, in which case they can change these instance types too.

PR #1253, opened 2023-10-26
add support for graviton3 with local nvme

joemiller:
Add the various graviton3 `*7gd` instance types (w/ local nvme instance storage)

I tried to spin up a queue with `c7gd.8xlarge` led to a failure due to the stack trying to use the amd64 instead of arm64 image. This should resolve that.

> CloudFormation stack, rollback requested (ROLLBACK_COMPLETE): ["The following resource(s) failed to create: [AgentAutoScaleGroup]. Rollback requested by user." "Resource handler returned message: \"You must use a valid fully-formed launch template. The architecture 'arm64' of the specified instance type does not match the architecture 'x86_64' of the specified AMI. Specify an instance type and an AMI that have matching architectures, and try again. You can use 'describe-instance-types' or 'describe-images' to discover the architecture of the instance type or AMI.


triarius (approved):
Thanks @joemiller!

PR #1254, opened 2023-10-30
Build fix-perms in Makefile

DrJosh9000:
In #1219 I overlooked the Makefile, which is used by both the pipeline (.buildkite/steps/packer.sh) and to build the stack without using the pipeline.
The result was that anyone just running `make clean && make packer` will be missing the new binary.

For now fix-perms is specific to the stack, so while one option is to split it into its own repo and generalise it and put a ribbon on it, the other is to make the Makefile responsible for building it, and using `docker` to run a container from the Makefile has precedent (hashicorp/packer).

### DIFF
--- a/.buildkite/steps/packer.sh
+++ b/.buildkite/steps/packer.sh
@@ -17,11 +17,11 @@ fi
 mkdir -p "build/"
 
 # Build a hash of packer files and the agent versions
-packer_files_sha=$(find Makefile "packer/${os}" plugins/ -type f -print0 | xargs -0 sha1sum | awk '{print $1}' | sort | sha1sum | awk '{print $1}')
-internal_files_sha=$(find go.mod go.sum internal/ -type f -print0 | xargs -0 sha1sum | awk '{print $1}' | sort | sha1sum | awk '{print $1}')
+packer_files_sha=$(find Makefile "packer/${os}" plugins/ -type f -print0 | xargs -0 sha256sum | awk '{print $1}' | sort | sha256sum | awk '{print $1}')
+internal_files_sha=$(find go.mod go.sum internal/ -type f -print0 | xargs -0 sha256sum | awk '{print $1}' | sort | sha256sum | awk '{print $1}')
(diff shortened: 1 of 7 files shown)


.buildkite/steps/packer.sh
```diff
@@ -16,17 +16,12 @@ fi
...
-  chmod 755 packer/linux/conf/buildkite-agent/scripts/fix-buildkite-agent-builds-permissions
-fi
-
 # Build a hash of packer files and the agent versions
 packer_files_sha=$(find Makefile "packer/${os}" plugins/ -type f -print0 | xargs -0 sha1sum | awk '{print $1}' | sort | sha1sum | awk '{print $1}')
+internal_files_sha=$(find go.mod go.sum internal/ -type f -print0 | xargs -0 sha1sum | awk '{print $1}' | sort | sha1sum | awk '{print $1}')
 stable_agent_sha=$(curl -Lfs "https://download.buildkite.com/agent/stable/latest/${agent_binary}.sha256")
 unstable_agent_sha=$(curl -Lfs "https://download.buildkite.com/agent/unstable/latest/${agent_binary}.sha256")
-packer_hash=$(echo "$packer_files_sha" "$arch" "$stable_agent_sha" "$unstable_agent_sha" | sha1sum | awk '{print $1}')
+packer_hash=$(echo "$packer_files_sha" "$internal_files_sha" "$arch" "$stable_agent_sha" "$unstable_agent_sha" | sha1sum | awk '{print $1}')
```
triarius, line 24: Maybe we shouldn't use sha1 in 2023. I guess someone with keys to our account could pre-build an malicious AMI, upload its sha1 to the s3 bucket and submit an innocuous PR that collides with the malicious AMI. But if they had the keys to build an AMI, they could just replace an existing released AMI with the malicious one. So I don't think there's a need to change this.

triarius (approved):
LGTM 🎉

.buildkite/steps/packer.sh
DrJosh9000: Changing the inputs to the hash seems like a good time to change the hash function.

PR #1258, opened 2023-11-03
Added BuildkiteAgentCancelGracePeriod option to linux stack

njgrisafi:
Following expectations from docs here https://buildkite.com/docs/agent/v3/configuration

We should allow users to configure the cancellation grace period.

### Added
- BuildkiteAgentCancelGracePeriod to `aws-stack.yml` (default 60)


DrJosh9000 (approved):

moskyb:
@njgrisafi i've done a wee bit of massaging + rebasing on this PR - just fixing a typo. thanks for your contribution!

moskyb:
worth noting: there's a difference between the windows and linux stacks here; linux stacks were defaulting to 60s for `cancel-grace-period`, but windows stacks didn't set it, which means that it would be the agent's default of 10s. i didn't wanna chuck it in there willy-nilly, so i'll add it in in another PR.

PR #1259, opened 2023-11-05
Ensure tag builds have the tag

triarius:
[This](https://buildkite.com/buildkite-aws-stack/buildkite-aws-stack/builds/4208#018b5fc8-d1a6-47ea-9c32-c6157f3aaba8) build was supposed to publish a "latest" template, but it did not because the tag did not exist in the local git repo.
I don't know exactly why it did not work (git mirrors is looking like a likely culprit), but we should error the tag build if the tag is not present.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -188,8 +188,6 @@ steps:
     command: .buildkite/steps/publish.sh
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
-    env:
-      BUILDKITE_GIT_FETCH_FLAGS: -v --prune --tags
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.yaml
```diff
@@ -188,8 +188,6 @@ steps:
     command: .buildkite/steps/publish.sh
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
-    env:
-      BUILDKITE_GIT_FETCH_FLAGS: -v --prune --tags
```
triarius, line 192: This does not work. The agent does not allow setting this environment variable.

DrJosh9000 (approved):
🫠

PR #1266, opened 2023-12-05
fix: allow hyphens in all `InstanceTypes` values

pH14:
The current `InstanceTypes` AllowedPattern allows hyphens for the first instance type listed, but not for the following ones. This is a tiny tweak to allow hyphens for all instance types listed.

Example failing case: `m7i.xlarge,m7i-flex.xlarge` will fail because the hyphen isn't the first type listed. The error message will be: `ValidationError: Parameter InstanceTypes failed to satisfy constraint: must contain 1-4 instance types separated by commas. No space before/after the comma.`

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -292,7 +292,7 @@ Parameters:
     Type: String
     Default: t3.large
     MinLength: 1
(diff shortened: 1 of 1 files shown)


triarius (approved):
Thank you, @pH14! LGTM

PR #1267, opened 2023-12-06
Launch elastic test elastic stacks using templates from S3

moskyb:
We were previously using templates as a cli param, but there's a 51KB limit on that, which we've recently hit


triarius (approved):
LGTM, thanks for taking it out of your other PR 💜

PR #1270, opened 2023-12-07
Add new values for ECRAccessPolicy that include `ecr:BatchImportUpstreamImage` 

triarius:
This IAM permission is necessary to use an ECR pull through cache. Technically, it is not `readonly`, as images will be written to the pull though cache on cache misses. Surprisingly, it is not in `poweruser` either. But we think customers will want to give Stacks the ability to use pull through caches without giving full permissions.

See https://docs.aws.amazon.com/AmazonECR/latest/userguide/pull-through-cache.html for more details.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -429,7 +429,7 @@ Parameters:
 
   ECRAccessPolicy:
     Type: String
-    Description: ECR access policy to give instances. The +pullthrough variants add `ecr:BatchImportUpstreamImage` which allows ECR pull through cache to work transparently.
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):

templates/aws-stack.yml
```diff
@@ -861,6 +868,16 @@ Resources:
...
       Policies:
+        - !If
+          - AddECRPullThrough
+          - PolicyName: ECRPullThrough
+            PolicyDocument:
+              Version: '2012-10-17'
+              Statement:
+                - Effect: Allow
+                  Action:
+                    - ecr:BatchImportUpstreamImage
```
yob, line 879: you may also want `ecr:CreateRepository` here. That permission is required for the first pull of a new image that's never been cached before. After that, `ecr:BatchImportUpstreamImage` is enough to pull new tags of the image into the repository.

DrJosh9000 (approved):

PR #1278, opened 2023-12-15
Stack: support configurable log retention days.

### LINKED ISSUE #1277: Support configuring the log retention for the autoscaling stack
**Describe the bug**

The stack creates a log group with a retention of 1 day, by nesting the serverless function https://github.com/buildkite/buildkite-agent-scaler/blob/master/template.yaml#L83-L86

```
LogRetentionDays:
    Type: Number
    Description: The number of days to retain the Cloudwatch Logs of the lambda.
    Default: "1"
```

This parameter can't be adjusted, and I think the only lift is to expose this option and forward it.

**Steps To Reproduce**

1. Enable the stack.
2. Check the log retention.

**Expected behavior**

Configurable log retention.

**Actual behaviour**

Static, 1 d retention.

**Stack parameters (please complete the following information):**
 - AWS Region: us-east-2
 - Version v6.12.0

**Additional context**
n/a

grahamc:
It comes from a general mandate to preserve logs generally for more time.

Close #1277 ... in case this is useful.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -49,6 +49,7 @@ Metadata:
         - BuildkiteWindowsAdministrator
         - BuildkiteAgentScalerServerlessARN
         - BuildkiteAgentScalerVersion
+        - LogRetentionDays
 
       - Label:
           default: Network Configuration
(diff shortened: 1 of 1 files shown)


triarius (approved):
LGTM, it will be useful for customer that have such a need. Thanks, @grahamc!

PR #1282, opened 2024-01-25
feat: Allow RootVolumeThroughput to be set for gp3 root volumes

### LINKED ISSUE #1026: Expose parameter for tuning EBS throughput
**Is your feature request related to a problem? Please describe.**
gp3 EBS volumes have a throughput parameter that can be configured. Currently this is not exposed in the elastic stack
When this is not set, it uses a default thruput 125MB/s which may not be sufficient for some heavy workloads.

**Describe the solution you'd like**
Add a new throughput parameter that is configurable when using gp3 EBS

cmanou:
We have been experiencing issues with `agent lost (-1)` after investigating with AWS support this turned out to happen when an instance with a gp3 root volume hits its throughput limit and becomes unresponsive. By exposing this people will be able to bump it and hopefully resolve their throttling issues. 

Resolves #1026

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -374,6 +374,11 @@ Parameters:
     Type: String
     Default: "gp3"
 
+  RootVolumeThroughput:
+    Description: If the `RootVolumeType` is gp3, the throughput to provision for the root volume
+    Type: Number
+    Default: 125
+
   RootVolumeIops:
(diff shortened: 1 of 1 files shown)


cmanou:
@moskyb Don't know if you were still looking at this one on your backlog but would appreciate if you could direct it to the correct reviewer that would be amazing

triarius (approved):
Thanks @cmanou LGTM

PR #1283, opened 2024-01-30
Allow specifiyng Iops for gp3

bradleyayers:
I want to adjust the Iops for a gp3 volume, but the current condition prevents `RootVolumeIops` from being passed through.


triarius (approved):
Thanks, @bradleyayers LGTM

PR #1286, opened 2024-01-30
Fix path for cfn-env on windows elastic stack

triarius:
`~` can resolve to `/c/Windows/system32/config/systemprofile` when we start the agent. We make sure it's in `/c/buildkite-agent/cfn-env` when we build the AMI, so let's reference it as being there.

I ran into this issue when doing some manual iteration on a Windows elastic stack. It should not be an issue outside of this, as there are integration tests that run jobs on Windows Elastic stacks.


DrJosh9000 (approved):
👍

PR #1298, opened 2024-03-04
Linux env hook was too eager, breaking ECR retries

lucaswilric:
When we introduced the `environment` hook error trap in #1179, we set it to propagate into all `source`-d plugin environment hooks. Unfortunately, this breaks the `retry` function in the ECR plugin's environment hook (which runs a command and then checks the exit code to determine whether to retry) by interrupting execution before the exit code can be checked.

Repro: https://github.com/buildkite-plugins/ecr-buildkite-plugin/pull/101

Fix by temporarily unsetting `-E`, stopping propagation of the trap command into the plugin environment hook.

- [ ] Before merging, add thank-you note to @-zzak and @-matthewd, who helped to diagnose and suggested the fix.

### DIFF
--- a/packer/linux/conf/buildkite-agent/hooks/environment
+++ b/packer/linux/conf/buildkite-agent/hooks/environment
@@ -95,8 +95,13 @@ if [[ "${BUILDKITE_ECR_POLICY:-}" != "none" && "${ECR_PLUGIN_ENABLED:-}" == "1" 
     export BUILDKITE_PLUGIN_ECR_ACCOUNT_IDS_0="${AWS_ECR_LOGIN_REGISTRY_IDS}"
   fi
 
+  # Do not propagate the error trap into the ECR plugin's environment hook, because its 'retry' function breaks when we do that.
+  set +E
+
   # shellcheck source=/dev/null
   source /usr/local/buildkite-aws-stack/plugins/ecr/hooks/environment
+
+  set -E
 fi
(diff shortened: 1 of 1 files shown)


packer/linux/conf/buildkite-agent/hooks/environment
```diff
@@ -95,8 +95,13 @@ if [[ "${BUILDKITE_ECR_POLICY:-}" != "none" && "${ECR_PLUGIN_ENABLED:-}" == "1"
     export BUILDKITE_PLUGIN_ECR_ACCOUNT_IDS_0="${AWS_ECR_LOGIN_REGISTRY_IDS}"
   fi
 
+  # Do not propagate the error trap into the ECR plugin's environment hook, because its 'retry' function breaks when we do that.
+  set +E
+
   # shellcheck source=/dev/null
   source /usr/local/buildkite-aws-stack/plugins/ecr/hooks/environment
+
+  set -E
```
triarius, line 104: Set `+E` is a nice fix! But I think you should do this in the ECR environment hook just before calling the `retry` function. Then, the other functions/substitutions/subshells of the ECR plugin will still cause this script to exit with status 53 if they fail, which is something the users that wanted this functionality would desire.

lucaswilric:
Superseded by buildkite-plugins/ecr-buildkite-plugin#102

PR #1332, opened 2024-06-26
Create dependabot.yml, update some Go things

DrJosh9000:
The go.mod file is for the internal fix-perms tool, which has a very narrow focus and very few dependencies, but strictly speaking it does have dependencies that might need updating from time to time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
--- a/.buildkite/docker-compose.yml
+++ b/.buildkite/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3'
 
 services:
   fixperms-tests:
-    image: golang:1.21
+    image: golang:1.22
     working_dir: /code
     environment:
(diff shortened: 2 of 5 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: gomod
+    directory: /
+    schedule:
+      interval: weekly
```
wolfeidau, line 6: if you add this you will only get security update PRs, rather than being spammed by every new version change.
```
    # only allow security updates for dependencies, not version updates  
    # https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file
    open-pull-requests-limit: 0
```

wolfeidau (changes requested):
Just a suggestion to ensure there is less noise from dependabot.

.github/dependabot.yml
DrJosh9000: I don't mind doing this, but there are only two dependencies in the go.mod file, I think we can afford to have a PR for every update?

wolfeidau (approved):
👍🏻

PR #1336, opened 2024-07-03
Fix 'Updating the Agent' page with link to guide.

wooly:
This page has been migrated into the Guide but still appears here, so I've added a breadcrumb to the new place.


DrJosh9000 (approved):
Thanks for this!

PR #1337, opened 2024-07-03
First pass at adding test-splitter to the elastic-ci linux stack.

wooly:
We're trying to make it as easy to get started with using the test splitter client as possible.

This PR adds the current release (v0.7.2) to the packer build for linux (we're not currently building a windows version).

### DIFF
--- a/packer/linux/scripts/install-buildkite-test-splitter.sh
+++ b/packer/linux/scripts/install-buildkite-test-splitter.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+case $(uname -m) in
+x86_64) ARCH=amd64 ;;
+aarch64) ARCH=arm64 ;;
+*) ARCH=unknown ;;
+esac
+
+TEST_SPLITTER_VERSION=0.7.2
+echo "Downloading test-splitter v${TEST_SPLITTER_VERSION}..."
+sudo curl -Lsf -o /usr/bin/test-splitter \
+  "https://github.com/buildkite/test-splitter/releases/download/v0.7.2/test-splitter-linux-${ARCH}"
+sudo chmod +x /usr/bin/test-splitter
+test-splitter --version
--- a/goss.yaml
+++ b/goss.yaml
@@ -139,6 +139,9 @@ command:
   make --version:
     exit-status: 0
 
+  test-splitter --version:
+    exit-status: 0
+
   wget --version:
     exit-status: 0
 
--- a/packer/linux/buildkite-ami.pkr.hcl
+++ b/packer/linux/buildkite-ami.pkr.hcl
@@ -109,6 +109,10 @@ build {
   }
 
   provisioner "shell" {
+    script = "scripts/install-buildkite-test-splitter.sh"
+  }
+
+  provisioner "shell" {
     script = "scripts/install-buildkite-utils.sh"
   }
 }


packer/linux/scripts/install-buildkite-test-splitter.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+case $(uname -m) in
+x86_64) ARCH=amd64 ;;
+aarch64) ARCH=arm64 ;;
+*) ARCH=unknown ;;
```
wooly, line 7: We don't currently have an unknown arch file being created. Can I snip this or should be we building one of those?

packer/linux/scripts/install-buildkite-test-splitter.sh
```diff
@@ -0,0 +1,19 @@
...
+esac
+
+echo "Creating test-splitter user and group..."
```
wooly, line 10: I'm curious as to whether this is required?

packer/linux/scripts/install-buildkite-test-splitter.sh
```diff
@@ -0,0 +1,19 @@
...
+sudo useradd --base-dir /var/lib --uid 2000 test-splitter
+sudo usermod -a -G docker test-splitter
+
+TEST_SPLITTER_VERSION=0.7.2
+echo "Downloading test-splitter v${TEST_SPLITTER_VERSION}..."
+sudo curl -Lsf -o /usr/bin/test-splitter \
+  "https://github.com/buildkite/test-splitter/releases/download/v0.7.2/test-splitter-linux-${ARCH}"
```
wooly, line 17: Are we happy to continue using github.com releases for this build step, or should we add a step to the test-splitter release pipeline to upload to s3 (download.buildkite.com) like the agent.

wooly (commented):
Few comments for review!

packer/linux/scripts/install-buildkite-test-splitter.sh
DrJosh9000: Very doubtful. The agent has its own user and group for privilege separation because it is a daemon (more or less) that runs arbitrary jobs. test-splitter seems like a tool that is run as needed by jobs run by the agent, so it would be acceptable to run under the same user and group as the agent.

packer/linux/scripts/install-buildkite-test-splitter.sh
DrJosh9000: If `$(uname -m)` doesn't match a case, currently the script will fail when `curl` receives a 404 Not Found. If this line is snipped then the script will fail on the same line with an undefined variable error instead. Which would you prefer to have when trawling through CloudWatch logs trying to diagnose problems?

packer/linux/scripts/install-buildkite-test-splitter.sh
DrJosh9000: I think there's a whole project in making all the various places (GitHub, S3, Buildkite Packages?) and packaging formats (raw, .zip, .deb, .rpm. ...) we provide binaries in consistent across all our projects. 

Since this only needs to work when packer builds the AMI, I think just GitHub is fine for now?

packer/linux/scripts/install-buildkite-test-splitter.sh
wooly: Turns out the buildkite-agent install wont work if the architecture is unknown also, since https://github.com/buildkite/elastic-ci-stack-for-aws/blob/main/packer/linux/scripts/install-buildkite-agent.sh#L21 this will resolve to https://download.buildkite.com/agent/stable/3.74.0/buildkite-agent-linux-unknown which doesn't exist.

packer/linux/scripts/install-buildkite-test-splitter.sh
wooly: Brill, thanks. Have snipped :+1:

DrJosh9000 (approved):
Thanks for doing this!

PR #1343, opened 2024-07-18
Update changelog for v6.22.2

wolfeidau:


DrJosh9000 (approved):
I think there's an argument for v6.23.0, but either way LGTM

PR #1366, opened 2024-09-06
Add the option to create a KMS key for signing of pipelines

wolfeidau:
This change adds the option to add a KMS key to the stack for signing and verifying of pipelines. Setting this to a keyspec will create the key and configure all agents to use this key.

Screenshot 2024-09-10 at 4 38 21 PM


DrJosh9000 (approved):
👍😎👍

PR #1370, opened 2024-09-12
Add signing verification behaviour and fix windows config option

wolfeidau:
This also fixes a misconfiguration on the windows agent. 

This setting is critical to rolling out signing as it is helpful during migration if you want to enable warn mode while updating pipelines.

> This setting determines the Buildkite agent's response when it receives a job without a proper signature, and also specifies how strictly the agent should enforce signature verification for incoming jobs. The agent will warn about missing or invalid signatures, but will still proceed to execute the job. If not explicitly specified, the default behavior is block, which prevents any job without a valid signature from running, ensuring a secure pipeline environment by default.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -40,6 +40,7 @@ Metadata:
         - PipelineSigningKMSKeyId
         - PipelineSigningKMSKeySpec
         - PipelineSigningKMSAccess
+        - PipelineSigningVerificationFailureBehavior
 
       - Label:
           default: Advanced Configuration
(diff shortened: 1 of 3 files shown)


DrJosh9000 (approved):

templates/aws-stack.yml
```diff
@@ -596,6 +596,14 @@ Parameters:
       - "verify"
     Default: "sign-and-verify"
 
+  PipelineSigningVerificationFailureBehavior:
+    Type: String
+    Description: The access level for the KMS key used to sign and verify pipelines
```
CerealBoy, line 601: I'm not sure if this description matches what the param does?

CerealBoy (approved):

DrJosh9000 (approved):
✅

templates/aws-stack.yml
wolfeidau: Thanks this is fixed 🤦🏻

PR #1372, opened 2024-09-16
Remove Test Splitter

nprizal:
We've decided to drop Test Splitter from Elastic CI Stack, considering the tool is not mature enough, and we now have multiple way to easily install the Test Splitter Client, such as via linux package manager (deb, rpm) and docker.


DrJosh9000 (approved):
✅

amybiyuliu (approved):
Noice!

niceking (approved):

PR #1373, opened 2024-09-16
fix: Add `r8g` to the `UsingArmInstances` logic

ParkMyCar:
As it says in the title, this PR adds the `r8g` EC2 instance class to the logic that's used to determine if the current instance is ARM based.


DrJosh9000 (approved):
Thanks, @ParkMyCar!

PR #1377, opened 2024-09-24
Pin amazon linux ami

patrobinson:
The new Amazon Linux AMI seems to have broken multi-platform docker image support.

```
[ec2-user@ip-172-31-57-192 ~]$ docker run --rm --platform linux/arm64/v8 -t arm64v8/ubuntu uname -m
exec /usr/bin/uname: exec format error
```

This pins the version for now


DrJosh9000 (approved):
👍🫠👍

PR #1380, opened 2024-09-30
Update the default queue name to match what's provisioned in a new org

amu-g:
By default, a new Buildkite org will have one cluster called Default, with a queue named `default-queue`. This PR updates the elastic ci stack default to match so that we get agents registered with the default queue out of the box.

This might be a breaking change for any orgs that have a queue named `default` that deploy the elastic stack without supplying a value for `BuildkiteQueue`.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -240,7 +240,7 @@ Parameters:
   BuildkiteQueue:
     Description: Queue name that agents will use, targeted in pipeline steps using "queue={value}"
     Type: String
-    Default: default
+    Default: default-queue
     MinLength: 1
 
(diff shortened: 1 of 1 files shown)


steveh:
related: https://linear.app/buildkite/issue/PIP-1198/default-queue-key-doesnt-match-queue-key-in-elastic-stack-default and https://linear.app/buildkite/issue/PIP-1209/the-default-queue-for-elastic-ci-aws-doesnt-match-the-default-queue-we

DrJosh9000:
This seems fine to me, but I'm curious how this PR compares with changing the default cluster's default queue name to be `default` instead?

Issue #1387, opened 2024-10-14
Linux AMI base is pinned to al2023-ami-minimal-2023.5.20240903.0


DrJosh9000:
In #1377 the base AMI was pinned to AL2023 release 20240903, since the newer version appeared to break running containers for other architectures. 

This issue should be resolved when the pin is removed.

DrJosh9000:
Fixed with the v6.31 release

PR #1390, opened 2024-10-28
fix: update goss to be more explicit with platform in docker tests

wolfeidau:


DrJosh9000:
Does this also happen to unblock updating the base AMI (which would fix #1387)?

DrJosh9000 (approved):

PR #1392, opened 2024-10-29
Add support for c8g instances.

toothbrush:
These new instance types https://aws.amazon.com/ec2/instance-types/c8g/ ought to be recognised as ARM64 workers.


DrJosh9000 (approved):
Thanks! LGTM

PR #1405, opened 2024-12-06
Update packer and the AMI with some fixes for binfmt

wolfeidau:
The AMI version we had pinned due to issues with multiarchecture docker builds failing was removed from list of public AMIs.
```bash
$ aws ec2 describe-images --filters 'Name=name,Values=al2023-ami-minimal-2023.5.20240903.0-*'
{
    "Images": []
}
```

This paired with a few changes which broke our packer build I have made a few changes:

1. Because we run update during the packer build, and a couple of the updates were 400Mb+ and filled up our 2Gb default ebs volume.
2. Added a command to cleanup downloads and cache files from dnf to keep the image lean after these upgrades...
3. Changes in systemd now remove cleanup binfmt_misc on restart https://github.com/systemd/systemd/issues/13129 so we need to add a `docker-binfmt.service` which runs at startup to support multiarchecture docker builds.
4. On top of this the proc-sys-fs-binfmt_misc.mount was being automounted on demand, again removing all the files we create via `docker-binfmt.service`. So I have enabled this mount to always activate on boot.
5. I have removed the pinned version in favour of the latest.

I also upgraded packer as it hadn't been touched in a while.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 VERSION = $(shell git describe --tags --candidates=1)
 SHELL = /bin/bash -o pipefail
 
-PACKER_VERSION ?= 1.9.4
+PACKER_VERSION ?= 1.11.2
 PACKER_LINUX_FILES = $(exec find packer/linux)
 PACKER_WINDOWS_FILES = $(exec find packer/windows)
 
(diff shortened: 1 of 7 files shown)


DrJosh9000 (approved):
Huzzah! Great effort!

PR #1410, opened 2024-12-11
Try harder to delete test S3 logging buckets created during CI

yob:
We create three CloudFormation Stacks on each build of the elastic stack, confirming instances boot and work as expected. There are specific steps in the build that attempt to delete all resources created for those tests, but from time to time they can fail.

This cleanup step runs at the end of the build and will delete any resources this pipeline might've left behind, regardless of whether they were created in this specific build or not.

I noticed that the AWS Account these test stacks were created in had many old S3 buckets like this:

* buildkite-aws-stack-test--managedsecretsloggingbuc-XXXXXXXXX

These are the logging buckets created by test stacks to holds access logs for the matching secrets S3 bucket. We attempt to delete them in `.buildkite/steps/delete.sh`, but some slip through the net.

This should help ensure we pick them up on a future build, and avoid the account growing an unbounded number of these buckets.

### DIFF
--- a/.buildkite/steps/cleanup.sh
+++ b/.buildkite/steps/cleanup.sh
@@ -2,6 +2,15 @@
 # shellcheck disable=SC2016
 set -uxo pipefail
 
+########################################################
+# We make an attempt to clean up the specific resources created during this pipeline elsewhere. However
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
LGTM

PR #1412, opened 2024-12-11
stop setting ACLs when publishing templates to S3

yob:
We recently changed the S3 bucket that holds public CloudFormation templates for the elastic stack from old fashioned ACLs to new fashioned Bucket Owner with Bucket Resource Policy that grants public read access.

Now, this script is failing with an error:

    upload failed: build/mappings.yml to s3://buildkite-aws-stack/secrets-2-2-0/mappings.yml An error occurred (AccessControlListNotSupported) when calling the PutObject operation: The bucket does not allow ACLs

I think removing the --acl flag from our publish command should fix that.

### DIFF
--- a/.buildkite/steps/publish.sh
+++ b/.buildkite/steps/publish.sh
@@ -4,8 +4,8 @@ set -euo pipefail
 s3_upload_templates() {
   local bucket_prefix="${1:-}"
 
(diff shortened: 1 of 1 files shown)


DrJosh9000 (approved):
LGTM

yob:
I confirmed this fixed publishing the templates to s3:

CleanShot 2024-12-12 at 12 29 16

PR #1414, opened 2024-12-17
New pipeline: deregister old AMIs

yob:
We've been publishing AMIs to 20 regions for ~6 years, and we've built quite a collection.

Many of them we want to keep indefinitely, like any that we've published in a released Cloud Formation template on [1]. Or any that were published in the last 12 months.

However, there's many many other images that it's safe to garbage collect. For example, we publish AMIs to every region for every main branch build, so we and customers can test them. They're useful for a short time, but any older than 12 months we can remove. By that time, customers have the option to move to a formal release with the features they want.

[1] https://github.com/buildkite/elastic-ci-stack-for-aws/releases

### DIFF
--- a/.buildkite/pipeline.cleanamis.yaml
+++ b/.buildkite/pipeline.cleanamis.yaml
@@ -1,0 +1,19 @@
+steps:
+  - name: ":broom: Delete AMIs ({{matrix}})"
+    command: .buildkite/steps/clean-old-amis
+    agents:
+      queue: "oss-deploy"
+    env:
+      DRY_RUN: true
+      AWS_REGION: "{{matrix}}"
+    matrix:
+      - "us-east-1"
+      - "us-west-2"
+      - "ap-southeast-2"
+    plugins:
+      - aws-assume-role-with-web-identity#v1.1.0:
+          role-arn: arn:aws:iam::172840064832:role/pipeline-buildkite-elastic-stack-for-aws-ami-cleaner
+      - docker-compose#v5.4.1:
+          run: ruby
+          config: .buildkite/docker-compose.yml
+          propagate-aws-auth-tokens: true
--- a/.buildkite/steps/clean-old-amis
+++ b/.buildkite/steps/clean-old-amis
@@ -1,0 +1,109 @@
+#!/usr/bin/env ruby
+
+require "bundler/inline"
+require "date"
+
+gemfile do
+  source "https://rubygems.org"
+
+  gem "oga" # an xml parser is required by aws-sdk
+  gem "aws-sdk-ec2"
+  gem "ostruct"
+  gem "logger"
+  gem "base64"
+end
+
+def die(msg)
+  $stderr.puts msg
+  exit 1
+end
+
+def log(msg)
+  puts "#{DateTime.now.iso8601} #{msg}"
+end
+
+region = ARGV[0] || ENV["AWS_REGION"]
+dry_run = ENV["DRY_RUN"]
+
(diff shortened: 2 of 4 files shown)


wolfeidau (approved):
👍🏻

.buildkite/pipeline.cleanamis.yaml
```diff
@@ -0,0 +1,19 @@
...
+    command: .buildkite/steps/clean-old-amis
+    agents:
+      queue: "oss-deploy"
+    env:
+      DRY_RUN: true
+      AWS_REGION: "{{matrix}}"
+    matrix:
+      - "us-east-1"
+      - "us-west-2"
+      - "ap-southeast-2"
```
yob, line 12: I'll add the other 17 regions in a follow up PR

.buildkite/steps/clean-old-amis
```diff
@@ -0,0 +1,110 @@
...
+  client.deregister_image({image_id: image.image_id}) unless dry_run
+
+  snapshot_ids.each do |snapshot_id|
+    puts "- #{tag} deleting snapshot #{snapshot_id}"
+    client.delete_snapshot({ snapshot_id: snapshot_id }) unless dry_run
+  end
+end
+
+one_year_ago = Time.now - (60 * 60 * 24 * 365)
+deleted_counter = 0
```
DrJosh9000, line 81: puts this at the end?

.buildkite/pipeline.cleanamis.yaml
```diff
@@ -0,0 +1,19 @@
+steps:
+  - name: ":broom: Delete AMIs ({{matrix}})"
+    command: .buildkite/steps/clean-old-amis
+    agents:
+      queue: "oss-deploy"
+    env:
+      DRY_RUN: true
```
yob, line 7: for now, we're only dry running it. We'll build confidence the output looks like we expect, then remove this

.buildkite/steps/clean-old-amis
```diff
@@ -0,0 +1,110 @@
...
+  gem "logger"
+  gem "base64"
+end
+
+def die(msg)
+  $stderr.puts msg
+  exit 1
+end
+
+MAX_DELETIONS = 10
```
DrJosh9000, line 21: Should this be an arg/env var? Wondering if this script would be useful to run in anger

DrJosh9000 (approved):

.buildkite/steps/clean-old-amis
yob: I was thinking of starting with 10, just to contain the blast radius of any early mistakes. Then increasing to something like 100 and letting it make progress via a scheduled build (daily? weekly?)

.buildkite/steps/clean-old-amis
```diff
@@ -0,0 +1,108 @@
...
+   all_images << image
+end
+
+# Filter the list of AMIs down to just those that were published by the elastic stack
+# pipeline. There might be other AMIs in this account, and we don't wantto mess with them
+all_images.select! { |image|
+  image.name.start_with?("buildkite-stack-") ||        # The name we used until mid 2019
+  image.name.start_with?("buildkite-stack-linux-") ||  # The name we used for linux amd64/arm64 from mid 2019
+  image.name.start_with?("buildkite-stack-windows-")   # The name we used for windows amd64 from mid 2019
+}
```
yob, line 43: I would prefer if we had a tag on the AMIs that we could use to find 100% of elastic stack images. Maybe `ElasticStack: true` or similar. This is a good start for now though

PR #1420, opened 2024-12-23
Fix PATH used by jobs

jonhiggs:
The PATH configured in `buildkite-agent.service` is inherited by all jobs executed by Buildkite.

Normally, a binary installed to `/usr/local/bin` would take precedence over similarly named files in `/usr/bin` or `/bin`. This commit restores that expectation allowing an administrator to replace a binary by installing a replacement into `/usr/local/bin`.

There is no entry for `/bin` because al2023 symlinks `/usr/bin` to `/bin`. A specific entry for `/bin` is unnecessary.

Finally, `/usr/local/sbin` and `/usr/sbin` have been removed because as the buildkite-agent is a regular user, it should not have the required permissions to use those binaries.


DrJosh9000 (approved):
LGTM!

PR #1424, opened 2024-12-31
Allow ARN to be provided for pipeline signing KMS key

MathewBoyles:
Alterative approach to https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1422 to ensure backwards-compatibility with existing stacks.

## Description

Our setup is not compatible with the current stack configuration. We have a KMS key in one AWS account which is shared across multiple accounts to sign and verify pipelines. This approach allows the user to specify an ARN of the KMS key so they may use a key cross-account or even cross-region. Existing behaviour is maintained as the Key ID may be provided using `PipelineSigningKMSKeyId` and the ARN is constructed using the current account and region. Key creation in the stack is also maintained.

### CHANGELOG

* Added `PipelineSigningKMSKeyArn` parameter to allow KMS key ARN to be provided.
* Updated "HasPipelineSigningKMSKey" rule to enforce only _one_ of key ID, ARN, or automatically generated key in stack.
* Updated references to `PipelineSigningKMSKeyId` to instead use `PipelineSigningKMSKeyArn` if provided, and construct ARN only if one is not provided and `PipelineSigningKMSKeyId` is
* Updated JSON object to YAML for readability

---

Co-authored-by: paul <paul.david@redbubble.com>

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -38,6 +38,7 @@ Metadata:
           default: Signed Pipelines Configuration
         Parameters:
         - PipelineSigningKMSKeyId
+        - PipelineSigningKMSKeyARN
         - PipelineSigningKMSKeySpec
         - PipelineSigningKMSAccess
         - PipelineSigningVerificationFailureBehavior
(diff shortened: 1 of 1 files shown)


MathewBoyles:
Closed in favour of https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1425

PR #1425, opened 2025-01-01
Let user provide full KMS Key ARN and update descriptions

toothbrush:
# Description

_(this is yet another approach to #1422 and #1424 - pick the one you prefer 😅)_

We have a use case where we want to have a single KMS key that Buildkite agents in multiple accounts can use to sign and verify pipeline steps. By letting the user specify the whole key ARN rather than just the key ID, the user is free to use a key wherever they want.

This third attempt has a different set of tradeoffs to the previous two:

* non-breaking change - this template will accept the same parameter as before, but the user can provide either the Key ID part or the full ARN, better than #1422's breaking change
* slightly less complex conditional and one less parameter - having to check that either-or Key ID or ARN are provided made #1424 a bit unwieldy

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -1389,10 +1389,7 @@ Resources:
                   PipelineSigningKMSKey: !If
                     - CreatePipelineSigningKMSKey
                     - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKey}
-                    - !If
-                      - PipelineSigningKeyIsARN
-                      - !Ref PipelineSigningKMSKeyId
-                      - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKeyId}
+                    - !Ref PipelineSigningKMSKeyId
               - !Sub
                 - |
                   Content-Type: multipart/mixed; boundary="==BOUNDARY=="
@@ -1467,10 +1464,7 @@ Resources:
                   PipelineSigningKMSKey: !If
                     - CreatePipelineSigningKMSKey
                     - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKey}
-                    - !If
-                      - PipelineSigningKeyIsARN
-                      - !Ref PipelineSigningKMSKeyId
-                      - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKeyId}
+                    - !Ref PipelineSigningKMSKeyId
 
   AgentAutoScaleGroup:
     Type: AWS::AutoScaling::AutoScalingGroup
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -627,12 +627,9 @@ Rules:
               - !Ref PipelineSigningKMSKeyId
               - ""
             - !Equals
-              - !Ref PipelineSigningKMSKeyARN
-              - ""
-            - !Equals
               - !Ref PipelineSigningKMSKeySpec
               - "none"
-        AssertDescription: "You must provide either a PipelineSigningKMSKeyId or PipelineSigningKMSKeyARN, or select a PipelineSigningKMSKeySpec, but not both"
+        AssertDescription: "You must provide either a PipelineSigningKMSKeyId, or select a PipelineSigningKMSKeySpec, but not both"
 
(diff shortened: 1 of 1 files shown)


templates/aws-stack.yml
```diff
@@ -1019,10 +1022,14 @@ Resources:
...
-                              - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKey}
-                              - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKeyId}
+                  Resource:
+                    !If
+                      - CreatePipelineSigningKMSKey
+                      - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKey}
+                      - !If
+                        - PipelineSigningKeyIsARN
+                        - !Ref PipelineSigningKMSKeyId
+                        - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKeyId}
```
wolfeidau, line 1032: This isn't neccessary as the API call in the Agent can take either KMS Key identifer or ARN. So there is no need to do anything special for the ARN https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/kms#SignInput 

And there is a precedent in the API for taking either in the same field, so you can remove the fancy ARN detection logic.

templates/aws-stack.yml
```diff
@@ -1432,12 +1452,25 @@ Resources:
...
+                    - UseCustomerManagedParameterPath
+                    - !Ref BuildkiteAgentTokenParameterStorePath
+                    - !Ref BuildkiteAgentTokenParameter
+                  PipelineSigningKMSKey: !If
+                    - CreatePipelineSigningKMSKey
+                    - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKey}
+                    - !If
+                      - PipelineSigningKeyIsARN
+                      - !Ref PipelineSigningKMSKeyId
+                      - !Sub arn:aws:kms:${AWS::Region}:${AWS::AccountId}:key/${PipelineSigningKMSKeyId}
```
wolfeidau, line 1473: As above.

wolfeidau (changes requested):
I prefer this solution as it matches the behaviour of the KMS APIs which also take either an KMS Key ID or KMS Key ARN.

templates/aws-stack.yml
toothbrush: Hey @wolfeidau, thanks for getting back to us.  You are quite right that the agent will have no trouble accepting either the Key ID or full ARN format.  We put this fancy conditional here because [to specify a KMS key in an IAM policy statement, you must use its key ARN](https://docs.aws.amazon.com/kms/latest/developerguide/cmks-in-iam-policies.html), that's my understanding of the documentation at least.

I'm definitely open to suggestions to simplify this stanza though.

templates/aws-stack.yml
toothbrush: We could remove the conditional here.  Although given what is described in https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1425#discussion_r1940374193, i still believe we'd need the `PipelineSigningKeyIsARN` check.

Slightly unrelated, but looking at the documentation at https://buildkite.com/docs/agent/v3/elastic-ci-aws/parameters#signed-pipelines-configuration i would suggest clarifying the exact format expected in `PipelineSigningKMSKeyId`.  E.g., do not provide `key/abcd-1234` but only `abcd-1234` - that kind of thing.

toothbrush:
I've pushed https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1425/commits/f1d74399ceede783bf01c1254b5e52ac11fe0391 and https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1425/commits/9e6f34ef7954c46c005a2df52d778139382a2928 – let me know if you think now.

wolfeidau (approved):

wolfeidau:
@toothbrush this looks great, I will wait for this build to go green and then merge it. 

I will work with the team to update the documentation as you suggested as well. 

Thanks for bearing with us.

wolfeidau:
@toothbrush this change can be tested using [refs/pull/1425/head/aws-stack.yml](https://s3.amazonaws.com/buildkite-aws-stack/refs/pull/1425/head/aws-stack.yml) if you were interested in sanity checking.

PR #1430, opened 2025-01-09
Add EBS volume tag for BK queue, keep optional cost allocation tags

petetomasik:
Adds EBS volume tags for `Name` and `BuildkiteQueue` that match the attached instance name, while also keeping optional cost allocation tags.


pzeballos (approved):
This is great! 👏🏻

PR #1432, opened 2025-01-16
Ec2 instance connect

blackliner:
ec2-instance-connect is a very useful tool to securely connect to your EC2 instances without having to open a port to the public internet

PR #1434, opened 2025-01-20
Revert "Merge pull request #1432 from blackliner/ec2-instance-connect"

mcncl:
## Changes
The addition of `ec2-instance-connect` seems to have changes the `gid` for Docker along the way, we should revert that change and test it locally instead to determine if we need to adjust a) test expected values, b) the `gid` given to one of those tools


mcncl:
Closing because https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1436 will fix the error

PR #1435, opened 2025-01-20
♻️ Ensures docker GID is 993 for tests

mcncl:
## Changes
Our GOSS tests rely on `docker` being in GID 993; https://github.com/buildkite/elastic-ci-stack-for-aws/blob/9965c106699eb60d938f81c813e81b9456b6a181/goss.yaml#L104

However, if anything is installed in the `install-utils` script it can/will overwrite this by taking that group and causing test failures.

Now, in `install-utils` we first add a group, `docker` and give it GID 993, tests should now pass and we ensure additional tools cannot take that group.


DrJosh9000:
I think the goss test should be fixed to accept a range of GIDs rather than trying to force `docker` to have a particular GID. What GID each group has should be an unimportant detail, and as you discovered it's easy to accidentally change them.

mcncl:
I'll close this in the interest of loosening our GID range requirements

PR #1436, opened 2025-01-20
➖ Docker can be a range of GIDs

mcncl:
## Issue
We've seen builds where the order of installs from the `install-utils` script can impact the `gid` of the `docker` group, but we specify that `docker` must be in `993` in our GOSS test

https://github.com/buildkite/elastic-ci-stack-for-aws/blob/9965c106699eb60d938f81c813e81b9456b6a181/goss.yaml#L104

The install order in our instance [build](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/9965c106699eb60d938f81c813e81b9456b6a181/packer/linux/buildkite-ami.pkr.hcl#L84) can determine the GIDs that packages get; `install-utils` runs first, so `993` may be taken by the time we install Docker.

## Change
We allow the `docker` group to have a GID between `990` and `995`, inclusive. We could expand this, but it _feels_ like a fine range to start with given we aren't aiming to add lots of packages.

### DIFF
--- a/goss.yaml
+++ b/goss.yaml
@@ -101,10 +101,6 @@ group:
 
   docker:
     exists: true
-    gid:
-      and:
-        gt: 989
-        lt: 996
 
   sshd:
     exists: true
@@ -217,3 +213,7 @@ command:
     timeout: 30000
     stdout:
     - x86_64
+
+  ## This allows us to support a range of group ids for the docker group, 990-995
+  'getent group docker | cut -d: -f3 | grep -E "^99[0-5]$"':
+    exit-status: 0
--- a/goss.yaml
+++ b/goss.yaml
@@ -42,12 +42,12 @@ port:
   tcp:22:
     listening: true
     ip:
-      - 0.0.0.0
+    - 0.0.0.0
 
   tcp6:22:
     listening: true
     ip:
-      - "::"
+    - "::"
 
 service:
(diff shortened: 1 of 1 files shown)


goss.yaml
```diff
@@ -214,3 +213,7 @@ command:
     timeout: 30000
     stdout:
     - x86_64
+
+  ## This allows us to support a range of group ids for the docker group, 990-995
+  'getent group docker | cut -d: -f3 | grep -E "^99[0-5]$"':
```
DrJosh9000, line 218: Having a quick look at the goss docs, I think you should be able to use a numeric matcher on `gid`. They have an example for users, surely groups would work similarly? (https://goss.readthedocs.io/en/stable/quickstart/#manually-editing-goss-files):

```
user:
  sshd:
    ...snip...
    exists: true
    uid:
      # Validate that UID is between 50 and 100
      and:
        gt: 50
        lt: 100
```

mcncl: Ah nice, I didn't see that in the configuration guide so let's see if the build errors when using it.

mcncl: Looks like that's not an option on `group`.

DrJosh9000: Darn! Was worth a shot I guess.

DrJosh9000 (approved):
LGTM ✅

Issue #882, opened 2021-07-23
Configure OOM killer in systemd to preserve the agent process


keithduncan:
We want the agent process to remain alive but the bootstrap and any processes under it in the systemd service should be candidates for being killed under memory pressure 🤔

n-tucker:
We're also experiencing this when instances running in our elastic stack run low on memory. This leaves our instances in a zombie state as they seem to occupy capacity in our autoscaling group without processing any jobs. Sometimes the instances are cleaned up automatically but we've seen instances alive for close to a week with no agents.

While we can reduce the number of agents per instance and/or increase the instance class we'd like the agents to stay alive in some form, so bumping this issue! 😄

PR #1450, opened 2025-03-07
fix: make sure buildkite-agent.service is not terminated by OOM

scadu:
Fixes https://github.com/buildkite/elastic-ci-stack-for-aws/issues/882

Set OOMScoreAdjust to -1000 preventing buildkite-agent processes from being terminated by OOM killer. 
https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#OOMScoreAdjust=


petetomasik (approved):
LGTM! The `buildkite-agent` process should be able to survive moments of memory pressure when tests/builds/etc consume excess resources on an ASG instance in order to be able to successfully report back job status to the Agent API.

pzeballos (approved):
nice!

PR #1456, opened 2025-03-19
⬆️ Bump Windows server to 2022

mcncl:
## Changes

- Bump Windows Server version to `2022` in the Packer file for Windows
- Adjust the README to show the new version of Windows Server which is supported
- Update [Session Manager](https://docs.aws.amazon.com/systems-manager/latest/userguide/plugin-version-history.html) plugin to 
1.2.707.0
- Update docker compose to 2.35.1

## Testing

- Testing is done in CI, however we can spin up a stack using this AFAIK to test out the new version beind used

## Revert

If things break then we can revert, I guess, but the version needs to be bumped, so if we do revert then an immediate fix should be considered


petetomasik (approved):
🚀

PR #1462, opened 2025-04-07
Update to latest aws lambda runtime

lswith:
This PR updates the `AzRebalancingSuspenderFunction` lambda to use the latest lambda runtime (currently python3.13). 

https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html


toote (approved):
Tested it and the function works with the new runtime and no further modifications :)

Screenshot 2025-05-16 at 10 00 43

PR #1516, opened 2025-07-11
Add script for generating changelog

scadu:
Automate generating CHANGELOG.md with automatically fetching agent’s release notes.

Required tools: ghch
Sample usage: 
```
GITHUB_TOKEN=$(op read "op://MyGitHub/token") bash .buildkite/scripts/generate-changelog.sh v6.40.7
```

`op read` comes from 1Password CLI. See https://developer.1password.com/docs/cli/reference/commands/read/.

### DIFF
--- a/.buildkite/scripts/generate-changelog.sh
+++ b/.buildkite/scripts/generate-changelog.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+set -euo pipefail
+
+if [[ -z "${1:-}" ]]; then
+  echo "Usage: $0 <VERSION>"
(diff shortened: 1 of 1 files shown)


tomowatt:
Where is the documentation for this for future reference? Else it is just a script in repository, so need to make sure people know to use it correctly

scadu:
> Where is the documentation for this for future reference? Else it is just a script in repository, so need to make sure people know to use it correctly

@tomowatt we do have internal documentation. I meant to update it once this PR is merged.

Issue #1512, opened 2025-07-11
Update the default IPv4 address pools to avoid conflict with the host network


haoxu333:
**Describe the bug**
The [default address pools in the CI stack ](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/bb55193bb4ba2e4c06f467bb44c43e55a189981d/packer/linux/conf/bin/bk-configure-docker.sh#L60)are hard coded to be two CIDR blocks:
"172.17.0.0/12" and "192.168.0.0/16"

When the BuildKite agent is created in a VPC with the "172.17.0.0/16" CIDR block, AWS would provide a DNS server at "172.17.0.2". If we start a docker container on the BuildKite agent, the container will have no connectivity, but it's unable to connect to the DNS server on the bridge network. 

**Steps To Reproduce**
1. Set up a VPC with CIDR block "172.17.0.0/16"
2. Create a BuildKite CI stack in the VPC at version 6.36.0 or above
3. Create a BuildKite pipeline to run a curl command (e.g. curl google.com) in a docker container.
4. Run the pipeline

**Expected behavior**
The curl command should success

**Actual behaviour**
The curl command fails with error message "Could not resolve host: google.com"

**Stack parameters (please complete the following information):**
 - AWS Region: us-east-1
 - Version v6.36.0

**Additional context**
One thing I don't understand is that this does not fail 100% of the time. I wonder if the "192.168.0.0/16" CIDR block is being used sometimes.

PR #1542, opened 2025-08-07
Allow Docker default address pools to be changed

petetomasik:
Make the Docker default address pools configurable parameters. This will allow users the ability to change the CIDRs if these addresses conflict with the VPC IP range the Elastic CI Stack has been deployed to.

Fixes: https://github.com/buildkite/elastic-ci-stack-for-aws/issues/1512

### DIFF
--- a/packer/linux/conf/bin/bk-configure-docker.sh
+++ b/packer/linux/conf/bin/bk-configure-docker.sh
@@ -57,8 +57,8 @@ if [[ "${DOCKER_NETWORKING_PROTOCOL}" == "ipv4" ]]; then
   # This is the default
   cat <<<"$(
     jq \
-      --arg pool1 "${DOCKER_ADDRESS_POOL_1:-172.17.0.0/12}" \
-      --arg pool2 "${DOCKER_ADDRESS_POOL_2:-192.168.0.0/16}" \
+      --arg pool1 "${DOCKER_IPV4_ADDRESS_POOL_1:-172.17.0.0/12}" \
+      --arg pool2 "${DOCKER_IPV4_ADDRESS_POOL_2:-192.168.0.0/16}" \
       '."default-address-pools"=[{"base":$pool1,"size":20},{"base":$pool2,"size":24}]' \
       /etc/docker/daemon.json
   )" >/etc/docker/daemon.json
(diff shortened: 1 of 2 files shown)


packer/linux/conf/bin/bk-configure-docker.sh
```diff
@@ -66,7 +68,9 @@ elif [[ "${DOCKER_NETWORKING_PROTOCOL}" == "dualstack" ]]; then
   DOCKER_EXPERIMENTAL=true
   cat <<<"$(
     jq \
-      '.ipv6=true | ."fixed-cidr-v6"="2001:db8:1::/64" | .ip6tables=true | ."default-address-pools"=[{"base":"172.17.0.0/12","size":20},{"base":"192.168.0.0/16","size":24},{"base":"2001:db8:2::/104","size":112}]' \
+      --arg pool1 "${DOCKER_ADDRESS_POOL_1:-172.17.0.0/12}" \
```
scadu, line 71: Should we make subnet configurable?

packer/linux/conf/bin/bk-configure-docker.sh
```diff
@@ -66,7 +68,9 @@ elif [[ "${DOCKER_NETWORKING_PROTOCOL}" == "dualstack" ]]; then
...
+      --arg pool2 "${DOCKER_ADDRESS_POOL_2:-192.168.0.0/16}" \
+      '.ipv6=true | ."fixed-cidr-v6"="2001:db8:1::/64" | .ip6tables=true | ."default-address-pools"=[{"base":$pool1,"size":20},{"base":$pool2,"size":24},{"base":"2001:db8:2::/104","size":112}]' \
```
scadu, line 73: We allow configuring IPv4, but IPv6 is hardcoded. Is it intended?

packer/linux/conf/bin/bk-configure-docker.sh
petetomasik: This is acting as a fallback if the default params for `DockerAddressPool1` and `DockerAddressPool2` are emptied/removed.

templates/aws-stack.yml
```diff
@@ -587,6 +587,21 @@ Parameters:
       - "dualstack"
     Default: "ipv4"
 
+  DockerIPv4AddressPool1:
+    Type: String
+    Description: Primary IPv4 CIDR block for Docker default address pools. Only applies to Linux instances, not Windows.
```
scadu, line 592: Nit, but I think we could make description a bit more clear, 
```suggestion
    Description: Primary IPv4 CIDR block for Docker default address pools. Must not conflict with host network or VPC CIDR. Only applies to Linux instances, not Windows.
```

PR #1559, opened 2025-08-22
Add Kernel Modules for CanUtils to work 

beekarthik:
Simple update so that socketcan can work in CI 

Reduce the need for on-prem instances

Tested  by building a new AMI 

`make packer-linux-amd64.output`

using the `autoscale-test` for CI test that needs socket can


beekarthik:
sorry - misclick again

PR #1592, opened 2025-09-18
Trigger base AMI base AMI rebuild

scadu:
After merging
https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1590 we'd need to build new base AMI without plugins, so the hash is built.

Also, remove `if_changed` conditional since `packer.sh` handles the logic based on the hash whether to rebuild base.


scadu:
Closed in favor of https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1593

PR #1593, opened 2025-09-18
Simplify base AMI build logic

scadu:
Remove hash-based logic in packer.sh in favor of `if_changed` mechanism provided by bk agent.

### DIFF
--- a/.buildkite/steps/packer.sh
+++ b/.buildkite/steps/packer.sh
@@ -17,65 +17,33 @@ fi
 
 mkdir -p "build/"
 
-# Build a hash of packer files and the agent versions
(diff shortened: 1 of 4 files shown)


.buildkite/steps/packer.sh
```diff
@@ -17,66 +17,34 @@ fi
 
 mkdir -p "build/"
 
-# Build a hash of packer files and the agent versions
-packer_files_sha=$(find Makefile "packer/${os}" plugins/ -type f -print0 | xargs -0 sha256sum | awk '{print $1}' | sort | sha256sum | awk '{print $1}')
-internal_files_sha=$(find go.mod go.sum internal/ -type f -print0 | xargs -0 sha256sum | awk '{print $1}' | sort | sha256sum | awk '{print $1}')
```
petetomasik, line 22: Do we still need to consider changes for `go.mod`, `go.sum`, etc. with `if_changed` for base/stack images?

PR #1604, opened 2025-09-26
Fix missing refresh_authorized_keys.timer

scadu:
After merging
https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1597, goss test started failing with missing refresh_authorized_keys.timer.

Looking closer, it looks like Packer overrides `/tmp/conf` directory with content of `../shared/conf`.
___

Verifying base AMI built with this change, we got the timer back:
```
systemctl status refresh_authorized_keys.timer
○ refresh_authorized_keys.timer - Download ssh authorized_keys file from s3
     Loaded: loaded (/etc/systemd/system/refresh_authorized_keys.timer; disabled; preset: disabled)
     Active: inactive (dead)
    Trigger: n/a
   Triggers: ● refresh_authorized_keys.service
```

### DIFF
--- a/packer/linux/base/base.pkr.hcl
+++ b/packer/linux/base/base.pkr.hcl
@@ -87,14 +87,8 @@ build {
   }
 
   provisioner "file" {
-    destination = "/tmp/shared-conf"
+    destination = "/tmp/conf/"
     source      = "../shared/conf/"
-  }
-
-  provisioner "shell" {
-    inline = [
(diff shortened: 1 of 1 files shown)


scadu:
> I think just adding [trailing slashes to the src and dest should merge](https://developer.hashicorp.com/packer/docs/provisioners/file#directory-uploads) but not going to nit-pick as I assume it's been tested and don't want to block

@JoeColeman95 I hoped it'd worked that way. Tested it now – works fine! Thanks for flagging.

PR #1631, opened 2025-11-04
Fix Renovate custom manager for buildkite-agent

scadu:
After merging https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1597, `managerFilePatterns` for `install-buildkite-agent` scripts no longer matched their actual path, resulting in breaking custom manager.

This PR fixes this issue. Additionally, `prBodyTemplate` is replaced with `prBodyNotes` as per Renovate's docs
(https://docs.renovatebot.com/configuration-templates/#pr-body): 
> * Change the entire layout/flow by using prBodyTemplate (we do not recommend this) 
> * Add a note by using prBodyNotes

This, hopefully, fixes the issue with missing version numbers in the PR body, example:
https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1601

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,8 +10,8 @@
     {
       "customType": "regex",
       "managerFilePatterns": [
-        "/^packer/linux/scripts/install-buildkite-agent\\.sh$/",
-        "/^packer/windows/scripts/install-buildkite-agent\\.ps1$/"
+        "/^packer/linux/stack/scripts/install-buildkite-agent\\.sh$/",
+        "/^packer/windows/stack/scripts/install-buildkite-agent\\.ps1$/"
       ],
       "matchStrings": [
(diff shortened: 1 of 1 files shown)


JoeColeman95:
> > Did they decide to mix whether you use triple or double handlebars?!
> 
> @mcncl, I've seen both and it's still not 100% sure which is expected to be used in what context. Renovate's documentation would use triple when double works, so that only adds to confusion. And verifying Renovate config locally doesn't always bring answers.

`{{}}` uses HTML escaping, whereas `{{{}}}` uses raw values, so that the value cannot be HTML escaped. In this case either work for version as there's no special chars used, but it would be breaking if we were ever to do anything like `v6.50.1-rc1`.

I suspect that `changelog` in this case will be mangled with `{{}}` as it'll likely include HTML-like anchors.

PR #1657, opened 2025-11-14
Make base AMI private

scadu:
We don't need them to be public (and potentially replicated across regions) as base images are only a foundation for the final image.


scadu:
> Looks good to me. Before merging, please verify that the resulting AMIs are listed as private in AWS.

@lucaswilric will verify after new base images are baked.

PR #1665, opened 2025-11-18
Add option to create VPC endpoints

scadu:
Enable optional VPC endpoints for S3, ECR, SSM, Secrets Manager, and KMS to reduce internet traffic and improve security.
Only available when stack creates its own VPC.

Disabled by default. To create VPC endpoints, use `EnableVpcEndpoints=true`.

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -1644,7 +1644,7 @@ Resources:
         - IpProtocol: tcp
           FromPort: 443
           ToPort: 443
-          CidrIp: 10.0.0.0/16
+          CidrIp: !GetAtt Vpc.CidrBlock
       Tags:
         - Key: Name
           Value: !Sub '${AWS::StackName}-vpc-endpoints'


templates/aws-stack.yml
```diff
@@ -1513,6 +1531,8 @@ Resources:
     Condition: CreateVpcResources
     Properties:
       CidrBlock: 10.0.0.0/16
```
tomowatt: TIL
I thought we allowed passing in a CIDR Block rather than hardcoding a value

templates/aws-stack.yml
```diff
@@ -1614,6 +1634,129 @@ Resources:
...
+    Type: AWS::EC2::SecurityGroup
+    Condition: EnableVpcEndpoints
+    Properties:
+      GroupDescription: Security group for VPC endpoints
+      VpcId: !Ref Vpc
+      SecurityGroupIngress:
+        - IpProtocol: tcp
+          FromPort: 443
+          ToPort: 443
+          CidrIp: 10.0.0.0/16
```
tomowatt: Might be over engineering but could use the `CidrBlock` attribute from the VPC itself to make sure it aligns e.g.,
```
CidrIp: !GetAtt Vpc.CidrBlock
```

PR #1666, opened 2025-11-18
fix: Add InstanceScaleInProtection parameter to stack

JoeColeman95:
Adds `InstanceScaleInProtection` as a configurable option, as this is currently just hard-coded to `true`.

This will default to `true` to replicate the same behaviour, but enables users to disable this when/if required.

Updated launch script for our own CI to inject this value as `false`.


AliSoftware:
Also, if I try setting `InstanceScaleInProtection = true` on my templates, my ASG enters a never-ending cycle of considering every newly created instance immediately unhealthy, which leads it to immediately terminate it and start a new one, wash, rince, repeat…
image

Shouldn't we set the ASG's Health check grace period to a value > 0 (or have it use a lifecycle hook) to prevent that from happening? 🤔 
image
image

PR #1675, opened 2025-11-21
Add BuildkiteQueue and AgentsPerInstance tags to Agent Autoscaling groups

fd-jonathanlinn:
## Description
Adds `BuildkiteQueue` and `AgentsPerInstance` tags to the AgentAutoscaleGroup resource, referencing the parameters of the same name.  This makes it a little easier to correlate different resources in tools like Datadog, where you can now rely on a single Template Variable to filter EC2 instances _and_ autoscaling Groups, as well as determining how many actual agents are running.

Deliberately set the tags not to propagate to the EC2 instances created by the ASG, as that tagging is handled by the LaunchTemplate.

Template changes validate fine using AWS CLI.

## Checklist

- [ ] Tests pass locally
- [x] Linting checks pass

## Release Notes

- [ ] My changes affect the public API (variable renames, new stack parameters, etc)

### DIFF
--- a/templates/aws-stack.yml
+++ b/templates/aws-stack.yml
@@ -2550,13 +2550,13 @@ Resources:
         - OldestLaunchConfiguration
         - ClosestToNextInstanceHour
       NewInstancesProtectedFromScaleIn: !Ref InstanceScaleInProtection
-        Tags:
-          - Key: BuildkiteQueue
-            PropagateAtLaunch: false
-            Value: !Ref BuildkiteQueue
-          - Key: AgentsPerInstance
-            PropagateAtLaunch: false
-            Value: !Ref AgentsPerInstance
+      Tags:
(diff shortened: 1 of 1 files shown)


deanmowat-fd (approved):

scadu (changes requested):
Thanks for the contribution, @fd-jonathanlinn!
The pipeline failed due to the issue with the template. From `cfn-lint templates/*`:
```
E3001 Additional properties are not allowed ('Tags' was unexpected)
templates/aws-stack.yml:2560:5
```

Moving `Tags` to the `Properties` block should fix the issue as now it's placed at the resource level alongside `UpdatePolicy` and `CreationPolicy`.
Moving the whole block to line 2553, just under `NewInstancesProtectedFromScaleIn` should fix it.

PR #1814, opened 2026-06-16
Enhance termination script with improved metadata fetching

nsuma8989:
Refactor instance termination logic to attempt graceful termination with and without capacity decrement. Added retry mechanism for fetching EC2 metadata.

## Description

Changes in the PR are addressing the issues with the terminate instance script such as silent failure before AWS API calls, no validation on scenario where min and max size is 1 etc.

## Context

SUP-6914

### DIFF
--- a/packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance
+++ b/packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance
@@ -9,6 +9,12 @@ terminate() {
     --should-decrement-desired-capacity
 }
 
+terminate_without_decrement() {
+  aws autoscaling terminate-instance-in-auto-scaling-group \
+    --region "$1" \
+    --instance-id "$2"
+}
+
 mark_as_unhealthy() {
   aws autoscaling set-instance-health \
     --region "$1" \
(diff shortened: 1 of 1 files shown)


omehegan:
This reminded me of https://github.com/buildkite/buildkite-agent-scaler/pull/300 - great work adding some more protection here!

packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance
```diff
@@ -9,44 +9,109 @@ terminate() {
...
 
+fetch_metadata() {
+  local endpoint="$1"
+  local token="$2"
+  local max_retries=3
+  local retry_count=0
+  local backoff=1
+
+  while [[ $retry_count -lt $max_retries ]]; do
+    if [[ -z "$token" ]]; then
```
omehegan, line 50: I notice that we call this function with `fetch_metadata "" ""` but here we only check if `$token` is unset. If we accidentally call the function with `fetch_metadata $token ""` then we will fall through to the next condition and try to run the `curl` command with an empty location value. Might just want to check against that.

packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance
```diff
@@ -9,44 +9,109 @@ terminate() {
...
+      # Fetching token (no token header needed)
+      if curl \
+        --fail --silent --show-error \
+        --connect-timeout 2 \
+        --max-time 5 \
+        -X PUT \
+        -H "X-aws-ec2-metadata-token-ttl-seconds: 60" \
```
omehegan: If I add up the various `curl` time limits and number of retries we're allowing, I get a maximum possible run time of 54 seconds. I think we might want to double the token expiry to 120 seconds to give ourselves more margin for future code changes that might slow things down. If we're about to terminate the instance anyway, I can't see any downside to extending this.

packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance
```diff
@@ -9,44 +9,109 @@ terminate() {
...
+      # If both termination attempts fail, mark as unhealthy
+      echo "Direct termination failed, marking instance as unhealthy..."
+      mark_as_unhealthy "$region" "$instance_id"
+    fi
+  fi
 else
-  # If we're not in terminate-after-job mode, then it's fine for this to fail. Systemd will restart
-  # the agent and it'll be as if the instance never got shut down.
-  terminate "$region" "$instance_id"
+  # In non-terminate-after-job mode, also attempt graceful termination with fallback
```
omehegan: At this point, the `else` branch is identical to the `if` branch above, so I'd suggest that we just collapse the logic and remove the else.

omehegan (changes requested):
Great improvements Suma! Just a couple of suggestions for you, if you can resolve these then I think this is good to merge.

packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance
nsuma8989: Increased the token expiry to 120

packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance
nsuma8989: Updated the function to validate the empty values for token and location

packer/linux/stack/conf/buildkite-agent/scripts/terminate-instance
nsuma8989: Resolved this and also update to include latest changes

nsuma8989:
> Great improvements Suma! Just a couple of suggestions for you, if you can resolve these then I think this is good to merge.

Thank you for the inputs. All the suggested changes are completed. Please can you review.

omehegan (approved):

PR #1815, opened 2026-06-17
Rework log collector to use AWS CLI directly

JoeColeman95:
Now collects logs for instances, and lambdas. Extended options.

## Description

- Updates the log collector as it was fragile and required JQ, depending on hardcoded log group names. Also only collected 4 instance log groups, silently swallowed AWS errors and left temp files behind.
- This discovers all lambda functions associated with a stack and any nested stacks. It resolves each function's log group and then collects them.
- Can now optionally collect instance logs and all of their associated log groups.
- Allows for region now
- Allows for a number of days worth of logs to be collected

## Checklist

- [x] Tests pass locally
- [x] Linting checks pass

## Release Notes

- [ ] My changes affect the public API (variable renames, new stack parameters, etc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,9 +282,17 @@ Provide Buildkite with logs from CloudWatch Logs:
 
 ### Collect logs via script
 
-An alternative method to collect the logs is to use the `log-collector` script in the `utils` folder.
-The script will collect CloudWatch logs for the Instance, Lambda function, and AutoScaling activity and package them in a
-zip archive which you can send via email to support@buildkite.com.
+Use [`utils/log-collector`](utils/log-collector) to collect CloudWatch logs from your stack and bundle them into a `.tar.gz` for support.
+
+```bash
+# Lambda + ASG logs
+./utils/log-collector -s my-stack-name
+
+# Also include a specific instance's logs
+./utils/log-collector -s my-stack-name -i i-0123456789abcdef0
+```
+
+The script requires the AWS CLI configured with credentials for the account the stack runs in. Run with `-h` for all options.
 
 You can also visit our [Forum](https://forum.buildkite.community) and post a question in the [Elastic CI Stack for AWS](https://forum.buildkite.community/c/elastic-ci-stack-for-aws/) section!
 
--- a/README.md
+++ b/README.md
@@ -282,17 +282,19 @@ Provide Buildkite with logs from CloudWatch Logs:
 
 ### Collect logs via script
 
-Use [`utils/log-collector`](utils/log-collector) to collect CloudWatch logs from your stack and bundle them into a `.tar.gz` for support.
+[`utils/log-collector`](utils/log-collector) pulls your stack's CloudWatch logs into a tarball you can send to support.
 
 ```bash
-# Lambda + ASG logs
+# Lambda and ASG logs
 ./utils/log-collector -s my-stack-name
 
-# Also include a specific instance's logs
+# Add a specific instance's logs too
 ./utils/log-collector -s my-stack-name -i i-0123456789abcdef0
 ```
 
-The script requires the AWS CLI configured with credentials for the account the stack runs in. Run with `-h` for all options.
+You'll need the AWS CLI set up with credentials for the account the stack runs in. It only reads (CloudFormation, Lambda, CloudWatch Logs and Auto Scaling), so a read-only role is fine. Run it with `-h` for the other flags.
+
+If a source has no logs you'll get a `.EMPTY` file instead of a `.log`, and a failed one leaves a `.ERROR` file. Whatever went wrong is also listed in `COLLECTION-ERRORS.txt`.
 
 You can also visit our [Forum](https://forum.buildkite.community) and post a question in the [Elastic CI Stack for AWS](https://forum.buildkite.community/c/elastic-ci-stack-for-aws/) section!
 
--- a/utils/log-collector
+++ b/utils/log-collector
@@ -268,6 +268,22 @@ package() {
   echo "Finished collecting logs, you can find them at $PWD/$archive"
 }
 
+summarise() {
+  local errors=0 logs
+  if [[ -f "$dir/COLLECTION-ERRORS.txt" ]]; then
+    errors=$(wc -l <"$dir/COLLECTION-ERRORS.txt" | tr -d ' ')
+  fi
+  logs=$(find "$dir" -maxdepth 1 -type f -name '*.log' | wc -l | tr -d ' ')
+
+  if ((errors > 0)); then
+    echo "WARNING: collection completed with $errors error(s). See COLLECTION-ERRORS.txt for more details." 1>&2
+  fi
+  if ((logs == 0 && errors > 0)); then
+    echo "ERROR: every source failed." 1>&2
+    exit 1
+  fi
+}
+
 main() {
   parse_args "$@"
 
@@ -294,6 +310,7 @@ main() {
   collect_asg
 
   package
+  summarise
 }
 
 main "$@"
--- a/utils/log-collector
+++ b/utils/log-collector
@@ -68,7 +68,7 @@ parse_args() {
     exit 1
   fi
 
-  if [[ -n "$instanceId" && ! "$instanceId" =~ ^i-[0-9a-f]+$ ]]; then
+  if [[ -n "$instanceId" && ! "$instanceId" =~ ^i-[0-9a-f]{8}([0-9a-f]{9})?$ ]]; then
     echo "Error: -i must be an EC2 instance ID (got '$instanceId')" 1>&2
     exit 1
   fi
--- a/utils/log-collector
+++ b/utils/log-collector
@@ -163,20 +163,20 @@ list_lambdas() {
 }
 
 find_lambdas() {
-  list_lambdas "$stackName"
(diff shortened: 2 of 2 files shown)


utils/log-collector
```diff
@@ -14,99 +14,286 @@
...
+  if is_empty "$file"; then
+    mark_empty "lambda-${safe}.log" "No log events in $logGroup within the last $days day(s)"
+  fi
+}
 
-# Iterate through all of the log streams we collected from the lambda log group
-echo "Collecting Lambda logs for $lambdaFunction"
+collect_lambdas() {
+  local functions fn
+  functions=$(find_lambdas | tr '\t' '\n' | sort -u)
```
petetomasik: This is the only AWS-touching command substitution that isn't guarded against `set -euo pipefail`. It's safe today (`find_lambdas` ends on a `while` loop so it exits 0, and `tr`/`sort` don't realistically fail), but if a future edit lets `find_lambdas` exit non-zero, `pipefail` will abort the whole script right here and `collect_asg` never runs. Worth guarding to match the `if ! x=$(...)` pattern used for every other capture.

```suggestion
  functions=$(find_lambdas | tr '\t' '\n' | sort -u) || true
```

`find_lambdas` already records its own failures via `record_error`, so dropping the pipeline status here doesn't hide anything.

utils/log-collector
```diff
@@ -14,99 +14,286 @@
...
-    ;;
-  \?)
-    echo "Invalid Option: -$OPTARG" 1>&2
     exit 1
-    ;;
-  :)
-    echo "Invalid option: $OPTARG requires an arguemnt" 1>&2
+  fi
+
+  if [[ -n "$instanceId" && ! "$instanceId" =~ ^i-[0-9a-f]+$ ]]; then
```
petetomasik: Nit: `^i-[0-9a-f]+$` accepts any hex length, so `i-0` or a 22-char string validate while real EC2 instance IDs are exactly 8 or 17 hex chars. No functional impact (AWS rejects a malformed ID anyway), but tightening the pattern gives a clearer up-front error.

```suggestion
  if [[ -n "$instanceId" && ! "$instanceId" =~ ^i-[0-9a-f]{8}([0-9a-f]{9})?$ ]]; then
```

README.md
```diff
@@ -282,9 +282,17 @@ Provide Buildkite with logs from CloudWatch Logs:
...
+
+```bash
+# Lambda + ASG logs
+./utils/log-collector -s my-stack-name
+
+# Also include a specific instance's logs
+./utils/log-collector -s my-stack-name -i i-0123456789abcdef0
+```
+
+The script requires the AWS CLI configured with credentials for the account the stack runs in. Run with `-h` for all options.
```
petetomasik: The README covers `-s`/`-i` but defers everything else to `-h`. Two things worth surfacing here, since this is where someone lands when they want more than the default: `-d <days>` is the flag that lifts the old "last 24h only" limitation, and the archive now contains per-source status markers that a reader should expect rather than mistake for a broken run. The expanded AWS API surface is also worth a one-liner so anyone on a restricted role knows what to grant.

```suggestion
The script requires the AWS CLI configured with credentials for the account the stack runs in; it calls read-only `cloudformation`, `lambda`, `logs`, and `autoscaling` APIs. By default it collects the last day of logs. Pass `-d <days>` (1-365) for a longer window, `-o <text|json|table>` to change the output format, and `-r <region>` to override the region. Run with `-h` for all options.

Inside the archive, sources with no events get a `.EMPTY` marker, failed collections get a `.ERROR` marker, and all errors are aggregated in `COLLECTION-ERRORS.txt`.
```

utils/log-collector
```diff
@@ -14,99 +14,286 @@
...
+  if is_empty "$file"; then
+    mark_empty "asg-scaling.log" "No scaling activity found for $asg"
+  fi
+}
+
+package() {
+  local archive
+  archive="$(basename "$dir").tar.gz"
+  tar -czf "$archive" "$dir"
+  echo "Finished collecting logs, you can find them at $PWD/$archive"
```
petetomasik, line 280: A run where the `describe-stacks` pre-flight passes but the actual collection fails (throttling, partial IAM, a session that expires mid-run) still reaches this line, prints "Finished collecting logs...", and exits 0. The failures do go to stderr and `COLLECTION-ERRORS.txt`, so it isn't silent, but a non-interactive run or a quick skim reads as success and the customer potentially ends up shipping a tarball of `.ERROR` markers to Support.

Since this is the tool people will be asked to use when something is already broken, it'd be worth having `main` track how many sources failed (or how many `.log` files survived) and surface it: print "N sources failed, see COLLECTION-ERRORS.txt" before this line, and exit non-zero when nothing was collected at all. Partial success can stay exit 0, since partial logs are still useful.

utils/log-collector
```diff
@@ -14,99 +14,286 @@
...
+    --stack-name "$stackName" \
+    --query "StackResourceSummaries[?ResourceType=='AWS::CloudFormation::Stack' || ResourceType=='AWS::Serverless::Application'].PhysicalResourceId" \
+    --output text 2>"$errFile" | tr '\t' '\n'); then
+    record_error "Failed to list nested stacks in $stackName: $(last_error)"
+    return 0
+  fi
+
+  while IFS= read -r child; do
+    [[ -z "$child" ]] && continue
+    list_lambdas "$child"
```
petetomasik: `find_lambdas` goes exactly one level deep: it lists the root stack's Lambdas, then for each immediate nested stack calls `list_lambdas` (not `find_lambdas`), so a Lambda in a grandchild stack is silently missed. That covers today's topology (the scaler `Serverless::Application` and the SSM-AMI stack are both depth-1), so it isn't a bug right now, but the PR description says it discovers Lambdas in "any nested stacks," which overstates it.

Two options: have the child branch call `find_lambdas` (or a recursive helper) so it actually walks the tree, or note the one-level limit in the description/usage. I'd lean toward recursing since it's a small change and makes the claim true, but documenting is fine if you're confident the stack won't nest deeper.

utils/log-collector
```diff
@@ -14,99 +14,286 @@
...
+}
 
-# Collect some information about the Lambda function and associated log groups
-lambdaFunction=$(aws cloudformation describe-stack-resources --stack-name "$stackName" --logical-resource-id Autoscaling --query "StackResources[*].PhysicalResourceId" --output text | sed -nr 's/.*stack\/(.*)\/.*/\1/p')
-lambdaLogGroup=$(aws cloudformation describe-stack-resources --stack-name "$lambdaFunction" --logical-resource-id LogGroup --query "StackResources[*].PhysicalResourceId" --output text)
-logStreams=$(aws logs describe-log-streams --log-group-name "$lambdaLogGroup" --order-by LastEventTime --query "logStreams[?creationTime > \`$logAge\`] | [*].logStreamName" --output text | tr -s '[:space:]' '\n')
+collect_instance() {
+  echo "Collecting logs for instance $instanceId..."
+  local name safe file err
+  for name in buildkite-agent system lifecycled docker-daemon cloud-init cloud-init/output cfn-init elastic-stack auth; do
```
petetomasik: This list is the nine Linux CloudWatch agent groups, which is right for Linux instances but noisy on Windows: a Windows stack uses a different set, so `cloud-init`, `cloud-init/output`, `auth`, and `docker-daemon` won't exist (each becomes a `.ERROR` marker), and the group that matters most for Windows boot/userdata debugging, `EC2Launch/UserdataExecution`, is never collected. Combined with the success-on-failure behavior flagged elsewhere, a Windows operator gets a "finished" tarball that's mostly errors and missing the useful log.

Adding `EC2Launch/UserdataExecution` to the loop is harmless on Linux (it just yields a `.EMPTY` marker), so unioning the two platforms' group lists is the low-risk fix. At minimum, note in usage that `-i` collection is Linux-oriented.

utils/log-collector
```diff
@@ -14,99 +14,286 @@
...
+  echo "$2" >"$dir/$1.ERROR"
+  rm -f "$dir/$1"
+}
 
-# CloudWatch stores the creationTime as milliseconds since epoch UTC.
-# Since macOS doesn't support calculating the time since epoch in milliseconds, we don't have perfect accuracy on this, but it should be as close to 24 hours as possible
-# We'll check to see if we're running on a macOS-based OS and adjust appropriately, otherwise use the standard date command
+is_empty() {
+  [[ ! -s "$1" ]] && return 0
+  [[ "$(wc -c <"$1" | tr -d ' ')" -le 8 && "$(tr -d '[:space:]' <"$1")" == "[]" ]]
```
petetomasik, line 115: Two small nits on `is_empty`, neither blocking:

- The `wc -c ... -le 8` byte guard before the `== "[]"` check is a micro-optimization that adds a subshell and a `tr`; `[[ "$(tr -d '[:space:]' <"$1")" == "[]" ]]` on its own is equivalent and reads cleaner.
- Empty detection only catches text (0 bytes) and JSON (`[]`). With `-o table`, an empty result is a few bytes of table chrome, so it gets archived as a `.log` rather than a `.EMPTY` marker. Harmless (no data loss), but worth either a one-line note that empty-marking is best-effort for table, or detecting "no data rows" per format.

petetomasik (approved):
🚀

PR #1817, opened 2026-06-19
Add scaler event schedule jitter parameter

vivster7:
## Summary
- Bump buildkite-agent-scaler to v1.12.0.
- Add `ScalerEventScheduleJitter` with a default of `0s` so stacks can opt into staggered scaler polling.
- Pass the jitter setting through to both x86_64 and arm64 scaler nested applications.

## Test plan
- `git diff --check`
- `mise x python@3.12.13 -- cfn-lint templates/*`

Made with [Cursor](https://cursor.com)


lizrabuya (approved):
Looks great!

PR #1836, opened 2026-07-15
CIS hardened build

omehegan:
### Goal

Make the standard Elastic CI Stack AMI compatible with CIS Level 1 hardening applied on top. Customers using EC2 Image Builder (or similar) to apply CIS benchmarks to our published AMI should have it work out of the box, with no configuration on their end.

This is an extension of #1832.

Ref. SUP-7508

### What we didn't want to do

- **Publish a separate CIS AMI.** That would mean marketplace licensing concerns, per-hour costs, and maintaining two image variants. Instead, all fixes are backward-compatible — the standard AMI becomes CIS-compatible.
- **Require customer-side workarounds.** The customer who reported this had already found the `TMPDIR` fix via systemd drop-in. We wanted that (and everything else) handled automatically.

### How

Two classes of CIS incompatibility were found and fixed:

**`/tmp` noexec (CIS Rule 1.1.2.3):** The agent writes hook wrapper scripts to `$TMPDIR` (defaults to `/tmp`) and executes them. CIS mounts `/tmp` noexec, causing `Permission denied`. Fixed by setting `TMPDIR=/var/tmp` in the agent's systemd service file, and redirecting all Packer provisioner scripts and `mktemp` calls to `/var/tmp`.

**Restrictive umask (CIS Rule 5.5.5):** CIS tightens umask to `027`, so files created by root during bootstrap are unreadable by the `buildkite-agent` user. Fixed by:
- `chown buildkite-agent:` on `cfn-env` and `env` files written at boot
- `chmod 755` instead of `chmod +x` throughout (the latter is umask-relative and produces `750` under `027`)
- Explicit `chmod 755` on directories that need to be traversable
- Explicit `groupadd` before `useradd` (CIS may disable automatic group creation)

A CIS build variant was added to the CI pipeline to validate these fixes on every build. It uses a CIS AL2023 marketplace AMI as the base, builds the stack on top, launches a CloudFormation stack, and runs the full goss test suite.

### Review notes

Buildsworth flagged a few items we intentionally left as-is:

- **File provisioners still use `/tmp`** — these files are read/copied, never executed. `noexec` doesn't affect reads. Moving 25+ references for zero benefit isn't worthwhile.
- **`copy-ami` doesn't depend on `test-linux-amd64-cis`** — intentional. CIS AMIs are private validation builds, not published.
- **`cfn-env` logged before `chown`** — the file explicitly doesn't contain secrets (existing code comment confirms this).How does that look? I'd suggest squashing before opening the PR — the commit history has a lot of debug/iteration commits (diagnostics, temporary pipeline stripping, hardcoded AMIs, reverts). Want me to squash it down, or would you prefer to keep the history?

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -417,6 +417,7 @@ steps:
     timeout_in_minutes: 60
     env:
       AMI_PUBLIC: false
+    retry: { automatic: { limit: 3 } }
     agents:
       queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
     if: build.source != "schedule"
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -142,6 +142,10 @@ steps:
             - organization_id
             - pipeline_slug
 
+  # ============================================================
+  # Windows AMD64
+  # ============================================================
+
   - id: "packer-base-windows-amd64"
     name: ":packer: :windows: (base)"
     command: ".buildkite/steps/packer.sh windows amd64 base"
@@ -220,6 +224,10 @@ steps:
       - "test-windows-amd64"
     plugins:
       - *aws_role_plugin
+
+  # ============================================================
+  # Linux AMD64
+  # ============================================================
 
   - id: "packer-base-linux-amd64"
     name: ":packer: :linux: AMD64 (base)"
@@ -301,6 +309,10 @@ steps:
     plugins:
       - *aws_role_plugin
 
+  # ============================================================
+  # Linux ARM64
+  # ============================================================
+
   - id: "packer-base-linux-arm64"
     name: ":packer: :linux: ARM64 (base)"
     command: ".buildkite/steps/packer.sh linux arm64 base"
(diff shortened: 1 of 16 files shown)


Makefile
scadu: This dependency still runs when BASE_AMI_ID is set, so a clean CIS job builds an unused normal base AMI first. Can we skip it when the caller provides a base AMI?

```suggestion
packer-linux-amd64.output: $(PACKER_LINUX_STACK_FILES) build/fix-perms-linux-amd64 build/goss-linux-amd64 $(if $(strip $(BASE_AMI_ID)),,packer-base-linux-amd64.output)
```

.buildkite/pipeline.yaml
```diff
@@ -380,6 +392,94 @@ steps:
...
+      - "packer-base-linux-amd64-cis"
+      - "linting"
+      - "fixperms-tests"
+    plugins:
+      - *aws_role_plugin
+
+  - id: "launch-linux-amd64-cis"
+    name: ":cloudformation: :linux: AMD64 CIS Launch"
+    command:
+      - .buildkite/steps/launch.sh linux amd64 cis
```
scadu, line 436: Using if_changed_stack_linux for the whole CIS chain means template-only and launch.sh changes skip CIS testing. Could we follow the normal Linux flow instead: keep the CIS Packer step narrow, run launch/test/delete with if_changed_launch_test_delete_linux, and extend ensure_ami_metadata.py to accept a cis variant?
The CIS fallback cannot use the published template. It could read a packer-linux-amd64-cis-latest.output pointer from S3 or find the latest private CIS AMI by tag.

```suggestion
      - .buildkite/steps/ensure_ami_metadata.py linux amd64 cis
      - .buildkite/steps/launch.sh linux amd64 cis
```

The ensure_ami_metadata.py script change should make cis select:
- metadata key: linux_amd64_cis_image_id
- fallback: latest private CIS stack AMI

This avoids rebuilding the CIS stack AMI for template-only changes while still running the CIS test.

.buildkite/pipeline.yaml
```diff
@@ -380,6 +392,94 @@ steps:
...
+    retry: { automatic: { limit: 3 } }
+    env:
+      ROOT_VOLUME_SIZE: "15"
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
+    artifact_paths: "build/aws-stack.yml"
+    if: build.source != "schedule"
+    if_changed: *if_changed_stack_linux
```
scadu: Related to https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1836/changes#r3585762819

```suggestion
    if_changed: *if_changed_launch_test_delete_linux
```

.buildkite/pipeline.yaml
```diff
@@ -380,6 +392,94 @@ steps:
...
+      - jq --version
+      - git --version
+      - sudo goss validate --format documentation
+    timeout_in_minutes: 5
+    retry: { automatic: { limit: 3 } }
+    agents:
+      stack: "buildkite-aws-stack-test-linux-amd64-cis-${BUILDKITE_BUILD_NUMBER}"
+      queue: "testqueue-linux-amd64-cis-${BUILDKITE_BUILD_NUMBER}"
+    if: build.source != "schedule"
+    if_changed: *if_changed_stack_linux
```
scadu: Related to https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1836/changes#r3585762819

```suggestion
    if_changed: *if_changed_launch_test_delete_linux
```

.buildkite/pipeline.yaml
```diff
@@ -380,6 +392,94 @@ steps:
...
+      - "launch-linux-amd64-cis"
+
+  - id: "delete-linux-amd64-cis"
+    name: ":cloudformation: :linux: AMD64 CIS Delete"
+    command: .buildkite/steps/delete.sh linux amd64 cis
+    allow_dependency_failure: true
+    agents:
+      queue: "${BUILDKITE_AGENT_META_DATA_QUEUE}"
+    if: build.source != "schedule"
+    if_changed: *if_changed_stack_linux
```
scadu: Related to https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1836/changes#r3585762819

```suggestion
    if_changed: *if_changed_launch_test_delete_linux
```

.buildkite/steps/launch.sh
```diff
@@ -3,8 +3,14 @@ set -euo pipefail
 
 os="${1:-linux}"
 arch="${2:-amd64}"
-stack_name="buildkite-aws-stack-test-${os}-${arch}-${BUILDKITE_BUILD_NUMBER}"
-stack_queue_name="testqueue-${os}-${arch}-${BUILDKITE_BUILD_NUMBER}"
+variant="${3:-}"
+if [[ -n "$variant" ]]; then
+  stack_name="buildkite-aws-stack-test-${os}-${arch}-${variant}-${BUILDKITE_BUILD_NUMBER}"
+  stack_queue_name="testqueue-${os}-${arch}-${variant}-${BUILDKITE_BUILD_NUMBER}"
```
scadu, line 9: The fallback cleanup expects the build number directly after the architecture, so it will not match names like linux-amd64-cis-6473. Can we allow an optional variant in cleanup.sh?

scadu (changes requested):

Makefile
```diff
@@ -240,6 +244,34 @@ packer-base-linux-amd64.output: $(PACKER_LINUX_BASE_FILES)
 			-var 'ami_users=$(AMI_USERS_LIST)' \
 			base.pkr.hcl | tee $@
 
+# CIS-hardened source AMI IDs
+CIS_SOURCE_AMI_AMD64 ?= ami-0ade66ab1b3aaa37a
```
scadu, line 254: AWS_REGION remains configurable, but the CIs source defaults to one fixed AMI ID. AMI IDs are region specific, so `AWS_REGION=us-west-2 make packer-base-linux-amd64-cis.output` will pass an AMI from another region unless callers also know to override the undocumented `CIS_SOURCE_AMI_AMD64`. 
We should resolve the CIS image per region, or explicitly restrict and document the target’s supported region(s).

omehegan: Ah yes, so the CIS images require a subscription (they charge $.06/hr to run) which is per-region. Given we normally only run the CI in us-east-1 I think it's fine to leave this hard-coded. I added a comment to explain.

scadu (approved):

PR #1842, opened 2026-07-21
Add Ubuntu 24.04 as a selectable Linux base OS

petetomasik:
## Description

Adds Ubuntu 24.04 LTS as a selectable Linux base OS, chosen at deploy time via a new `LinuxDistribution` CloudFormation parameter. It defaults to `amazonlinux2023`, so existing stacks are unaffected; setting it to `ubuntu2404` builds and launches agents on Ubuntu. Both distributions are published for amd64 and arm64.

The implementation keeps a single `packer/linux/` tree parameterized by a new `os_distro` Packer variable. A sourced `distro.sh` helper abstracts the package manager (`pkg_update`/`pkg_install`/`pkg_install_local`/`pkg_clean`) and per-distro facts (login user, CloudWatch log paths, SSM packaging), so the shared install and binary-download logic stays distro-agnostic. The Makefile and CI pipeline gain parallel `ubuntu2404` build targets, and the region mappings plus the template gain `ubuntu2404amd64`/`ubuntu2404arm64` AMI keys.

Distro-specific handling on Ubuntu: `amazon-ssm-agent` installs via snap, `amazon-cloudwatch-agent` and `session-manager-plugin` install from `.deb` packages (with download-integrity hardening), and `aws-cfn-bootstrap` (cfn-signal) installs from Amazon's tarball with its Python dependencies sourced from apt. Amazon Linux 2023 behavior is unchanged. The launch template also now roots Ubuntu (and Windows) AMIs on `/dev/sda1` so the `RootVolumeSize` override applies to the real root volume.

Validated with live AMI builds (goss) and a runtime-validation pipeline that exercises real agent jobs on both distributions across amd64 and arm64.

## Checklist

- [x] Tests pass locally
- [ ] Updated documentation (if applicable)
- [x] Linting checks pass

## Release Notes

**Public API change:** this adds a new `LinuxDistribution` stack parameter to select the Linux base OS (`amazonlinux2023` default, or `ubuntu2404`). Existing stacks default to `amazonlinux2023` and are unaffected.

- Add Ubuntu 24.04 LTS as a selectable Linux base OS via the `LinuxDistribution` CloudFormation parameter (`amazonlinux2023` default, `ubuntu2404` opt-in), published for amd64 and arm64.

  - [x] I have added a note at the top of the release notes highlighting the API changes

### DIFF
--- a/packer/linux/base/base.pkr.hcl
+++ b/packer/linux/base/base.pkr.hcl
@@ -12,6 +12,12 @@ variable "arch" {
   default = "x86_64"
 }
 
+variable "os_distro" {
+  type        = string
+  description = "Base Linux distribution to build on: amazonlinux2023 or ubuntu2404."
+  default     = "amazonlinux2023"
+}
+
 variable "instance_type" {
   type    = string
   default = "m7a.xlarge"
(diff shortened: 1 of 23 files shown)


scadu (commented):
Aside from buildsworth's finding https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1842#discussion_r3625945055, it looks good. Happy to give a stamp once it's patched up ;)

packer/linux/base/base.pkr.hcl
```diff
@@ -124,33 +165,43 @@ build {
     source      = "scripts/versions.sh"
   }
 
+  provisioner "file" {
+    destination = "/tmp/"
+    source      = "../shared/scripts/distro.sh"
```
scadu, line 170: Minor: Could we add `packer/linux/shared` to both `PACKER_LINUX_BASE_FILES` and `PACKER_LINUX_STACK_FILES`? The base and stack Packer configurations both upload `distro.sh`, and the base provisioning scripts use it for package installation and cleanup. Their Make prerequisites currently only enumerate their respective `base/` and `stack/` directories.

Once an `.output` file exists, changing `distro.sh` or `cleanup.sh` won’t trigger either image to rebuild, so an incremental local build could silently reuse older provisioning logic.

scadu (approved):

PR #1846, opened 2026-07-27
Feat: SUP-7653 Implement Git Mirror Seed

dahtey-bk:
## Description

When `BuildkiteAgentEnableGitMirrors=true` on ephemeral stack instances, the first job on every fresh instance performs a full `git clone --mirror` of each repository it builds. For large monorepos this cold start dominates job startup time. Investigation was done and found that tar archives of pre-built bare mirrors are the right seeding mechanism: they preserve the existing `.pack`/`.idx` files, so extraction is pure I/O (benchmarked at ~2 min for a Linux-kernel-sized repo, with near-instant first checkouts) — unlike git bundles, which require a CPU-bound `git index-pack` on clone that takes longer than cloning from the remote.

This PR lets customers pre-seed the git-mirrors directory from S3 at instance boot, before the agent starts. The agent's first job then fetches only the delta since the archive was created. No agent changes are required: the agent already handles pre-existing mirrors, automatically correcting stale remote URLs (`updateRemoteURL()`) and fetching missing refs (`updateGitMirror()`).

### Usage

Set two stack parameters:

- `BuildkiteAgentEnableGitMirrors=true`
- `GitMirrorSeedBucket=<bucket-name>` (new, default `""`) — an existing S3 bucket containing seed archives

Archives live under a fixed `git-mirror-seeds/` prefix, one per repository, in `.tar`, `.tar.gz`, or `.zip` format. Filenames must match the agent's mirror directory name for the repository (all non-alphanumeric characters replaced with `-`):

```
s3://<bucket>/git-mirror-seeds/
  git-github-com-org-repo-git.tar
  https---github-com-acme-inc-project-git.tar.gz
```

At boot, the bootstrap script lists the prefix and extracts every archive into the git-mirrors path (`.tar`/`.tar.gz` are streamed straight from S3; `.zip` is staged to a temp file since the format can't be streamed).

### Behavior notes for reviewers

- **Condition requires both parameters.** `HasGitMirrorSeedBucket` is only true when git mirrors are enabled *and* a bucket is named. Otherwise no IAM policy is created and no seeding is attempted.
- **IAM is read-only and prefix-scoped**: `s3:GetObject` on `git-mirror-seeds/*` plus `s3:ListBucket` conditioned on the same prefix. Customers who want stack instances to *build* seed archives grant `s3:PutObject` separately. Instances using a custom `InstanceRoleARN` must be granted read access separately.
- **Seeding is best-effort; the agent always starts.** An unreachable or empty bucket logs a warning and boots with empty mirrors (existing cold-start behavior). A corrupt archive fails extraction, removes only that archive's partial mirror directory, and continues with the remaining archives.
- **Stale archives are fine** — the agent fetches the delta and fixes the mirror's remote URL on first use; fresher archives just mean smaller deltas.
- **Linux only.** 

Fixes the cold-start bottleneck described in SUP-7653 (design) / SUP-6596 (research).

## Checklist

- [x] Tests pass locally
- [ ] Added tests for new features/fixes
- [ ] Updated documentation (if applicable)
- [x] Linting checks pass

## Release Notes

### DIFF
--- a/packer/linux/stack/conf/bin/bk-install-elastic-stack.sh
+++ b/packer/linux/stack/conf/bin/bk-install-elastic-stack.sh
@@ -321,26 +321,74 @@ if [[ "${BUILDKITE_AGENT_ENABLE_GIT_MIRRORS:-false}" == "true" ]]; then
     echo Not mounting git-mirrors to instance storage as instance storage is disabled.
   fi
 
-  if [[ -n "${BUILDKITE_AGENT_GIT_MIRRORS_SEED_BUCKET:-}" ]]; then
-    GIT_MIRRORS_SEED_URI="s3://${BUILDKITE_AGENT_GIT_MIRRORS_SEED_BUCKET}/git-mirrors-seed.tar"
-    echo "Seeding git-mirrors from ${GIT_MIRRORS_SEED_URI}..."
+  if [[ -n "${BUILDKITE_GIT_MIRROR_SEED_BUCKET:-}" ]]; then
+    GIT_MIRROR_SEED_PREFIX="git-mirror-seeds"
+    echo "Seeding git-mirrors from s3://${BUILDKITE_GIT_MIRROR_SEED_BUCKET}/${GIT_MIRROR_SEED_PREFIX}/..."
 
-    # A missing or unreadable seed must not fail the boot, it is only an optimisation. The
-    # agent will fall back to cloning mirrors from scratch as usual.
-    if aws s3 cp "$GIT_MIRRORS_SEED_URI" /tmp/git-mirrors-seed.tar; then
-      if tar -xf /tmp/git-mirrors-seed.tar -C "$BUILDKITE_AGENT_GIT_MIRRORS_PATH"; then
-        echo "Extracted git-mirrors seed to $BUILDKITE_AGENT_GIT_MIRRORS_PATH"
-      else
-        # A partially extracted mirror would break git fetches, so start clean instead
-        echo "WARNING: Failed to extract git-mirrors seed, clearing git-mirrors and continuing..."
-        rm -rf "${BUILDKITE_AGENT_GIT_MIRRORS_PATH:?}"/*
-      fi
-      rm -f /tmp/git-mirrors-seed.tar
+    # Seeding is best-effort, it is only an optimisation. If listing or extraction fails, the
+    # agent falls back to cloning mirrors from scratch as usual, and the boot must not fail.
+    seed_keys=()
+    if seed_listing="$(
+      aws s3api list-objects-v2 \
+        --bucket "$BUILDKITE_GIT_MIRROR_SEED_BUCKET" \
+        --prefix "${GIT_MIRROR_SEED_PREFIX}/" \
+        --output json \
+        | jq -r '.Contents[]?.Key // empty'
+    )"; then
+      while IFS= read -r seed_key; do
+        if [[ "$seed_key" == *.tar || "$seed_key" == *.tar.gz || "$seed_key" == *.zip ]]; then
+          seed_keys+=("$seed_key")
+        fi
+      done <<<"$seed_listing"
     else
-      echo "WARNING: Failed to fetch git-mirrors seed from ${GIT_MIRRORS_SEED_URI}, continuing with empty git-mirrors..."
+      echo "WARNING: Failed to list git mirror seeds in s3://${BUILDKITE_GIT_MIRROR_SEED_BUCKET}/${GIT_MIRROR_SEED_PREFIX}/, continuing with empty git-mirrors..."
     fi
+
+    if [[ ${#seed_keys[@]} -eq 0 ]]; then
+      echo "No git mirror seeds found."
+    fi
+
+    for seed_key in "${seed_keys[@]}"; do
+      seed_archive="${seed_key##*/}"
+      seed_uri="s3://${BUILDKITE_GIT_MIRROR_SEED_BUCKET}/${seed_key}"
+      extracted=false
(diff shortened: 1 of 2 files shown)


packer/linux/stack/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -321,8 +321,30 @@ if [[ "${BUILDKITE_AGENT_ENABLE_GIT_MIRRORS:-false}" == "true" ]]; then
     echo Not mounting git-mirrors to instance storage as instance storage is disabled.
   fi
 
+  if [[ -n "${BUILDKITE_AGENT_GIT_MIRRORS_SEED_BUCKET:-}" ]]; then
+    GIT_MIRRORS_SEED_URI="s3://${BUILDKITE_AGENT_GIT_MIRRORS_SEED_BUCKET}/git-mirrors-seed.tar"
```
omehegan: I realize you may be hard-coding this for now as this is a draft PR. But just noting that, per the design doc, there may be more than one tarfile in the bucket, so the expectation is that the agent downloads all of them. Also, we should make sure that we support .tar.gz and .zip files, because compressing them will speed up the data transfer significantly.

packer/linux/stack/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -321,8 +321,78 @@ if [[ "${BUILDKITE_AGENT_ENABLE_GIT_MIRRORS:-false}" == "true" ]]; then
...
+      *.tar)
+        seed_mirror_dir="${seed_archive%.tar}"
+        if aws s3 cp "$seed_uri" - | tar -xf - -C "$BUILDKITE_AGENT_GIT_MIRRORS_PATH"; then
+          extracted=true
+        fi
+        ;;
+      *.zip)
+        # zip cannot be streamed: its central directory is at the end of the file
+        seed_mirror_dir="${seed_archive%.zip}"
+        seed_tmp="$(mktemp /tmp/git-mirror-seed-XXXXXX.zip)"
```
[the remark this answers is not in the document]
dahtey-bk: @buildsworth-bk review

packer/linux/stack/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -321,8 +321,80 @@ if [[ "${BUILDKITE_AGENT_ENABLE_GIT_MIRRORS:-false}" == "true" ]]; then
     echo Not mounting git-mirrors to instance storage as instance storage is disabled.
   fi
 
+  if [[ -n "${BUILDKITE_GIT_MIRROR_SEED_BUCKET:-}" ]]; then
```
JoeColeman95, line 324: Linux stack creation has a five minute default deadline which could easily be exceeded here, I'd suggest adjusting this or disabling it when seeding is enabled

packer/linux/stack/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -321,8 +321,80 @@ if [[ "${BUILDKITE_AGENT_ENABLE_GIT_MIRRORS:-false}" == "true" ]]; then
...
+          extracted=true
+        fi
+        ;;
+      *.zip)
+        # Stage zip file in the mirrors path rather than /tmp, because /tmp is a memory-backed
+        # tmpfs by default (MountTmpfsAtTmp), which large archives would exhaust,
+        # while the mirrors path is on a volume sized for mirror data.
+        seed_mirror_dir="${seed_archive%.zip}"
+        seed_tmp="$(mktemp "${BUILDKITE_AGENT_GIT_MIRRORS_PATH}/.git-mirror-seed-XXXXXX.zip")"
+        if aws s3 cp "$seed_uri" "$seed_tmp" && unzip -oq "$seed_tmp" -d "$BUILDKITE_AGENT_GIT_MIRRORS_PATH"; then
```
JoeColeman95, line 376: Archives are extracted straight into the live mirrors director here, but a bad archive can overwrite another mirror,  which would break checkouts, or if named .tar delete the whole directory during cleanup. I'd say this should be extracted and validated in a separate dir first.

JoeColeman95 (changes requested):

packer/linux/stack/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -321,8 +321,80 @@ if [[ "${BUILDKITE_AGENT_ENABLE_GIT_MIRRORS:-false}" == "true" ]]; then
...
+        if aws s3 cp "$seed_uri" - | tar -xf - -C "$BUILDKITE_AGENT_GIT_MIRRORS_PATH"; then
+          extracted=true
+        fi
+        ;;
+      *.zip)
+        # Stage zip file in the mirrors path rather than /tmp, because /tmp is a memory-backed
+        # tmpfs by default (MountTmpfsAtTmp), which large archives would exhaust,
+        # while the mirrors path is on a volume sized for mirror data.
+        seed_mirror_dir="${seed_archive%.zip}"
+        seed_tmp="$(mktemp "${BUILDKITE_AGENT_GIT_MIRRORS_PATH}/.git-mirror-seed-XXXXXX.zip")"
```
omehegan: I think if mktemp fails here, it will prevent the agent from starting. We should guard this.

omehegan (changes requested):
In addition to my inline comment, it might be worth creating a simple test harness for this functionality, which we could run during the CI pipeline. We could do that in a separate PR/ticket though.

JoeColeman95 (approved):

omehegan (approved):

PR #1879, opened 2026-09-03
Update buildkite-agent to v4

scadu:
## Description

Upgrade the bundled Linux and Windows agents from v3.137.2 to v4.0.3. 
Update the stack parameters and generated configuration for Agent v4, remove oldstable, and add an upgrade guide covering parameter and `AgentEnvFileUrl` changes.

## Checklist
 - [x] Tests pass locally
 - [x] Linting checks pass

## Release Notes

  - [x] I have added a note at the top of the release notes highlighting the API changes
- [x] Any changes to external libraries (agent, scaler function, etc) have been flagged in release notes

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -210,7 +210,6 @@ steps:
       - jq --version
       - git --version
       - buildkite-agent --version
-      - powershell -NoProfile -Command "$c = Get-Content C:\\buildkite-agent\\buildkite-agent.cfg; if (($c -notcontains 'opentelemetry-tracing=false') -or ($c -notcontains 'cancel-signal-timeout=10s') -or ($c -notcontains 'cancel-cleanup-timeout=5s')) { exit 1 }"
       - docker info
       - docker run --rm mcr.microsoft.com/windows/nanoserver:ltsc2022 cmd.exe /c echo hello
     timeout_in_minutes: 5
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -209,6 +209,8 @@ steps:
       - aws --version
       - jq --version
       - git --version
+      - buildkite-agent --version
+      - powershell -NoProfile -Command "$c = Get-Content C:\\buildkite-agent\\buildkite-agent.cfg; if (($c -notcontains 'opentelemetry-tracing=false') -or ($c -notcontains 'cancel-signal-timeout=10s') -or ($c -notcontains 'cancel-cleanup-timeout=5s')) { exit 1 }"
       - docker info
       - docker run --rm mcr.microsoft.com/windows/nanoserver:ltsc2022 cmd.exe /c echo hello
     timeout_in_minutes: 5
--- a/docs/upgrading-to-v7.md
+++ b/docs/upgrading-to-v7.md
@@ -1,0 +1,58 @@
+# Upgrading to Elastic CI Stack v7
+
+Elastic CI Stack v7 uses Buildkite Agent v4. Read the [Agent v3 to v4 upgrade guide](https://buildkite.com/docs/agent/v3-v4-upgrade-guide) first; this page only covers the stack-specific steps.
+
+## Before upgrading
+
+1. Export your current stack parameters:
+
+   ```bash
+   aws cloudformation describe-stacks --stack-name MY_STACK_NAME \
+     --query 'Stacks[].Parameters[].[ParameterKey,ParameterValue]' --output table
+   ```
+
+2. Remove deleted parameters from parameter files and deployment automation. CloudFormation rejects parameters that aren't in the v7 template.
+3. If you use `AgentEnvFileUrl`, review that file against the Agent upgrade guide. CloudFormation can't check its contents.
+4. Preview the update with a [CloudFormation change set](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-changesets.html).
+
+If you need Agent v3, remain on Elastic CI Stack v6. Stack v7 no longer provides the `oldstable` channel.
+
+## Parameter changes
+
+| Stack v6 | Stack v7 |
+| --- | --- |
+| `BuildkiteAgentTimestampLines` | Removed. Agent v4 always emits ANSI timestamps. |
+| `BuildkiteAgentTracingBackend` | Removed. Use `BuildkiteAgentOpenTelemetryTracing=true` for OpenTelemetry. Agent v4 doesn't support the Datadog backend. |
+| `BuildkiteAgentCancelGracePeriod` and `BuildkiteAgentSignalGracePeriod` | Replaced by `BuildkiteAgentCancelSignalTimeout` and `BuildkiteAgentCancelCleanupTimeout`. |
+| `BuildkiteAgentRelease=oldstable` | Removed. Use `stable`, `beta`, or `edge`, or remain on stack v6. |
+
+## Cancellation timing
+
+Agent v4 separates the time given to a job process from the time reserved for agent cleanup:
+
+- `BuildkiteAgentCancelSignalTimeout` controls how long the process has before SIGKILL.
+- `BuildkiteAgentCancelCleanupTimeout` gives a stopping agent extra time to upload logs and artifacts.
+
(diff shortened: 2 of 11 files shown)


packer/linux/stack/conf/bin/bk-install-elastic-stack.sh
```diff
@@ -446,20 +446,39 @@ fi
...
+  agent_cancellation_config="cancel-grace-period=${BUILDKITE_AGENT_CANCEL_GRACE_PERIOD}
+signal-grace-period-seconds=${BUILDKITE_AGENT_SIGNAL_GRACE_PERIOD_SECONDS}"
+else
+  cancel_signal_timeout=$BUILDKITE_AGENT_SIGNAL_GRACE_PERIOD_SECONDS
+  if ((cancel_signal_timeout < 0)); then
+    cancel_signal_timeout=$((BUILDKITE_AGENT_CANCEL_GRACE_PERIOD + cancel_signal_timeout))
+  fi
+  if ((cancel_signal_timeout < 0 || cancel_signal_timeout >= BUILDKITE_AGENT_CANCEL_GRACE_PERIOD)); then
+    echo "BuildkiteAgentSignalGracePeriod must resolve to at least 0 and less than BuildkiteAgentCancelGracePeriod."
+    exit 1
```
petetomasik: Does this `exit 1` bypass the [trap](https://github.com/buildkite/elastic-ci-stack-for-aws/blob/1378fb86fa35641d9bd5a14f2e522a2797eac172/packer/linux/stack/conf/bin/bk-install-elastic-stack.sh#L46) at the beginning of this script?

scadu: Righto. Addressed in 8a23124

image

petetomasik (approved):

moskyb (commented):
is this going to be a major version release for the elastic stack? my argument would be for yes, because there's changes to agent behaviour that could break existing pipelines, and customers could get new agent versions by upgrading the stack(?).

if this is for a major release, i think it'd be a good idea to make the flags/inputs to the stack match the v4 agent flags and not have the compatibility shims

scadu:
@moskyb, I thought it’d be still v6, but with agent v4 it'd be a good reason to bump the Elastic CI Stack to v7.

~As for shims, in the current form (having options to choose between stable v4 and oldstable v3), it might still be needed. Although maybe not translating values from one parameter to the other, but throwing a warning when a parameter that’s no longer there in the agent was provided.~

Discussed internally and decided to drop `oldstable`, so no shims would be required.

.buildkite/pipeline.yaml
```diff
@@ -209,6 +209,8 @@ steps:
       - aws --version
       - jq --version
       - git --version
+      - buildkite-agent --version
+      - powershell -NoProfile -Command "$c = Get-Content C:\\buildkite-agent\\buildkite-agent.cfg; if (($c -notcontains 'opentelemetry-tracing=false') -or ($c -notcontains 'cancel-signal-timeout=10s') -or ($c -notcontains 'cancel-cleanup-timeout=5s')) { exit 1 }"
```
petetomasik: This check can't fail, for two independent reasons, so it's a permanent pass rather than a verification.

First, `$c` never reaches the nested PowerShell. Buildkite substitutes unescaped `$` at pipeline upload time (the docs require `$$` or `\$`), and this file already relies on that substitution for `${BUILDKITE_BUILD_NUMBER}` five lines down. Even if it survived upload, the Windows test stack runs job commands through PowerShell because `packer/windows/stack/conf/bin/bk-install-elastic-stack.ps1:155` sets `shell=powershell`, and the outer shell expands `$c` inside the double-quoted argument. Either way the child receives a script beginning with ` = Get-Content ...`, which is a parse error, and the cfg file is never read.

Second, even with `$c` escaped, `exit 1` can't fail the step. In agent v4 only `CMD.EXE` gets per-line error propagation via a generated batch script (`internal/job/executor.go:1306-1324`, `writeBatchScript` at `:1413-1428`). Every other shell gets the newline-joined command list handed straight to the interpreter, and PowerShell doesn't abort on a non-zero native exit. The nested `powershell` sets only the child's exit code, the outer script continues, and the step's status comes from the last command, `docker run ... echo hello`.

Worth flagging that the green build on this commit isn't evidence against the above: the check ran, parse-errored, and was swallowed.

Suggested fix, dropping the nesting since the step already runs under PowerShell, and using a terminating error so it actually propagates:

```yaml
      - $$cfg = Get-Content C:\buildkite-agent\buildkite-agent.cfg; foreach ($$k in 'opentelemetry-tracing=false','cancel-signal-timeout=10s','cancel-cleanup-timeout=5s') { if ($$cfg -notcontains $$k) { throw "missing $$k in buildkite-agent.cfg" } }
```

docs/upgrading-to-v7.md
```diff
@@ -0,0 +1,53 @@
...
+     --query 'Stacks[].Parameters[].[ParameterKey,ParameterValue]' --output table
+   ```
+
+2. Update parameter files and deployment automation. CloudFormation rejects parameters that aren't in the v7 template:
+   - Remove `BuildkiteAgentTimestampLines`. Agent v4 always emits ANSI timestamps.
+   - Replace `BuildkiteAgentTracingBackend` with `BuildkiteAgentOpenTelemetryTracing`: `""` becomes `false`, `opentelemetry` becomes `true`, and `datadog` has no direct replacement.
+   - Replace `BuildkiteAgentCancelGracePeriod` and `BuildkiteAgentSignalGracePeriod` with `BuildkiteAgentCancelSignalTimeout` and `BuildkiteAgentCancelCleanupTimeout`.
+   - Change `BuildkiteAgentRelease=oldstable` to `stable`, `beta`, or `edge`.
+3. If you use `AgentEnvFileUrl`, review that file against the Agent upgrade guide. CloudFormation can't check its contents.
+4. Preview the update with a [CloudFormation change set](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-changesets.html).
```
petetomasik: The "Before upgrading" checklist needs a step for customers running a custom AMI, because the v6-to-v7 update puts them into an instance churn loop with nothing in the CloudFormation events to explain it.

The v7 userdata no longer exports `BUILDKITE_AGENT_TIMESTAMP_LINES`, `BUILDKITE_AGENT_TRACING_BACKEND`, `BUILDKITE_AGENT_CANCEL_GRACE_PERIOD`, or `BUILDKITE_AGENT_SIGNAL_GRACE_PERIOD_SECONDS`. A v6 stack AMI's boot script expands all four inside its config heredoc with no `:-` fallback (`packer/linux/stack/conf/bin/bk-install-elastic-stack.sh:484,496-498` on `main`) under `set -Eeuo pipefail`. So on an AMI baked from the v6 base: unbound variable, `trap 'on_error $LINENO' ERR` fires, `on_error` calls `aws autoscaling set-instance-health --health-status Unhealthy`, the ASG replaces the instance, repeat. No jobs ever run.

`ImageId` and `ImageIdParameter` are documented, supported parameters (`templates/aws-stack.yml:839-847`), and baking extra tooling onto the stack AMI is a common enough pattern that this will find people.

Suggested step, after the parameter-file item:

> If you set `ImageId` or `ImageIdParameter`, rebuild your derived AMI from the v7 base AMI and update the parameter in the same stack update. A v6-based AMI cannot boot under the v7 template.

Worth adding `:-` defaults to the three new variables in `bk-install-elastic-stack.sh` as well. That doesn't help the case above (the abort happens inside the already-baked v6 AMI), but it fixes the reverse skew, which currently fails worse and silently on Windows: `packer/windows/stack/conf/bin/bk-install-elastic-stack.ps1:159-161` expands unset vars to empty, writing `cancel-signal-timeout=`, and `cliconfig/loader.go:246` in v4.0.1 is `value, _ = time.ParseDuration(configFileValue)`, which discards the error and yields 0. Agents then start normally and SIGKILL canceled jobs instantly, with no log line saying why. The existing `endpoint` [line](https://github.com/buildkite/elastic-ci-stack-for-aws/pull/1879/changes#diff-03c84324d14b7d5df342d35447de081589faf47189778c0967d6080dea3342caR459) already uses this pattern.

petetomasik (changes requested):

petetomasik (approved):

